PR #230, opened 2019-10-15
[website] docusaurus alpha26: correct use of the classic presets

3janeAI:
## Motivation

The current use of 3 parameters (show time, show author & url) is incorrect. Place them in their correct location.

Yes

## Test Plan

Tested locally, but didn't see the updated at/by footer.
This is the correct location for these properties, so it's better to make the change, or remove altogether.

## Related PRs

#224

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -17,9 +17,6 @@ module.exports = {
         googleAnalytics: {
           trackingID: 'UA-149862507-1',
         },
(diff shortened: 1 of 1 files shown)


3janeAI:
Looks like a transient error (for Python, so irrelevant to this PR), but I cannot re-run.
No point resubmitting for this -- it's not a Python PR.

omry:
Thanks, let's see if this works.

PR #234, opened 2019-10-16
creating objects: asserting on the `params` type; closes #231

### LINKED ISSUE #231: Better error message if config structure passed to instantiate is invalid.
# 🐛 Bug

In [common patterns#creating objects](https://cli.dev/docs/patterns/objects), if the `params` are mistakenly valued with a single string, the [`instantiate`](https://github.com/facebookresearch/hydra/blob/cb7904c666b34a53e09194993a0dd7b10fd97c99/hydra/utils.py#L44-L53) function exists with the error mentioned in "Stack trace/error messages":

This is fine to some extent, which is do you (Hydra team) aim at providing this kind of output messages? 

## To reproduce

Set `params` to `bug_report` as in:

```YAML
class: ...
params: bug_report
```

```
> AttributeError: 'str' object has no attribute 'merge_with'
[2019-10-16 02:25:21,592][hydra.utils][ERROR] - Error instantiating [[OBJECT_NAME]] : 'str' object has no attribute 'merge_with'
```
**Optional video**
If applicable, record a short [video](https://asciinema.org) with to help explain your problem.
You can also 

## Expected Behavior

I expect to have some sort of an error message explaining that `params` should be a collection of values where [[TYPE]] was provided. 
 
## System information
-  Hydra Version: 0.9.0

-  Operating System: Arch Linux

## Additional context
If you confirm that this should be fixed, I've already implemented the solution (IMO).

andrewnaguib:
## Motivation

Asserting on the `params` attribute when using common patterns and raising concise message is much more better than low level debugging messages to the user. 

Yes

## Test Plan

I'm not yet sure how Hydra does validate the types or layers of architecture. 

## Related Issues and PRs

Fixes #231

### DIFF
--- a/hydra/utils.py
+++ b/hydra/utils.py
@@ -1,7 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 import logging.config
 
-from omegaconf import OmegaConf
+from omegaconf import OmegaConf, DictConfig
 
 # pylint: disable=C0103
 log = logging.getLogger(__name__)
@@ -46,6 +46,11 @@ def instantiate(config, *args, **kwargs):
     try:
         clazz = get_class(config["class"])
         params = config.params if "params" in config else OmegaConf.create()
+        assert isinstance(
+            params, DictConfig
+        ), "Input config params are expected to be key-value items, found {}".format(
+            type(config.params)
+        )
         params.merge_with(OmegaConf.create(kwargs))
         return clazz(*args, **params)
     except Exception as e:
--- a/hydra/utils.py
+++ b/hydra/utils.py
@@ -48,7 +48,7 @@ def instantiate(config, *args, **kwargs):
         params = config.params if "params" in config else OmegaConf.create()
         assert isinstance(
             params, DictConfig
-        ), "Input config params are expected to be key-value items, found {}".format(
+        ), "Input config params are expected to be a mapping, found {}".format(
             type(config.params)
         )
         params.merge_with(OmegaConf.create(kwargs))


andrewnaguib:
I'm wondering why would the checks fail if I have the pre-commit hooks installed.

omry:
The failures are actually unit test errors.
Run them locally and see what needs to be fixed.

See [contibuting](https://cli.dev/docs/development/contributing) for help on testing locally.
About your question: you can lint locally with `nox lint`

hydra/utils.py
```diff
@@ -43,6 +43,11 @@ def get_static_method(full_method_name):
 
 def instantiate(config, *args, **kwargs):
     assert config is not None, "Input config is None"
+    assert isinstance(
+        config.params, (DictConfig, ListConfig)
```
andrewnaguib: @omry The reason why I'm keeping `ListConfig` is that [OmegaConf handles that](https://github.com/omry/omegaconf/blob/6c7c0f044281cdc1cecfebdce385466b7f48559b/omegaconf/config.py#L396) by raising 

> Merging DictConfig with ListConfig is not supported

The exception says "not supported" and doesn't throw a semantic error. If later on `OmegaConf` decides to add that support, we would still be limiting that in the `Hydra` layer even though it would work.

hydra/utils.py
```diff
@@ -43,6 +43,11 @@ def get_static_method(full_method_name):
 
 def instantiate(config, *args, **kwargs):
     assert config is not None, "Input config is None"
+    assert (
+        isinstance(config.params, (DictConfig, ListConfig)) or config.params is None
```
andrewnaguib: @omry The reason why I'm keeping ListConfig is that OmegaConf handles that by raising

    Merging DictConfig with ListConfig is not supported

The exception says "not supported" and doesn't throw a semantic error. If later on OmegaConf decides to add that support, we would still be limiting that in the Hydra layer even though it would work.

andrewnaguib:
BTW, I've run `nox -s test_core` on _your_ `master` branch and it fails for some reason on my machine.

hydra/utils.py
omry: So if you removed ListConfig here, wouldn't this code throw here instead which is the desired result?

andrewnaguib: Nope; I _think_ that we should leave `ListConfig` handling to `OmegaConf`.

omry: I suggest that you just check that the params is None or DictConfig.
Not seeing an extra value by allowing ListConfig here.

omry:
The failing tests are related to this change. try to reproduce them inside local pytest and fix them.
Because of strict mode, config.params throws if there is no params in config. you can check like this:

if 'params' in config and isinstance(config.params, DictConfig):

andrewnaguib:
I'm not sure of using `assert` inside `try..except` block, but since you're raising in the `except` part, it should be OK. If you think `if` stmt would be more readable, I could do it.

omry:
> 
> 
> I'm not sure of using `assert` inside `try..except` block, but since you're raising in the `except` part, it should be OK. If you think `if` stmt would be more readable, I could do it.

Yeah, it's a bit stinky but the catch there is really a catchall for all sorts of errors that can fly because of bad user input (wrong module/ class name etc).
I think this is reasonable.

BTW:
Did you figure out the nox problem on your local machine?

omry:
Awesome, thanks for the fix!

andrewnaguib:
Yeah, I did not have `expect` installed. You're welcome!

PR #256, opened 2019-10-23
Example Launcher plugin

omry:


omry:
1. moved example plugins into plugins/examples
2. new example Launcher plugin

PR #286, opened 2019-11-12
Ax plugin

omry:
Work in progress


omry:
Created #291  which is a cleaner version of this (without the example plugin, which was landed in #289)

PR #291, opened 2019-11-14
initial hydra_ax plugin

omry:
Ax (ax.dev) plugin for Hydra

PR #312, opened 2019-12-01
Hydra Plugin to interface with Adaptive Experimentation Platform

shagunsodhani:
## Motivation

Hydra Plugin to interface with [Adaptive Experimentation Platform](https://ax.dev/)

Yes

## Test Plan

To be added

## Related Issues and PRs

https://github.com/facebookresearch/hydra/pull/291


plugins/hydra_ax/README.md
```diff
@@ -1,7 +1,19 @@
-# Hydra example Launcher plugin
+# Hydra Ax Launcher plugin
 
-This plugin provides an example for how to write a custom Sweeper for Hydra.
+This plugin provides a mechanism for Hydra applications to use the [Adaptive Experimentation Platform aka Ax](https://ax.dev/). Ax can be used to optimize any kind of experiment - machine learning experiments, A/B tests and simulations. 
+
+The plugin can be installed by running the following command in the `plugins/hydra_ax` directory:
+
+```
+python setup.py install
```
omry: In general, do not use setup.py to install.
install with 
```bash
# current directory
pip install .
# in editable (develop) mode:
pip install -e .

# or package:
pip install hydra-ax
```
For this doc, please document the package method.

plugins/hydra_ax/README.md
```diff
@@ -1,7 +1,19 @@
...
+```
+
+We include an example of how to use this plugin. The file `plugins/hydra_ax/example/banana.py` implements the [Rosenbrock function (aka Banana function)](https://en.wikipedia.org/wiki/Rosenbrock_function). Ax Platform expects two return values from the function (that is being evaluated/optimized) - mean and SEM value. For more details on this, refer the [Ax documentation](https://ax.dev/tutorials/gpei_hartmann_service.html#3.-Define-how-to-evaluate-trials). Since we are using a deterministic function, we invoke the function only once, return the function value as the mean and the SEM value as 0.
```
omry: I think we should clean this up and return a single value from the function (handling the extra crap in the plugin).

plugins/hydra_ax/example/banana.py
```diff
@@ -8,12 +8,13 @@
...
 def banana(cfg):
+    print(cfg)
     a = cfg.banana.a
     b = cfg.banana.b
     x = cfg.banana.x
     y = cfg.banana.y
-    z = (a - x) ** 2 + b * (y - x ** 2) ** 2
-    log.info("Banana_a_{}_b_{}({}, {})={}".format(a, b, x, y, z))
+    z = (a - x) ** 2 + b * ((y - x ** 2) ** 2)
+    log.info("Banana_Function_a={}_b={}(x={}, y={})={}".format(a, b, x, y, z))
```
omry: Feel free to use Python 3 fstring here, this plugin depends on Python 3 through Ax anyway.

plugins/hydra_ax/hydra_plugins/ax_sweeper/ax_sweeper.py
```diff
@@ -155,48 +173,51 @@ def sweep(self, arguments):
...
                         "name": key,
                         "type": "range",
-                        "bounds": [x1, x2],
+                        "bounds": [range_start, range_end],
                         "parameter_type": param_type,
                     }
                 )
             else:
-                # fixed
+                # This is a Fixed Parameter.
```
omry: we need to test fixed parameters, I think this does not work. (I remember some exception).

plugins/hydra_ax/setup.py
```diff
@@ -22,7 +22,7 @@
             "Programming Language :: Python :: 3.7",
             "Operating System :: OS Independent",
         ],
-        install_requires=["hydra-core"],
+        install_requires=["hydra-core", "ax-platform==0.1.6", "SQLAlchemy==1.3.11"],
```
omry: Can you talk to the Ax team and see what are the proper dependencies they recommend?

omry:
Also, fix integration test. see [contributing](https://cli.dev/docs/development/contributing) doc for help on how to run the tests locally.

shagunsodhani:
> I would like to support a few more things for v0:
> 
> 1. expose logscale.
> 2. allow defining the sweep parameters in the config file (not sure what is the best way to do this).
>    The objective being that for things like logscale (and possibly other in the future) the user will be able to specify the type in the config.
>    config should also be optional per variable: if it's defined in the config we use it as base, otherwise we use what's in the command line. I propose something like:
> 
> ```yaml
> ax:
>   params:
>      x:
>        type: range
>        logscale: true # defaults to false
> ```
> 
> does that make sense?

I agree with the suggestion for the logscale. I am wondering if the user should have to specify the parameter type as the current code infers it (covers the reasonable cases).

plugins/hydra_ax/README.md
shagunsodhani: We do not even have to handle the SEM value. The [Ax documentation](https://ax.dev/docs/trial-evaluation.html#evaluation-function) mentions that the evaluate function `can also return only the mean as a float, in which case Ax will treat SEM as unknown and use a model that can infer it.` So we can just return the mean value. I will update this.

plugins/hydra_ax/README.md
shagunsodhani: To use the package method, do we not have to upload the package to pypi ?

plugins/hydra_ax/README.md
shagunsodhani: Fixed in : https://github.com/facebookresearch/hydra/pull/312/commits/08675cb8306a999b7939e11f6417bad01e17f0f9

plugins/hydra_ax/README.md
omry: Yes, and I will.

omry:
Yes, we can allow it to be optionally specified. (this can take precedence over inferring from the command line).

PR #356, opened 2020-01-08
testing all plugins in the same nox session to speed testing up

omry:


omry:
cc @shagunsodhani, please rebase your Ax plugin diff on top of this once this lands, it should make nox testing for plugins much faster.

PR #375, opened 2020-01-16
GitHub ConfigSource plugin

omry:
This is a somewhat practical demo of how configsource plugins can be used.


omry:
GitHub does not support sparse checkout and is too slow for this.
Aborting this attempt.

PR #387, opened 2020-01-20
Integrating structured config - WIP

omry:


omry:
closing in favor of new PR with the same code (avoiding accidental overwrite of important changes by keeping this as is instead of force pushing onto it).

PR #388, opened 2020-01-22
Parent node's information is lost when performing deepcopy.

shagunsodhani:
## Motivation

Parent node's information is lost when performing deepcopy.

OmegaConf misbehaves with the deepcopy function and the information about the parent (node) is lost. This leads to errors where a variable's value can not be resolved. The error message looks like this -
`KeyError: "str interpolation key 'hydra.ax.experiment' not found"`. This commit provides a temporary workaround for the problem, by setting the parent of the copied config as that of the original config.
One testcase is also added to check for this bug.

Yes

## Test Plan

Added a testcase.

### DIFF
--- a/hydra/utils.py
+++ b/hydra/utils.py
@@ -51,8 +51,7 @@ def instantiate(config: DictConfig, *args: Any, **kwargs: Any) -> Any:
     assert config is not None, "Input config is None"
     # copy config to avoid mutating it when merging with kwargs
     config_copy = copy.deepcopy(config)
-    if config._get_parent() is not None:
-        config_copy._set_parent(config._get_parent())
+    config_copy._set_parent(config._get_parent())
     config = config_copy
 
     try:
--- a/hydra/utils.py
+++ b/hydra/utils.py
@@ -50,7 +50,10 @@ def instantiate(config: DictConfig, *args: Any, **kwargs: Any) -> Any:
 
     assert config is not None, "Input config is None"
     # copy config to avoid mutating it when merging with kwargs
-    config = copy.deepcopy(config)
+    config_copy = copy.deepcopy(config)
+    if config._get_parent() is not None:
+        config_copy._set_parent(config._get_parent())
+    config = config_copy
 
     try:
         clazz = get_class(config["class"])
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -105,6 +105,35 @@ def test_class_instantiate_passthrough(
     assert obj == expected
 
 
+@pytest.mark.parametrize(  # type: ignore
+    "input_conf, expected",
+    [
+        (
+            {
+                "all_params": {
+                    "main": {
+                        "class": "tests.test_utils.Bar",
+                        "params": {"a": 10, "b": 20, "c": "${all_params.aux.c}"},
+                    },
+                    "aux": {"c": 30},
+                }
+            },
+            Bar(10, 20, 30, 40),
+        ),
+    ],
+)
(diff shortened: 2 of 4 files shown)


hydra/utils.py
```diff
@@ -50,7 +50,10 @@ def instantiate(config: DictConfig, *args: Any, **kwargs: Any) -> Any:
 
     assert config is not None, "Input config is None"
     # copy config to avoid mutating it when merging with kwargs
-    config = copy.deepcopy(config)
+    config_copy = copy.deepcopy(config)
+    if config._get_parent() is not None:
```
omry: you can set a None parent. did you ran into issues?

tests/test_utils.py
```diff
@@ -105,6 +105,35 @@ def test_class_instantiate_passthrough(
...
+            Bar(10, 20, 30, 40),
+        ),
+    ],
+)
+def test_string_iterpolation_during_deepcopy(
+    input_conf: Dict[str, Any], expected: Any
+) -> None:
+    # Check if the instantiate method maintains the parent when making a deepcopy
+    conf = OmegaConf.create(input_conf)
+    orig = copy.deepcopy(conf)
```
omry: I don't think you need to do the deepcopy here.
Instantiate is doing the deepcopy.

omry: You are also testing passthrough.
just make your test another case of test_class_instantiate.
(You can add additional parameters set to the list there).

hydra/utils.py
shagunsodhani: Ohh I did not try. I can add a separate testcase for that. I added the check to be on the safe side.

omry: Yes. I don't think you can even create the test set. params always has a parent.

shagunsodhani: How about this: 
```@pytest.mark.parametrize(  # type: ignore
    "input_conf, expected",
    [
        (
            {
                    "main": {
                        "class": "tests.test_utils.Bar",
                        "params": {"a": 10, "b": 20, "c": "${all_params.aux.c}"},
                    },
                    "aux": {"c": 30},
                },
            Bar(10, 20, 30, 40),
        ),
    ],
)
def test_string_iterpolation_during_deepcopy(
    input_conf: Dict[str, Any], expected: Any
) -> None:
    # Check if the instantiate method maintains the parent when making a deepcopy
    conf = OmegaConf.create(input_conf)
    orig = copy.deepcopy(conf)
    assert isinstance(conf, DictConfig)
    obj = utils.instantiate(conf.main, d=40)
    assert orig == conf
    assert obj == expected
```

tests/test_utils.py
shagunsodhani: I can get rid of the deepcopy part. Regarding adding extra params to `test_class_instantiate`, I think it will make it a little messy as we need a nested config and need to call `instantiate` over a specific node in the config. If that is fine, I can merge the two testcases.

hydra/utils.py
shagunsodhani: Ohh and I can set a None parent :)

tests/test_utils.py
omry: As long as the test code is the same (and it can be), you can reduce duplication with parametrize.

hydra/utils.py
shagunsodhani: Removed the none check

tests/test_utils.py
shagunsodhani: Merged the test cases

PR #404, opened 2020-02-10
hydra.utils.instantiate() : allow passthrough of non primitive objects by name

### LINKED ISSUE #400: [Bug] issues with named arguments in hydra.utils.instantiate
`hydra.utils.instantiate` converts to OmegaConf named arguments, making it impossible to use the named arguments syntax for arguments which cannot be represented in OmegaConf.

## To reproduce:

`conf.yaml`:

```yaml
class: torch.optim.Adam
params:
  lr: 1E-4
```

`test.py`:

```python
import hydra
import torch
import torchvision

@hydra.main(config_path="conf.yaml")
def my_app(cfg):
    print(cfg.pretty())
    model = torchvision.models.alexnet()
    hydra.utils.instantiate(cfg, model.parameters())  # OK
    hydra.utils.instantiate(cfg, params=model.parameters())  # FAILS

if __name__ == "__main__":
    my_app()
```

```
ValueError: key params: generator is not a primitive type
```

As a side note, the fact that `hydra.utils.instantiate` takes as input an argument named `config` makes it impossible to instantiate an object which expects the same argument using named arguments, which seems an unnecessary restriction.

## The following solves both issues:

```python
def instantiate_f(config):
    class_name = config["class"]
    clazz = get_class(class_name)
    params = config.params if "params" in config else OmegaConf.create()
    assert isinstance(
        params, DictConfig
    ), "Input config params are expected to be a mapping, found {}".format(
        type(config.params)
    )
    params = OmegaConf.to_container(params, resolve=True)

    def f(*args, **kwargs):
        try:
            kwargs = {**params, **kwargs}
            return clazz(*args, **kwargs)
        except Exception as e:
            log.error("Error instantiating {} : {}".format(class_name, e))
            raise e

    return f

@hydra.main(config_path="conf.yaml")
def my_app(cfg):
    print(cfg.pretty())
    model = torchvision.models.alexnet()
    instantiate_f(cfg)(params=model.parameters())  # OK
    instantiate_f(cfg)(model.parameters())  # OK

if __name__ == "__main__":
    my_app()
```

A note in the documentation should be added explaining the behavior wrt `config`, as it gets resolved immediately.

omry:
Closes #400

### DIFF
--- a/examples/patterns/objects/conf/db/mysql.yaml
+++ b/examples/patterns/objects/conf/db/mysql.yaml
@@ -1,5 +1,5 @@
 db:
-  class: tutorial.objects_example.my_app.MySQLConnection
+  cls: tutorial.objects_example.my_app.MySQLConnection
   params:
     host: localhost
     user: root
--- a/examples/patterns/objects/conf/db/postgresql.yaml
+++ b/examples/patterns/objects/conf/db/postgresql.yaml
@@ -1,5 +1,5 @@
 db:
-  class: tutorial.objects_example.my_app.PostgreSQLConnection
+  cls: tutorial.objects_example.my_app.PostgreSQLConnection
   params:
     host: localhost
     user: root
--- a/hydra/conf/__init__.py
+++ b/hydra/conf/__init__.py
@@ -25,6 +25,8 @@ hydra_defaults = [
 
 
 @dataclass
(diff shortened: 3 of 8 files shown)


omry:
unfortunately this depends on some other fix I have in another branch that is not ready yet.
I will land it later.

Issue #421, opened 2020-02-16
[Bug] sporadic segfaults from joblib tests in CircleCI


omry:
Example:

https://app.circleci.com/jobs/github/facebookresearch/hydra/8035

@jan-matthis, can you take a look? this one is specifically on Linux.

1. I think we need to report to joblib.
2. I suspect threading (even before looking at the stacktrace). we might want to disable the threading test temporarily.

jan-matthis:
The relevant part of the CircleCI logs is:

```text
tests/test_joblib_launcher.py .......Fatal Python error: Segmentation fault

Thread 0x00007fba5687d700 (most recent call first):
  File "/usr/local/lib/python3.6/threading.py", line 295 in wait
  File "/home/circleci/project/.nox/test_plugins-3-6-pip-install/lib/python3.6/site-packages/joblib/externals/loky/backend/queues.py", line 138 in _feed
  File "/usr/local/lib/python3.6/threading.py", line 864 in run
  File "/usr/local/lib/python3.6/threading.py", line 916 in _bootstrap_inner
  File "/usr/local/lib/python3.6/threading.py", line 884 in _bootstrap

Thread 0x00007fba5607c700 (most recent call first):
  File "/usr/local/lib/python3.6/selectors.py", line 376 in select
  File "/usr/local/lib/python3.6/multiprocessing/connection.py", line 911 in wait
  File "/home/circleci/project/.nox/test_plugins-3-6-pip-install/lib/python3.6/site-packages/joblib/externals/loky/process_executor.py", line 615 in _queue_management_worker
  File "/usr/local/lib/python3.6/threading.py", line 864 in run
  File "/usr/local/lib/python3.6/threading.py", line 916 in _bootstrap_inner
  File "/usr/local/lib/python3.6/threading.py", line 884 in _bootstrap
```

The segfault involves the loky backend, namely the files:
- `joblib/externals/loky/backend/queues.py` ([L138](https://github.com/joblib/joblib/blob/master/joblib/externals/loky/backend/queues.py#L138))
- `joblib/externals/loky/process_executor.py` ([L615](https://github.com/joblib/joblib/blob/master/joblib/externals/loky/process_executor.py#L615))

Loky is the default backend used by `Joblib.Parallel`. It spawns workers in different processes and uses threading for managing queues. Since loky is involved in the segfault, disabling the test that uses the threading backend (workers in threads), probably won't help as a fix. 

[...]

PR #423, opened 2020-02-18
Hard-code Joblib backend to loky

jan-matthis:
## Motivation

Launching jobs using thread-based parallelism is incompatible with Hydra. This PR updates the Joblib Launcher plugin to hard-code the backend to process-based parallelism using Joblib's loky backend. 

Yes

## Test Plan

Run multiple tests on CircleCI in order to cover sporadically occurring failures

## Related Issues and PRs

https://github.com/facebookresearch/hydra/issues/421

### DIFF
--- a/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
+++ b/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
@@ -79,7 +79,7 @@ class JoblibLauncher(Launcher):
 
         # Joblib's backend is hard-coded to loky since the threading
         # backend is incompatible with Hydra's logging
-        joblib_keywords = OmegaConf.to_container(self.joblib).copy()
+        joblib_keywords = OmegaConf.to_container(self.joblib, resolve=True)
         joblib_keywords["backend"] = "loky"
 
         log.info(
--- a/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
+++ b/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence
 
 from joblib import Parallel, delayed  # type: ignore
-from omegaconf import DictConfig, open_dict
+from omegaconf import DictConfig, OmegaConf, open_dict
 
 from hydra.core.config_loader import ConfigLoader
 from hydra.core.config_search_path import ConfigSearchPath
@@ -76,20 +76,25 @@ class JoblibLauncher(Launcher):
         configure_log(self.config.hydra.hydra_logging, self.config.hydra.verbose)
         sweep_dir = Path(str(self.config.hydra.sweep.dir))
         sweep_dir.mkdir(parents=True, exist_ok=True)
+
+        # Joblib's backend is hard-coded to loky since the threading
+        # backend is incompatible with Hydra's logging
+        joblib_keywords = OmegaConf.to_container(self.joblib).copy()
+        joblib_keywords["backend"] = "loky"
+
         log.info(
             "Joblib.Parallel({}) is launching {} jobs".format(
-                ",".join([f"{k}={v}" for k, v in self.joblib.items()]),
+                ",".join([f"{k}={v}" for k, v in joblib_keywords.items()]),
                 len(job_overrides),
             )
         )
         log.info("Launching jobs, sweep output dir : {}".format(sweep_dir))
-
-        singleton_state = Singleton.get_state()
-
         for idx, overrides in enumerate(job_overrides):
             log.info("\t#{} : {}".format(idx, " ".join(filter_overrides(overrides))))
 
-        runs = Parallel(**self.joblib)(
+        singleton_state = Singleton.get_state()
+
+        runs = Parallel(**joblib_keywords)(
             delayed(execute_job)(
                 idx,
                 overrides,
@@ -116,9 +121,6 @@ def execute_job(
     singleton_state: Dict[Any, Any],
 ) -> JobReturn:
     """Calls `run_job` in parallel
-
-    Note that Joblib's default backend runs isolated Python processes, see
-    https://joblib.readthedocs.io/en/latest/parallel.html#shared-memory-semantics
     """
     setup_globals()
     Singleton.set_state(singleton_state)
--- a/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
+++ b/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
@@ -78,7 +78,7 @@ class JoblibLauncher(Launcher):
         sweep_dir.mkdir(parents=True, exist_ok=True)
 
         # Joblib's backend is hard-coded to loky since the threading
-        # backend is incompatible with Hydra's logging
+        # backend is incompatible with Hydra
         joblib_keywords = OmegaConf.to_container(self.joblib, resolve=True)
         joblib_keywords["backend"] = "loky"
 
--- a/website/docs/plugins/joblib_launcher.md
+++ b/website/docs/plugins/joblib_launcher.md
@@ -18,7 +18,7 @@ defaults:
   - hydra/launcher: joblib
 ```
 
-By default, process-based parallelism using all available CPU cores is used. By overriding the default configuration, it is e.g. possible to switch to thread-based parallelism and limit the number of parallel executions.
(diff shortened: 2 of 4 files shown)


plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
```diff
@@ -76,20 +76,25 @@ def launch(self, job_overrides: Sequence[Sequence[str]]) -> Sequence[JobReturn]:
         configure_log(self.config.hydra.hydra_logging, self.config.hydra.verbose)
         sweep_dir = Path(str(self.config.hydra.sweep.dir))
         sweep_dir.mkdir(parents=True, exist_ok=True)
+
+        # Joblib's backend is hard-coded to loky since the threading
+        # backend is incompatible with Hydra's logging
```
omry:
```suggestion
        # backend is incompatible with Hydra.
```

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
```diff
@@ -76,20 +76,25 @@ def launch(self, job_overrides: Sequence[Sequence[str]]) -> Sequence[JobReturn]:
...
+        joblib_keywords = OmegaConf.to_container(self.joblib).copy()
```
omry: 1. Why are you converting to primitive dict?
2. no need to copy, to_container is already copying.
3. use resolve=True to ensure any interpolation is resolved.

website/docs/plugins/joblib_launcher.md
```diff
@@ -58,21 +49,20 @@ hydra:
...
     # thresholds size of arrays that triggers automated memmapping
-    max_nbytes: 1M
+    max_nbytes: null
 
     # memmapping mode for numpy arrays passed to workers
     mmap_mode: r
 ```
 
-`n_jobs` defaults to -1 (all available CPUs) and `prefer` defaults to `processes`. Depending on the application, `threads` might be a good alternative. All arguments specified in `joblib` are passed to `Joblib.Parallel` (see [`Joblib.Parallel` documentation](https://joblib.readthedocs.io/en/latest/parallel.html) for full details). 
+`n_jobs` defaults to -1 (all available CPUs). See [`Joblib.Parallel` documentation](https://joblib.readthedocs.io/en/latest/parallel.html) for full details on arguments. Note that the backend is hard-coded to use process-based parallelism (Joblib's loky backend), since thread-based parallelism is incompatible with Hydra's logging.
```
omry:
```suggestion
`n_jobs` defaults to -1 (all available CPUs). See [`Joblib.Parallel` documentation](https://joblib.readthedocs.io/en/latest/parallel.html) for full details on arguments. Note that the backend is hard-coded to use process-based parallelism (Joblib's loky backend), since thread-based parallelism is incompatible with Hydra.
```

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
jan-matthis: 1. I want to "backend" to the arguments for `Joblib.Parallel` without having it in the config (to hard-code it). If I did `self.joblib["backend"] = "loky"`, I'd get: `AttributeError: Accessing unknown key in a struct : hydra.joblib.backend`

Will fix 2. and 3.

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
```diff
@@ -76,20 +76,25 @@ def launch(self, job_overrides: Sequence[Sequence[str]]) -> Sequence[JobReturn]:
         configure_log(self.config.hydra.hydra_logging, self.config.hydra.verbose)
         sweep_dir = Path(str(self.config.hydra.sweep.dir))
         sweep_dir.mkdir(parents=True, exist_ok=True)
+
+        # Joblib's backend is hard-coded to loky since the threading
+        # backend is incompatible with Hydra
+        joblib_keywords = OmegaConf.to_container(self.joblib, resolve=True)
+        joblib_keywords["backend"] = "loky"
```
tomMoral: If I may, just a comment: you should warn the user that keyword arg `backend` is bypassed if it is set explicitely by the user.

There is other process based backends than `loky` for `joblib` such as `dask`, `hadoop` and others. Bypassing an explicit without warning will create weird behavior and might confuse some users.

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
omry: Gotcha.
Read [this](https://omegaconf.readthedocs.io/en/latest/usage.html#configuration-flags) for more context on why you get it.
converting to a container is one way to deal with it, another is to open the structure to changes with open_dict.

I actually think having backend: "loki" hard coded in the config and raising an exception if the user try to set it to threads is better.
1. I will prevent threads from being used.
2. I will make it easy for others to try a different backend.
it will most likely fail but at least they can try to make with config changes only.
We can documented that the only tested and backend is loki

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
omry: This is not overriding but preventing the user from set the backend  at all from the config.
```
>>> from omegaconf import OmegaConf
>>> c1 = OmegaConf.create({"foo": 10})
>>> OmegaConf.set_struct(c1, True)
>>> OmegaConf.merge(c1, {"foo": 20})
{'foo': 20}

>>> OmegaConf.merge(c1, {"foo": 20, "backend": "loki"})
...
AttributeError: Accessing unknown key in a struct : backend
```
The concern is that there are issues with other backends as well that we do not have the time to test right now.
I agree with you that it's best to just block the threading backend and allow users to play with the others.

If you have some cycles, It would be great if you can help testing some of the other backends.
If they need some additional config we will probably need to change the plugin a bit to support it.

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
jan-matthis: I can easily do that, however, this will be the situation:

- The `backend` keyword of Joblib 0.14.1 allow selection between `['loky', 'multiprocessing', 'sequential', 'threading']`:
  - `multiprocessing` won't work without major plugin changes. It fails when trying to pickle the task function and config loader. Beyond that, Joblib advises using `loky` rather than `multiprocessing` in the docs, since it's considered more stable
  - `sequential` enables sequential computation
  - `threading` is incompatible
- In effect, `loky` is the only reasonable choice for `backend` at the moment
- Alternative backends like `dask`, `hadoop`, or `ray` (!), are not selected through the `backend` keyword of `Joblib.Parallel` but use context managers and require additional configuration and plugin changes

What do you think?

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
tomMoral: I agree that this is a good idea to restrict the backend to `loky`. My only concern is the fact that this is silent if the user give a different `backend` argument. I would say just keep this behavior but raise a warning. I believe no one will see it but just in case, it will avoid some troubles in debugging as it will be clear why the backend is changed back to `loky` :)

jan-matthis: Ah, so, the current version of the PR takes out `backend` from the plugin config altogether. In effect, users cannot provide a `backend` argument to the plugin at all. `loky` will always be used. 

Alternatively, we might offer `backend` in the config and raise an exception if set to `threading`. However, I'm not sure this offers any benefits, also see my reply here: https://github.com/facebookresearch/hydra/pull/423#discussion_r381162808

tomMoral: Nevermind then if users cannot pass the argument directly!
Thanks for your explanation.

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py
omry: Okay - it sounds like Loki is the only sane choice right now, so your solution sounds good.

omry:
oh dang, it fails CI. I didn't notice that CI did not run on this PR.
I will fix it.

In the future, please at least run nox locally if CI is not running on your PR.

omry:
https://github.com/facebookresearch/hydra/pull/424

PR #434, opened 2020-02-20
Add ax plugin

shagunsodhani:
## Motivation

Hydra Plugin to interface with [Adaptive Experimentation Platform](https://ax.dev/)

Yes

## Test Plan

To be added

## Related Issues and PRs

https://github.com/facebookresearch/hydra/pull/291


noxfile.py
```diff
@@ -129,6 +132,31 @@ def get_plugin_python_version(session, plugin):
     )
 
 
+def get_plugin_os_names(session: nox.session, plugin: str) -> List[str]:
+    setup_py = os.path.join(BASE, "plugins", plugin["path"], "setup.py")
+    out = session.run("python", setup_py, "--classifiers", silent=True).splitlines()
+    oses = list(filter(lambda line: "Operating System" in line, out))
+    if len(oses) == 1 and oses[0] == "Operating System :: OS Independent":
```
omry: What if OS is not specified at all?

noxfile.py
```diff
@@ -143,17 +171,25 @@ def test_plugins(session, install_cmd):
     all_plugins = get_all_plugins()
     # Install all supported plugins in session
     for plugin in all_plugins:
-        pythons = get_plugin_python_version(session, plugin)
         # Verify this plugin supports the python we are testing on, skip otherwise
         plugin_python_versions = get_plugin_python_version(session, plugin)
-        plugin_enabled[plugin["path"]] = session.python in pythons
+        plugin_enabled[plugin["path"]] = session.python in plugin_python_versions
```
omry: Simplify this function. this and see comments below.

```python
os_supported = check_if_os_supports_plugin(session, plugin)
python_supported = session.python in plugin_python_versions
plugin_enabled[plugin["path"]] = os_supported and python_supported
```

noxfile.py
```diff
@@ -185,20 +221,23 @@ def coverage(session):
...
     # Install all plugins in session
     for plugin in get_all_plugins():
-        session.run(
-            "pip",
-            "install",
-            "-e",
-            os.path.join("plugins", plugin["path"]),
-            silent=SILENT,
-        )
+        if check_if_os_supports_plugin(session, plugin):
```
omry: not good. you need to check both python is supported and os is supported.

```python
if not plugin_enabled[plugin["path"]]:
```

noxfile.py
```diff
@@ -143,17 +171,25 @@ def test_plugins(session, install_cmd):
...
         if not plugin_enabled[plugin["path"]]:
             py_str = ",".join(plugin_python_versions)
             session.log(
                 f"Not testing {plugin['name']} on Python {session.python}, supports [{py_str}]"
             )
             continue
 
+        # Verify this plugin supports the OS we are testing on, skip otherwise
+        plugin_enabled[plugin["path"]] = check_if_os_supports_plugin(session, plugin)
+        if not plugin_enabled[plugin["path"]]:
```
omry: if not os_supported:

noxfile.py
```diff
@@ -143,17 +171,25 @@ def test_plugins(session, install_cmd):
     all_plugins = get_all_plugins()
     # Install all supported plugins in session
     for plugin in all_plugins:
-        pythons = get_plugin_python_version(session, plugin)
         # Verify this plugin supports the python we are testing on, skip otherwise
         plugin_python_versions = get_plugin_python_version(session, plugin)
-        plugin_enabled[plugin["path"]] = session.python in pythons
+        plugin_enabled[plugin["path"]] = session.python in plugin_python_versions
         if not plugin_enabled[plugin["path"]]:
```
omry: if not python_supported:

noxfile.py
```diff
@@ -185,20 +221,23 @@ def coverage(session):
...
+            )
 
     session.run("coverage", "erase")
     session.run("coverage", "run", "--append", "-m", "pytest", silent=SILENT)
     for plugin in list_plugins():
         plugin_python_versions = get_plugin_python_version(session, plugin)
         if session.python not in plugin_python_versions:
             continue
+        if not check_if_os_supports_plugin(session, plugin):
+            continue
```
omry:
```python
if not plugin_enabled[plugin["path"]]:
```

plugins/hydra_ax_sweeper/README.md
```diff
@@ -0,0 +1,72 @@
...
+```
+pip install hydra_ax_sweeper
+```
+
+We include an example of how to use this plugin. The file [`plugins/hydra_ax_sweeper/example/banana.py`](plugins/hydra_ax/example/banana.py) implements the [Rosenbrock function (aka Banana function)](https://en.wikipedia.org/wiki/Rosenbrock_function). The return value of the function should be the value that we want to optimize.
+
+To compute the best parameters for the Banana function, clone the code and run the following command in the `plugins/hydra_ax_sweeper` directory:
+
+```
+python example/banana.py -m banana.x=-5:5 banana.y=-5:10.1
```
omry: Since we want the user to use floating point for banana function, let's show a correct example.

plugins/hydra_ax_sweeper/README.md
```diff
@@ -0,0 +1,72 @@
...
+ log_scale: false
+```
+
+The `x` parameter takes on a "range" of values, between `-5` to `10`,  and the values should not be sampled from the log-space. In general, the plugin supports all the [Parameters](https://ax.dev/api/core.html?highlight=range#module-ax.core.parameter) that Ax supports. According to the [Ax documentation](https://ax.dev/api/service.html#ax.service.ax_client.AxClient.create_experiment), the required elements in the config are:
+
+* `name` - Name of the parameter. It is of type string.
+* `type` - Type of the parameter. It can take the following values: `range`, `fixed`, or `choice`.
+* `bounds` - Required only for the `range` parameters. It should be a list of two values, with the lower bound first.
+* `values` - Required only for the `choice` parameters. It should be a list of values.
+* `value` - Required only for the `fixed` parameters. It should be a single value.
```
omry: 1. Why is parameter_type not documented above?
2. Document interaction with the command line type inference.

plugins/hydra_ax_sweeper/README.md
```diff
@@ -0,0 +1,72 @@
...
+ax:
+    experiment:
+      # If the function should be minimized (Ax defaults to maximize)
+      minimize: true
+
+    early_stop:
+      # Number of epochs without a significant improvement from
+      # the currently known best parameters
+      # An Epoch is defined as a batch of trials executed in parallel
+      max_epochs_without_improvement: 100
```
omry: give a saner max_epochs_without_improvement value in the example.

plugins/hydra_ax_sweeper/example/conf/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+    
+    params:
+      banana.x:
+        type: range
+        bounds: [-5, 5]
+        log_scale: false
+      banana.y:
+        type: range
+        bounds: [-5, 10.1]
+        log_scale: false
```
omry: In this example, since we always want to use float for this function, we should probably use parameter_type to enforce it. (and tell the user what we are doing in a comment.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/ax_sweeper.py
```diff
@@ -0,0 +1,303 @@
...
+                    num_trials_left = -1
+                    break
+
+            current_parallelism_index += 1
+
+        with open(f"{self.sweep_dir}/optimization_results.yaml", "w") as f:
+            results_to_serialize = {"optimizer": "ax", "ax": best}
+            results_to_serialize = json.loads(json.dumps(results_to_serialize))
+            # This step is to convert all the numpy floats into python floats
+            OmegaConf.save(OmegaConf.create(results_to_serialize), f)
```
omry: Much code, This is way too complicated.
try those two lines instead of the entire with block.
```python
result = OmegaConf.create({"optimizer": "ax", "ax": best.tolist()})
OmegaConf.save(result, f"{self.sweep_dir}/optimization_results.yaml")
```

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/ax_sweeper.py
```diff
@@ -0,0 +1,303 @@
...
+                        "values": [int(x) for x in value_choices],
+                        "parameter_type": ParameterType.INT,
+                    }
+                elif all(_is_float(x) for x in value_choices):
+                    param = {
+                        "name": key,
+                        "type": "choice",
+                        "values": [float(x) for x in value_choices],
+                        "parameter_type": ParameterType.FLOAT,
+                    }
```
omry: Looking here, it appears that the command line decides on parameter_type.
I suggest something else:

If parameter_type is specified in the config, this is it. the command line should be converted to it.
if it's not specified, you infer it from the types.

Don't forget to add tests covering this logic.


This is a bit much. we can do it in a followup diff after we land this one.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/conf/hydra/sweeper/ax.yaml
```diff
@@ -0,0 +1,38 @@
...
+  sweeper:
+    class: hydra_plugins.hydra_ax_sweeper.AxSweeper
+    params:
+      experiment: ${hydra.ax.experiment}
+      early_stop: ${hydra.ax.early_stop}
+      verbose_logging: false
+      max_trials: ${hydra.ax.max_trials}
+      # set random seed here to make Ax results reproducible
+      random_seed: null
+      ax_params: ${hydra.ax.params}
```
omry:
```yaml
  sweeper:
    class: hydra_plugins.hydra_ax_sweeper.AxSweeper
    params:
      ax:  ${hydra.ax}
```

You can simplify this by passing hydra.ax as is.
the code will need to be changed to accept the cfg object instead of the individual values.

This is better because the config is much more user visible.

plugins/hydra_ax_sweeper/setup.py
```diff
@@ -0,0 +1,24 @@
...
+        long_description_content_type="text/markdown",
+        url="https://github.com/facebookresearch/hydra/",
+        packages=find_packages(exclude=["tests", "example"]),
+        classifiers=[
+            "License :: OSI Approved :: MIT License",
+            "Programming Language :: Python :: 3.6",
+            "Programming Language :: Python :: 3.7",
+            "Operating System :: POSIX :: Linux",
+        ],
+        install_requires=["hydra-core>=1.0.0-rc1", "ax-platform[mysql]==0.1.9"],
```
omry: 1. Is [mysql] really requires here?

shagunsodhani: yes.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/ax_sweeper.py
shagunsodhani: No. cmd is supposed to override the config. Why would we want to confuse the user by requiring them to specify the type via the config file? Doesnt that break the entire point of using cmd?

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/conf/hydra/sweeper/ax.yaml
shagunsodhani: Not sure I follow what is missing here.

plugins/hydra_ax_sweeper/README.md
shagunsodhani: What is wrong in this example? We clearly explain how the parameters are being type casted.

plugins/hydra_ax_sweeper/README.md
shagunsodhani: Would 10 be fine?

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/ax_sweeper.py
shagunsodhani: This fails as `AttributeError: 'tuple' object has no attribute 'tolist'` and `best` is a tuple.
`result = OmegaConf.create({"optimizer": "ax", "ax": best})` fails as 
`omegaconf.errors.UnsupportedValueType: key ax: tuple is not a supported type`. 

I think it is fine to load and unload a dict as json. I can get rid of line 83 and merge 84 with 85.

plugins/hydra_ax_sweeper/example/conf/config.yaml
shagunsodhani: `parameter_type` is not a required parameter (even as per Ax). Ax Plugin will promote types (wether launched via cmd or config) when it finds a mismatch (int mixed with float). If the user wants, they can specify `parameter_type`. I do not see the point of forcing this.

noxfile.py
shagunsodhani: Testing for all OS now

plugins/hydra_ax_sweeper/README.md
shagunsodhani: 1. We are only documenting the `required elements` in the config.
2. Done

plugins/hydra_ax_sweeper/README.md
omry: For this function, the correct choice is to use floating point numbers. So while the example demos the plugin is gives confusing guidance about this function.

plugins/hydra_ax_sweeper/README.md
omry: sounds good to me.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/conf/hydra/sweeper/ax.yaml
omry: This is not about missing, but about style.
Instead of manually unpacking ${hydra.ax} into params, you can just pass it directly per my example above.

plugins/hydra_ax_sweeper/README.md
omry: We have 4 different threads about parameter_type, let's chat about it.

noxfile.py
shagunsodhani: Yes, you are right. I have fixed this in the new commit. The comment `Testing for all OS now` is the behavior after the new commit. I should have been more explicit :)

plugins/hydra_ax_sweeper/README.md
shagunsodhani: Ahh got it :)

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/conf/hydra/sweeper/ax.yaml
shagunsodhani: Got it :)

plugins/hydra_ax_sweeper/example/conf/config.yaml
shagunsodhani: Tacking in https://github.com/facebookresearch/hydra/issues/437

noxfile.py
shagunsodhani: Fixed in https://github.com/facebookresearch/hydra/pull/434/commits/f1a5692a97a3f5a3a172f48f27e3bfaba7f1f94a

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/ax_sweeper.py
shagunsodhani: Fixed in https://github.com/facebookresearch/hydra/pull/434/commits/3ea96d5c6bfb6bae377d35448e4c1d133a96d1ce

plugins/hydra_ax_sweeper/example/conf/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+  # To be minimized
+  x: ???
+  y: ???
+  
+
+hydra:
+  ax:
+    experiment:
+      # Banana function should be minimized (Ax defaults to maximize)
+      minimize: true
```
omry: I think we should default to minimize like nevergrad.
it makes more sense for ML.
Can be a in followup diff standardizing the two sweepers.

plugins/hydra_ax_sweeper/example/conf/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+# Banana function
+# https://en.wikipedia.org/wiki/Rosenbrock_function
+banana:
+  # To be minimized
+  x: ???
+  y: ???
+  
+
+hydra:
+  ax:
```
omry: Can you confirm indicate that the hydra.ax node is optional?
(Test it and add a comment saying it).

plugins/hydra_ax_sweeper/tests/config/quadratic.yaml
```diff
@@ -0,0 +1,36 @@
...
+      quadratic.x:
+        type: range
+        bounds: [-1, 1]
+        log_scale: false
+      
+      quadratic.y:
+        type: range
+        bounds: [-1, 1]
+        log_scale: false
+
```
omry: this is showcasing why it's good to make it optional.
There is a lot of unneeded boilerplate here.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/conf/hydra/sweeper/ax.yaml
shagunsodhani: Fixed via https://github.com/facebookresearch/hydra/pull/434/commits/f385339a2228783ac59ff89a1c5de574b3fcc242

plugins/hydra_ax_sweeper/example/conf/config.yaml
shagunsodhani: Changed the default to min via https://github.com/facebookresearch/hydra/pull/434/commits/ceac692d0027b76f615f1b349418eafba3d62dc6

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/conf/hydra/sweeper/ax.yaml
```diff
@@ -20,7 +20,7 @@ hydra:
       # Minimize objective:
       #   true: minimize
       #   false: maximize
-      minimize: false
+      minimize: true
```
omry: update the docs.

website/docs/plugins/ax_sweeper.md
```diff
@@ -46,7 +46,7 @@ The parameters for the optimization process can also be set in the config file.
 ```
 ax:
     experiment:
-      # If the function should be minimized (Ax defaults to maximize)
+      # If the function should be minimized (Plugin defaults to minimize)
       minimize: true
```
omry: if the plugin defaults to minimize this line is not needed right?

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/conf/hydra/sweeper/ax.yaml
shagunsodhani: https://github.com/facebookresearch/hydra/pull/434/commits/183d62ec8b20384c3bffb638e912fb4a0d421e66

plugins/hydra_ax_sweeper/example/conf/config.yaml
omry: This is the canonical example.
Add a comment saying the hydra.ax node is optional otherwise everyone will always copy paste it.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/ax_sweeper.py
```diff
@@ -222,9 +232,15 @@ def setup_ax_client(self, arguments: List[str]) -> AxClient:
                     break
             else:
                 parameters.append(cmd_param)
+
+        log.info(
+            f"AxSweeper is optimizing the following parameters: {encoder_parameters_into_string(parameters)}"
```
omry: What does the output look like?

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/ax_sweeper.py
```diff
@@ -103,14 +114,13 @@ class AxSweeper(Sweeper):
...
-        self, verbose_logging: bool, ax_config: DictConfig, random_seed: int,
+        self, ax_config: DictConfig,
     ):
         self.launcher: Optional[Launcher] = None
         self.job_results = None
         self.experiment = ax_config.experiment
         self.early_stopper = EarlyStopper(**ax_config.early_stop)
-        self.verbose_logging = verbose_logging
-        self.random_seed = random_seed
+        self.ax_client_config = ax_config.client
```
omry: might as well just store the whole config:
```
self.ax_config = ax_config
```
and use things from it directly.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/ax_sweeper.py
shagunsodhani:
```
[2020-02-24 12:04:19,842][HYDRA] AxSweeper is optimizing the following parameters: 

        type = range
        bounds = [-5, 5]
        log_scale = False
        name = banana.x
        parameter_type = ParameterType.INT

        type = range
        bounds = [-5.0, 10.1]
        log_scale = False
        name = banana.y
        parameter_type = ParameterType.FLOAT
```

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/ax_sweeper.py
shagunsodhani: I was making the variables for different scopes (in the config). Each scope (say `experiment`, `client` etc) corresponds to one component in the system.

website/docs/plugins/ax_sweeper.md
shagunsodhani: Fixed :)

plugins/hydra_ax_sweeper/example/conf/config.yaml
shagunsodhani: How does this sound: `# The following part of config is used to setup the Hydra Ax plugin and is optional`

omry: A bit wordy :)

omry:
```yaml
hydra:
  # optional Ax specific configuration
  ax:
    ..
```

omry:
rebase and rerun tests. I worked around a new issue with Windows on master.

plugins/hydra_ax_sweeper/tests/apps/quadratic.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - hydra/sweeper: ax
+
+quadratic:
+  # To be minimized
+  x: ???
+  y: ???
+
+hydra:
+  ax:
+
```
omry: remove newline after ax:

plugins/hydra_ax_sweeper/tests/config/quadratic.yaml
```diff
@@ -0,0 +1,31 @@
+defaults:
```
omry: if this is exactly the same as plugins/hydra_ax_sweeper/tests/apps/quadratic.yaml can you reuse?

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -0,0 +1,166 @@
...
+        "plugins/hydra_ax_sweeper/tests/apps/quadratic.py",
+        "-m",
+        "hydra.run.dir=" + str(tmpdir),
+        "quadratic.x=-5:-2",
+        "quadratic.y=-1:1",
+        "hydra.ax.max_trials=3",
+    ]
+    result = subprocess.check_output(cmd).decode("utf-8").rstrip()
+    assert "quadratic.x: range=[-5, -2], type = int" in result
+    assert "quadratic.y: range=[-1, 1], type = int" in result
```
omry: great.
Can you cover other types as well?
choice and fixed?

plugins/hydra_ax_sweeper/tests/config/quadratic.yaml
shagunsodhani: Updated

omry:
Fun never ends, you have mac issue.
restarting the run to see if it persists.

.coveragerc
```diff
@@ -2,6 +2,7 @@
 omit = 
     .nox/*
     *tests*
+    plugins/hydra_ax_sweeper/*
```
omry: If you disabled it for plugins, go all the way.

```suggestion
    plugins/*
```

shagunsodhani: Done

PR #450, opened 2020-02-27
fixing joblib config

omry:


plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/config.py
```diff
@@ -35,7 +35,7 @@ class JobLibConf:
     temp_folder: Optional[str] = None
 
     # thresholds size of arrays that triggers automated memmapping
-    max_nbytes: str = "1M"
+    max_nbytes: Optional[str] = None
```
omry: cc @jan-matthis

website/docs/plugins/joblib_launcher.md
```diff
@@ -21,43 +27,64 @@ defaults:
...
     # memmapping mode for numpy arrays passed to workers
-    mmap_mode: r
+    mmap_mode: str = "r"
 ```
 
-`n_jobs` defaults to -1 (all available CPUs). See [`Joblib.Parallel` documentation](https://joblib.readthedocs.io/en/latest/parallel.html) for full details on arguments. Note that the backend is hard-coded to use process-based parallelism (Joblib's loky backend), since thread-based parallelism is incompatible with Hydra.
+See [`Joblib.Parallel` documentation](https://joblib.readthedocs.io/en/latest/parallel.html) for full details about the parameters above.
+
+<div class="alert alert--info" role="alert">
+NOTE: The only supported JobLib backend is Loky (Processes bases parallelism).
```
jan-matthis: Small typo here, should be "process-based parallelism"

PR #452, opened 2020-02-28
joblib plugins params is now directly JobLibConf

omry:
Simplified joblib config


omry:
@jan-matthis, this is becoming quite elegant.

Issue #459, opened 2020-03-02
Unexpected structured config merger (should it be a type error?)


omry:
This allows re-typing of MySQLConf to PostGreSQLConfig (merging them together)

```
python examples/structured_configs_tutorial/4_config_groups/my_app.py db=postgresql
```

code:
```
from dataclasses import dataclass
from typing import Any

from omegaconf import MISSING, DictConfig

import hydra
from hydra.core.config_store import ConfigStore



@dataclass
class MySQLConfig:
    driver: str = "mysql"
    host: str = "localhost"
    port: int = 3306
    user: str = "omry"
    password: str = "secret"


@dataclass
class PostGreSQLConfig:
    driver: str = "postgresql"
    host: str = "localhost"
    port: int = 5432
    timeout: int = 10
    user: str = "postgre_user"
    password: str = "drowssap"


# Config is extending DictConfig to allow type safe access to the pretty() function below.
@dataclass
class Config(DictConfig):
    db: MySQLConfig = MySQLConfig()


cs = ConfigStore.instance()
cs.store(group="db", name="mysql", path="db", node=MySQLConfig)
cs.store(group="db", name="postgresql", path="db", node=PostGreSQLConfig)
cs.store(name="config", node=Config)


@hydra.main(config_name="config")
def my_app(cfg: Config) -> None:
    print(cfg.pretty())
    print(cfg.db._type)


if __name__ == "__main__":
    my_app()

```

PR #460, opened 2020-03-03
Specialized plugin configs now have to extend hydra.conf.PluginConf.

omry:


tests/test_config_loader.py
```diff
@@ -511,25 +511,35 @@ def test_load_schema_as_config(restore_singletons: Any) -> None:  # noqa: F811
...
-            foo: int = 10
+    params: FoobarParams = FoobarParams()
 
-        params: FoobarParams = FoobarParams()
 
-    @dataclass
-    class Config:
-        plugin: Plugin = Plugin()
+@dataclass
+# A plugin that does not extend from the parent Plugin class
```
omry:
```suggestion
# A plugin that does not extend the parent Plugin class
```

omry:
Closes #459

PR #465, opened 2020-03-03
req files

omry:
Introducing requirement files instead of relying on setup.py to hold requirements.


omry:
@shagunsodhani fyi, major cleanup of noxfile.

omry:
Skipped plugins are now printed as a clean warning:
Screenshot 2020-03-03 16 58 23

PR #498, opened 2020-03-18
Allow instantiation from class methods

dhpollack:
* instantiate an object from a class method of that object
* add tests for this
* simplify get_static_method function

Signed-off-by: David Pollack <d.pollack@solvemate.com>

## Motivation

I wanted to use hydra with the huggingface transfromers library and they extensively use a class method named `from_pretrained`.  

Currently one could do this by using the `get_static_method` and then manually converting the `params` key and passing them to the gotten method.  However that's not very straight forward or intuitive and this doesn't add much logic to the utility function.

Yes - passed tests in test_utils.py and flake8

## Test Plan

I added a test that tests the basic case.

### DIFF
--- a/hydra/utils.py
+++ b/hydra/utils.py
@@ -18,36 +18,21 @@ def get_method(path: str) -> type:
 
 def get_class(path: str) -> type:
     try:
-        from importlib import import_module
+        from pydoc import locate
 
-        module_path, _, class_name = path.rpartition(".")
-        mod = import_module(module_path)
-        try:
-            klass: type = getattr(mod, class_name)
-        except AttributeError:
-            raise ImportError(
-                "Class {} is not in module {}".format(class_name, module_path)
-            )
+        klass = locate(path)
+        if not klass:
+            log.error(f"Error finding module in class path {path}")
+            raise ValueError(f"Error finding module in class path {path}")
         return klass
-    except ValueError as e:
-        log.error("Error initializing class " + path)
-        raise e
-
-
-def get_class_method(clazz: Type[Any], method_name: str) -> type:
-    try:
-        clazz_method = getattr(clazz, method_name)
-        return clazz_method
     except Exception as e:
-        log.error("Error getting method {} from class {} : {}".format(clazz, method_name, e))
+        log.error(f"Error initializing class {path}")
         raise e
 
 
 def get_static_method(full_method_name: str) -> type:
     try:
-        class_name, method_name = full_method_name.rsplit(".", 1)
-        clz = get_class(class_name)
-        ret: type = getattr(clz, method_name)
+        ret: type = get_class(full_method_name)
         return ret
     except Exception as e:
         log.error("Error getting static method {} : {}".format(full_method_name, e))
@@ -56,11 +41,8 @@ def get_static_method(full_method_name: str) -> type:
 
 def instantiate(config: PluginConf, *args: Any, **kwargs: Any) -> Any:
     classname = _get_class_name(config)
-    methodname = _get_method_name(config)
     try:
         clazz = get_class(classname)
-        if methodname:
-            clazz = get_class_method(clazz, methodname)
         return _instantiate_class(clazz, config, *args, **kwargs)
     except Exception as e:
         log.error(f"Error instantiating '{classname}' : {e}")
@@ -107,13 +89,6 @@ def _get_class_name(config: PluginConf) -> str:
             return config.cls
         else:
             raise ValueError("Input config does not have a cls field")
-
-
-def _get_method_name(config: PluginConf) -> str:
-    if "method" in config:
-        return config.method
-    else:
-        return ""
 
 
 def _instantiate_class(
--- a/hydra/utils.py
+++ b/hydra/utils.py
@@ -2,7 +2,8 @@
 import logging.config
 import warnings
 from pathlib import Path
-from typing import Any, Type
+from typing import Any, Type, Optional, cast
+from types import ModuleType
 
 from omegaconf import DictConfig, OmegaConf, _utils
 
@@ -12,18 +13,66 @@ from hydra.core.hydra_config import HydraConfig
 log = logging.getLogger(__name__)
 
 
+def _safeimport(path: str) -> Optional[ModuleType]:
+    """
+    Import a module; handle errors; return None if the module isn't found.
+    This is a typed simplified version of the `pydoc` function `safeimport`.
+    """
+    import sys
+    from importlib import import_module
+    try:
+        module = import_module(path)
+    except ImportError as e:
+        if e.name == path:
+            return None
+        else:
+            log.error(f"Error importing module: {path}")
+            raise e
+    except Exception as e:
+        log.error(f"Non-ImportError while importing module {path}: {e}")
+        raise ValueError(f"Non-ImportError while importing module {path}: {e}")
+    for part in path.replace(module.__name__, "").split('.'):
+        if not hasattr(module, part):
+            break
+        module = getattr(module, part)
+    return module
+
+
+def _locate(path: str) -> ModuleType:
+    """
+    Locate an object by name or dotted path, importing as necessary.
+    This is similar to the pydoc function `locate`, except that it checks for
+    the module from the end of the path to the beginning.
+    """
+    parts = [part for part in path.split('.') if part]
+    module = None
+    for n in reversed(range(len(parts))):
+        try:
+            module = _safeimport('.'.join(parts[:n]))
+        except:
+            continue
+        if module:
+            break
+    if module:
+        obj = module
+    else:
+        log.error(f"Module not found: {path}")
+        raise ValueError(f"Module not found: {path}")
+    for part in parts[n:]:
+        if not hasattr(obj, part):
+            log.error(f"Error finding attribute ({part}) in class ({obj.__name__}): {path}")
+            raise ValueError(f"Error finding attribute ({part}) in class ({obj.__name__}): {path}")
+        obj = getattr(obj, part)
+    return obj
+
+
 def get_method(path: str) -> type:
     return get_class(path)
 
 
 def get_class(path: str) -> type:
     try:
-        from pydoc import locate
-
-        klass = locate(path)
-        if not klass:
-            log.error(f"Error finding module in class path {path}")
-            raise ValueError(f"Error finding module in class path {path}")
+        klass = cast(type, _locate(path))
         return klass
     except Exception as e:
         log.error(f"Error initializing class {path}")
@@ -32,7 +81,7 @@ def get_class(path: str) -> type:
 
 def get_static_method(full_method_name: str) -> type:
     try:
-        ret: type = get_class(full_method_name)
+        ret = get_class(full_method_name)
         return ret
     except Exception as e:
         log.error(f"Error getting static method {full_method_name} : {e}")
@@ -108,7 +157,7 @@ def _instantiate_class(
     params = config.params if "params" in config else OmegaConf.create()
     assert isinstance(
         params, DictConfig
-    ), f"Input config params are expected to be a mapping, found {type(config.params)}"
+    ), f"Input config params are expected to be a mapping, found {type(config.params).__name__}"
     primitives = {}
     rest = {}
     for k, v in kwargs.items():
--- a/hydra/utils.py
+++ b/hydra/utils.py
@@ -2,8 +2,8 @@
 import logging.config
 import warnings
 from pathlib import Path
-from typing import Any, Type, Optional, cast
 from types import ModuleType
+from typing import Any, Optional, Type, cast
 
 from omegaconf import DictConfig, OmegaConf, _utils
 
@@ -18,8 +18,8 @@ def _safeimport(path: str) -> Optional[ModuleType]:
     Import a module; handle errors; return None if the module isn't found.
     This is a typed simplified version of the `pydoc` function `safeimport`.
     """
-    import sys
     from importlib import import_module
+
     try:
         module = import_module(path)
     except ImportError as e:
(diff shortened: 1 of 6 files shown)


tests/test_utils.py
```diff
@@ -118,6 +124,16 @@ def test_get_static_method(path: str, return_value: Any) -> None:
             {"a": 10, "d": 40},
             Bar(10, 200, 200, 40),
         ),
+        (
+            {
+                "cls": "tests.test_utils.Baz",
+                "method": "class_method",
```
omry: The syntax I would like to support is:
```yaml
foo:
   cls: tests.test_utils.Baz.class_method
```
Is this implementable?
You may need to differentiate which of the two modes we are in automatically.

dhpollack: Yea, I wanted that too, but it didn't seem easy if you use the `instantiate` method because you would have to always check the last and second to last items to see if the last item in the string was a method or a class of the method of the second to last item.  I thought it would create extra logic that made things more confusing.  My other idea was to create a `instantiate_from_class_method` function (or whatever you'd want it to be called) which assumes that form.  Should work either way, it just depends where you want to put the logic.  

I'll take a look at it and see if I can put it all into the `cls` key neatly.

dhpollack: Yea, I wanted that too, but it didn't seem easy if you use the `instantiate` method because you would have to always check the last and second to last items to see if the last item in the string was a method or a class of the method of the second to last item.  I thought it would create extra logic that made things more confusing.  My other idea was to create a `instantiate_from_class_method` function (or whatever you'd want it to be called) which assumes that form.  Should work either way, it just depends where you want to put the logic.  

I'll take a look at it and see if I can put it all into the `cls` key neatly.

omry: Yeah, the implementation would be more complicated but I think it's okay.
Let's see what it looks like.

By the way, I just realized that PluginConf is now a part of the public API. I originally thought of it as an internal Hydra detail. I will have to promote it to be a part of the API. (probably also rename it).

dhpollack: @omry alright, it now works.  I used the the `locate` from from `pydoc` which basically does all the work.  I also simplified the other parts of the code that were duplicating functionality of this such as your static method function.

hydra/utils.py
```diff
@@ -18,32 +18,24 @@ def get_method(path: str) -> type:
...
-        mod = import_module(module_path)
-        try:
-            klass: type = getattr(mod, class_name)
-        except AttributeError:
-            raise ImportError(
-                "Class {} is not in module {}".format(class_name, module_path)
-            )
+        from pydoc import locate
+
+        klass = locate(path)
```
omry: This can't be implemented with importlib?

hydra/utils.py
```diff
@@ -116,9 +108,7 @@ def _instantiate_class(
     params = config.params if "params" in config else OmegaConf.create()
     assert isinstance(
         params, DictConfig
-    ), "Input config params are expected to be a mapping, found {}".format(
-        type(config.params)
-    )
+    ), f"Input config params are expected to be a mapping, found {type(config.params)}"
```
omry:
```suggestion
    ), f"Input config params are expected to be a mapping, found {type(config.params).__name__}"
```

tests/test_utils.py
```diff
@@ -59,6 +59,12 @@ def __eq__(self, other: Any) -> Any:
         return False
 
 
+class Baz(Foo):
+    @classmethod
```
omry: can you test it on `a @staticmethod ` as well?

website/docs/patterns/objects.md
```diff
@@ -49,7 +53,8 @@ conf/
 ├── config.yaml
 └── db
     ├── mysql.yaml
-    └── postgresql.yaml
+    ├── postgresql.yaml
+    └── postgresql_default.yaml
```
omry: I think we don't need a complete example.
It's enough to say in the main doc that this supports:
 - classes
 - class methods
 - static methods
 - (and if you test) static class methods.

omry:
As for the failing CI:
see the developer guide for info on how to run the tests and the lint locally.

website/docs/patterns/objects.md
dhpollack: Ok, where in the main doc is this?  I can add that to this PR.

hydra/utils.py
dhpollack: Under the hood it is implemented with `__import__` but with a bunch of logic to get the class methods and to recursive import the correct module.  I believe `importlib` is just a friendly wrapper around `__import__`.  You can see the code for [locate](https://github.com/python/cpython/blob/master/Lib/pydoc.py#L1620) and [safeimport (the underlying importer)](https://github.com/python/cpython/blob/master/Lib/pydoc.py#L318).  One could implement this with `importlib`. but it's not going to be as good as this implementation.

dhpollack: I ended up re-implementing and simplifying the pydoc function because there was some strange behavior if a submodule on the chain doesn't load.

tests/test_utils.py
dhpollack: done

hydra/utils.py
```diff
@@ -12,38 +13,82 @@
...
+    Import a module; handle errors; return None if the module isn't found.
+    This is a typed simplified version of the `pydoc` function `safeimport`.
+    """
+    from importlib import import_module
+
+    try:
+        module = import_module(path)
+    except ImportError as e:
+        if e.name == path:
+            return None
```
omry: what case is this handling?

hydra/utils.py
```diff
@@ -12,38 +13,82 @@
...
+        else:
+            log.error(f"Error importing module: {path}")
+            raise e
+    except Exception as e:
+        log.error(f"Non-ImportError while importing module {path}: {e}")
+        raise ValueError(f"Non-ImportError while importing module {path}: {e}")
```
omry: the log.error is probably going to cause the message to be printed twice.

hydra/utils.py
```diff
@@ -12,38 +13,82 @@
...
-
-        module_path, _, class_name = path.rpartition(".")
-        mod = import_module(module_path)
-        try:
-            klass: type = getattr(mod, class_name)
-        except AttributeError:
-            raise ImportError(
-                "Class {} is not in module {}".format(class_name, module_path)
-            )
+        klass = cast(type, _locate(path))
```
omry: try : 

```python
t = _locate(path)
assert isinstance(t, type)
return t
```

hydra/utils.py
```diff
@@ -12,38 +13,82 @@
...
         return klass
-    except ValueError as e:
-        log.error("Error initializing class " + path)
+    except Exception as e:
+        log.error(f"Error initializing class {path}")
```
omry: try:
```python
log.error(f"Error initializing class {path} : {e}")
```

website/docs/patterns/objects.md
```diff
@@ -33,6 +33,10 @@ class PostgreSQLConnection(DBConnection):
         self.user = user
         self.password = password
         self.database = database
+    
+    @classmethod
+    def default_init_method(self, database):
+        return self("localhost", "admin", "password", database)
```
omry: I think this can be removed now.

omry: I propose that you add a new section to this page describing `hydra.utils.instantiate()`.
It's probably should come near the end, after we introduce it for the first time.

website/docs/patterns/objects.md
omry: Does not exist yet, see my comment. we can add a new section to objects.md for now.

hydra/utils.py
dhpollack: You're right. I took it from the original file but I've noticed you've only used the log for BaseException errors that don't have a message attached.

omry:
ping

hydra/utils.py
dhpollack: I tried this but it didn't work.  The return types of static methods and class methods were different than class objects.  I tried explicitly typing `t` with `t: type`, but assert statement still failed.  I also tried changing the assert statement to `Callable` instead of `type`, but then I recieved a failure from the mypy check.  In this case, all the tests passed.

I also tried changing all of the return types to `ModuleType` and adjust the function signature of `_instantiate_class` accordingly, but this caused more mypy errors and some of the tests started failing as well.

hydra/utils.py
dhpollack: this was a relic from the pydoc method which checks the method from front to back, which is required if one uses `__import__`.  I removed it now.

dhpollack:
The test failues look like they are due to the upgrade in omegaconf.  Do you prefer a `git rebase` or a `git merge` on the master?  I tested both on branches locally.  I would assume rebasing, but I figured I'd ask.

omry:
I prefer rebase.

hydra/utils.py
omry: _locate signature suggests that the return type is ModuleType.
Since this function is now used for more than returning classes, maybe we need to refactor things such that it only get called for actual classes (by the user) and the internal implementation operates on ModuleType.

Does it make sense?
btw, can you join the chat?

website/docs/patterns/objects.md
```diff
@@ -95,3 +95,7 @@ Change the instantiated object class and override values from the command line:
 $ python my_app.py db=postgresql db.params.password=abcde
 PostgreSQL connecting to localhost with user=root and password=abcde and database=tutorial
 ```
+
+In addition to instantiating objects from classes, the `instantiate` method can also
+be used for class and static methods.  Simply put the path to the method in the `cls`
+field and use `params` if needed and instatiate your object exactly like a normal class.
```
omry: Please describe each of the use cases in more details.
Something along those lines (suggestion, feel free to diverge):

#### class names
An object of that class will be instantiated and returned. passing the config arguments to the `__init__` method
Example class:
```python
class Foo:
  def __init__(x: int, y: int) -> None:
    self.x = x
    self.y = y
```
Example usage:
config:
```yaml
object:
  cls: Foo
  params:
    x: 10
    y: 20
```
```python
result = hydra.utils.instantiate(cfg.object)
```

#### static_method
TODO: similar small example

#### class method
TOOD: similar small example

tests/test_utils.py
```diff
@@ -59,6 +59,16 @@ def __eq__(self, other: Any) -> Any:
...
+    def class_method(self, y: int) -> Any:
+        return self(y + 1)
+
+    @staticmethod
+    def static_method() -> Any:
```
omry: pass some parameters to the static method as well.

hydra/utils.py
dhpollack: I am in Europe so I can join a chat today (for me, probably tomorrow for you) if you're free.  I just registered for the chat, so I'll be in there.

hydra/utils.py
```diff
@@ -133,5 +162,25 @@ def _instantiate_class(
...
+) -> Any:
+    final_kwargs = _get_kwargs(config, **kwargs)
+    return fn(*args, **final_kwargs)
+
+
+# aliases
+instantiate = call
+get_static_method = get_fn_or_method
+get_method = get_fn_or_method
+_get_class_name = _get_cls_name
```
omry: There is only one usage of `_get_class_name()` in Hydra outside of this class. just fix it. no need to introduce an alias.

website/docs/patterns/objects.md
```diff
@@ -95,3 +95,62 @@ Change the instantiated object class and override values from the command line:
...
+import hydra
+
+@hydra.main(config_path="config.yaml")
+def app(cfg):
+  myobject = hydra.utils.call(cfg.myobject)
+  myclassmethod = hydra.utils.call(cfg.myclassmethod)
+  mystaticmethod = hydra.utils.call(cfg.mystaticmethod)
+  myfunction = hydra.utils.call(cfg.myfunction)
+```
+Note, the old `instantiate` function is now an alias of the `call` function used above.
```
omry:
```suggestion
Note: `hydra.utils.instantiate()` is an alias for `hydra.utils.call()`. They are in fact the same function.
```

dhpollack:
Alright, I think that I solved the mypy issues by using `_instantiate_class` for classes and `_call_callable` for functions and methods.  I also created a bunch of aliases for the old naming scheme and renamed functions as well.  I also added a few extra tests for builtins and nested methods.  

As for the documentation, let me know what you think.  I tried to keep it as compact as possible, but may be doing too many things with the same class.

P.S.  the win37 failure looks unrelated to this PR.  Looks like the conda servers glitched.

hydra/utils.py
dhpollack: :+1:

website/docs/patterns/objects.md
dhpollack: done

website/docs/patterns/objects.md
```diff
@@ -95,3 +95,63 @@ Change the instantiated object class and override values from the command line:
...
+Now to test these instantiate / call them as follows:
+```python
+import hydra
+
+@hydra.main(config_path="config.yaml")
+def app(cfg):
+  myobject = hydra.utils.call(cfg.myobject)
+  myclassmethod = hydra.utils.call(cfg.myclassmethod)
+  mystaticmethod = hydra.utils.call(cfg.mystaticmethod)
+  myfunction = hydra.utils.call(cfg.myfunction)
```
omry:
```suggestion
  foo1: Foo = hydra.utils.call(cfg.myobject)
  foo2: Foo = hydra.utils.call(cfg.myclassmethod)
  ret1: int = hydra.utils.call(cfg.mystaticmethod)
  ret2: int = hydra.utils.call(cfg.myfunction)
```

omry:
As for the docs, they look good here but I would like to see them in the website.
Did you start the website locally to test that they look okay?

dhpollack:
I didn't but I'll check it tomorrow

omry:
Awesome. edited the doc a bit and added the issue #505 for API cleanup before 1.0.

omry:
Ohh, forgot.
Can you add a news fragment file?
see the developer guide for details.

omry:
Awesome, thanks David!

PR #502, opened 2020-03-25
typo in config file prevented AxSweeper from loading correctly

dhpollack:
this is a super mini PR.  The config file points to the wrong location.  I bashed my head against the wall for a while trying to figure out what was going wrong.

### DIFF
--- a/plugins/hydra_ax_sweeper/example/conf/config.yaml
+++ b/plugins/hydra_ax_sweeper/example/conf/config.yaml
@@ -10,7 +10,7 @@ banana:
 
 hydra:
   sweeper:
-    cls: hydra_plugins.hydra_ax_sweeper.AxSweeper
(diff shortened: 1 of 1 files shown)


omry:
@shagunsodhani, we need a test that covers the example app properly.

omry:
Thanks David!

dhpollack:
Also, it wasn't completely obvious to me, but you could remove that line entirely as long as you have the defaults line loading the sweeper.  It makes sense to override some of the default settings, but it doesn't really ever make sense to override the class itself.

omry:
> Also, it wasn't completely obvious to me, but you could remove that line entirely as long as you have the defaults line loading the sweeper. It makes sense to override some of the default settings, but it doesn't really ever make sense to override the class itself.


Yup, I think there is a standing PR for something else that removes that line already.

shagunsodhani:
Hey @dhpollack Thank you for fixing this :)

PR #523, opened 2020-04-11
refactoring: early return 'if else' syntax -> 'if' syntax

marload:
## Motivation

Personally, I think if is better than if else in the early return pattern.

Yes

## Test Plan

## Related Issues and PRs


omry:
Thanks, for the suggestion.
I personally think that in this case it makes sense to have the code structured as it as because there are two clear cases we are dealing with.

PR #528, opened 2020-04-12
Lint cleanup

### LINKED ISSUE #397: Clean up .isort.cfg
master isort.cfg file currently need to contain all third party needed by plugins.
this can be improved by utilizing per project isort config.

### LINKED ISSUE #436: Create a nox target to lint plugins
Plugins should be linted after they are installed (with their dependencies).
Current plugin linting logic does not do it.

As this will be slower, separate to a different nox session.

remove `#type ignore` from the `Nevergrad plugin` once this is enabled.

omry:
Fixes #436
Fixes #397

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@ __pycache__
 /.nox
 /report.json
 /.coverage
-/.mypy_cache
+.mypy_cache
 pip-wheel-metadata
 .ipynb_checkpoints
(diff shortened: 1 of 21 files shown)


omry:
cc @shagunsodhani 
FYI: this may break your linting, let me know if you run into problems after you rebase (will land this soon).

Issue #468, opened 2020-03-04
Specifying output type (via cmd) for Ax/Nevergrad Plugins


shagunsodhani:
Related to #430

jrapin:
Following hydra example and simplified for my use case, the file currently looks like this with nevergrad:
```yaml
nevergrad:
  batch_size: 4
  db: mnist
  dropout: 0.381
  lr: 0.094

optimizer: nevergrad
```


I'm fine with @omry proposition:
```python
@dataclass
class SweeperResult
  name : str = MISSING
  best_parameters: Dict[str, Any] = MISSING # optimal parameters found
  best_achieved_result: float = MISSING     # optimal return value obtained
  sweeper_cfg : Any = MISSING               # The hydra.sweeper.params node
```
I would tend to rename `best_parameters` to `recommendation` because that fits better with nevergrad, but anything will do.

A small detail though: sometimes nevergrad (probably the same with botorch?) recommendation has never been evaluated, so the best achieved result may not be available, or at least not be directly related to the recommendation. Another option is to always recommend an evaluated set of parameters. I think this seems more logical in this use case so I would fo for it, but that would however require some more lines of codes (nothing dramatic though :D)

omry:
Thanks @jrapin.
We have been running into those exact considerations with Ax and decided for now to ignore the predicted result and just print and store the result of the best evaluation.

With Nevergrad, do you ever see a significant improvement between the best result and the un-evalulated final recommendation?

jrapin:
I know that my colleague Olivier works on such cases :D, this happens mostly on noisy functions, the mean of the best points in the optimizer population of points is usually a much better estimate of the solution. I doubt it would apply to cases that are relevant for hydra.

As long as we can eventually **add** another field I don't think we need to care for that, most users will probably want to see the best run only, if some need more information we can deal with this later

omry:
No cases are relevant to Hydra, it's general purpose.
In fact RL training, which often uses Hydra is very noisy.

I think we can already add both predicted best numbers and predicted best result.

jrapin:
I would tend to go for something along those lines (slightly different names and structure), but anything should work for me:

```
name  # eg: nevergrad?
best_evaluation
   value: xxx
   parameters:
      xxxx
      xxxx
recommendation
   xxxx
   xxxx
sweeper_config  # not sure what that is supposed to be
```

omry:
[...]

Issue #437, opened 2020-02-22
Specifying parameter type (via cmd) for Ax/Nevergrad Plugins


shagunsodhani:
Determine how to specify the optimization type per parameters via command-line  for Ax/Nevergrad plugins. (float? int?).

Related to https://github.com/facebookresearch/hydra/issues/302 and https://github.com/facebookresearch/hydra/issues/330

omry:
Let's try to agree on standard command line and configuration structure for the variable definitions for both the Ax and Nevergrad plugins.

shagunsodhani:
Linking to the relevant discussion https://github.com/facebookresearch/hydra/pull/430#discussion_r389033167

shagunsodhani:
I am summarising the onversation so far. In case I missed something relevant, please add it here (the discussion was a little scattered).

Main points to discuss:

* Should cmd always have preference over config? The current consensus seems to be that anything explicitly defined in the cmd should win over config. 
* Should we infer the type from the cmd or require it to be specified explicitly. (ie `3:5` vs `int:3:5`)

@jrapin suggested that we support only range via the cmd for now: 
```
# bounds (min:max)
0:1           # specs are optional, assumed float if int not provided
int:0:1
log:0.001:1 (log 0 somehow means unbounded...)
log:int:0.001:1
int:log:0.001:1 # order of specs does not matter

anything,whatever,something
```

I would build on @jrapin 's suggestion (and include some examples of choice and fixed parameters since Ax supports them currently).

Range Parameters

| Input           | Interpretation                              | Explanation                           |
|-----------------|---------------------------------------------|----------------------------------------|
| 0:1             | float(0):float(1)                           | Numbers are float by default           |
| 0:1.1           | float(0):float(1.1)                         | Numbers are float by default           |
| int:0:1         | int(0):int(1)                               | Numbers are typecasted to int          |
| int:0:1.0       | int(0):int(1)                               | Numbers are typecasted to int          |
| int:0:1.2       | Error                                       | float to int casting loses information |
| log:1:2         | float(1):float(2) Sampling in the log space | Numbers are float by default           |
| log:float:1:2   | float(1):float(2) Sampling in the log space | Numbers are float by default           |
| float:log:1:2   | float(1):float(2) Sampling in the log space | Order of log and float does not matter |
| int:log:1:2     | int(1): int(2) Sampling in the log space    | Numbers are typecasted to int          |
[...]

PR #531, opened 2020-04-14
Update nevergrad parsing and output

jrapin:
## Motivation

- compliant parsing #437 
- more standard output #468

Yes

## Test Plan

Tests are included


plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -24,41 +26,104 @@ def manipulate_search_path(self, search_path: ConfigSearchPath) -> None:
         )
 
 
-def convert_to_deduced_type(string: str) -> Union[int, float, str]:
-    """Converts a string into a float or an int if it makes sense,
-    or returns the string.
+@dataclass
+class CommandlineSpec:
```
jrapin, line 30: This part could be moved somewhere else to share with other plugins

omry: cc @shagunsodhani

shagunsodhani: Lets put it in a separate sub dir in the plugins dir (as discussed previously)

jrapin:
> I was thinking about taking it a step further and consider the variable spec in the config as well.

Not sure what you mean here.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jrapin: I didn't get where that would be and how that would be imported that's why i had not so far, I can do it if you explain further though.
In the meantime I moved it to hydra/plugins/sweeper.py which seemed a good location

omry:
> > I was thinking about taking it a step further and consider the variable spec in the config as well.
> 
> Not sure what you mean here.

Given a variable parameterization the config, and given  a command override for the same parameter, what are we ending up with?

if config says  x is a float, and cli says 1, are we ending up with float?
if config says x is an int and we have "foo", we we ending up with an error? can this error be consistent across the plugins of this family?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
omry: This is not something that is relevant to all sweeper plugins, so I would rather keep it outside of the core.
it is also likely to evolve faster than the Hydra core as we start getting feedback about the nevergrad and ax plugins.

I propose:
plugins/common/hydra-hpo-core

which can be a regular pip package (not a plugin).
(hpo stands for hyper parameter optimization, open to other names).

shagunsodhani: I do not have strong opinions but I find `hydra-opt-core` to more precise than `hydra-hpo-core`

jrapin:
> Given a variable parameterization the config, and given a command override for the same parameter, what are we ending up with?

The commandline completely bypasses the config file


> if config says x is a float, and cli says 1, are we ending up with float?

yes, but only because the cli defaults to float

> if config says x is an int and we have "foo", we we ending up with an error?

no, the cli overrides the config

> can this error be consistent across the plugins of this family?

`ParameterSpec` only reads the parameter specification and puts structure to it, there is no logic of combining with information from anywhere else, but at least it's closer to the spec we agreed on.

Adding the extra consistency may be a pain and beyond this PR, especially since parameters can be expressed differently in the configuration.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jrapin: I won't add another installable package with this PR, I wouldnt have time to dig into all the little details that would need to be handled for that

omry: > I do not have strong opinions but I find hydra-opt-core to more precise than hydra-hpo-core

likestamp

> I won't add another installable package with this PR, I wouldnt have time to dig into all the little details that would need to be handled for that

it shouldn't be hard. you can base your setup.py it on one of the example plugins.
don't worry about actually publishing it. I will.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -70,70 +165,74 @@ def make_parameter_from_commandline(string: str) -> Any:
...
+        except ValueError:
+            pass
+    # convert scalar commandline specs to dict
+    if isinstance(description, CommandlineSpec) and description.bounds is not None:
+        description = ScalarConfigSpec(
+            lower=description.bounds[0],
+            upper=description.bounds[1],
+            log=description.log,
+            integer=description.cast == "int",
         )
```
omry, line 204: can you explain what is happening here?

you seem to be construvting the ScalarConfigSpec from multiple input types.
What are the cases?

is there a case when you use the one provided by Hydra as is?
(the Dictconfig that has the underlying ScalarConfigSpec type)?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -24,41 +26,134 @@ def manipulate_search_path(self, search_path: ConfigSearchPath) -> None:
...
-    read_value("1.0")
-    >>> 1.0
+    bounds: Optional[Tuple[float, float]] = None
+    options: Optional[List[str]] = None
+    cast: str = "float"
+    log: bool = False
 
-    read_value("blublu")
-    >>> "blublu"
+    def __post_init__(self) -> None:
```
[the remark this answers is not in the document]
jrapin: This class is not directly used in the config. More on it below

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jrapin: The docstring has details on this.
Basically this method converts any input to nevergrad parameter. That means:
- strings from commandline (which can actually be used in the config file as well) which are first read with `CommandlineSpec`, and then converted to `ScalarConfigSpec` if need be (i.e. if it's a range variable). This is what you pointed at
- list in the config files, which are considered as categorical variables
- dict in the config file, which are cast to `ScalarConfigSpec` for safety, then converted to the actual nevergrad type.

It's messy because there are several ways to define the parameters, and several type of parameters. I don't see simpler solutions though, so I tested it as much as possible.

considering `ScalarConfigSpec`, it's not used in the config, because it's not the only option to describe the variable (remember the `Union`?). It's basically only there to clarify the structure, and serves to check that there are no extra parameters which should not be there. Does that make more sense?

omry: Yes.
Where I would like us to converge is something like:

`hydra_opt_core`:
- defines common config structure and command line interface
- Implement conversion of the command line and the config into an object describing the variable (ScalarConfigSpec?), handling logic edge cases like specifying type in config but populating values from the line).

Now each plugin can call a single method, pass in it's config and command lines into a shared function and get back a list of ScalarConfigSpec it can use in it's logic.

How does that sound?
(This is not about this current diff, but more about where I would like us to go).

omry:
Please rebase, there are some merge conflicts now.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jrapin: I know it's been what we discussed about. I agree with it but that will take a while. I think the CommandlineParser is a small step in that direction (although it may need minor adaptation for identifying if cast was implicit or explicit).
However in the discussions we have overlooked the configuration file spec, which is probably critical to this end.

`ScalarConfigSpec` is only an intermediary structure: it wouldnt work in the config, since it's only dealing with scalars, not with choice variables.
Another aspect is that different sweeper may need different information to define a variable. For scalars I would like users to be able to define a step size, but I think this would not make sense for ax.
So should the config handle all possible cases and each plugins ignored irrelevant info? That could become messy, and may lead to incompatibilities
Or should the config file differentiate between the parametrization of each sweeper? That would lead to more robust spec, but would require a lot of duplications

jrapin: Maybe to summarize: we focused on a spec for the commandline so far. Commandline being a bit constrained, it has forced us to support only minimal features that all sweepers will probably share. On the other hand, config goes the other way round, so there are several options.

omry: Those are all things we can discuss as we start morphing the code in that direction.
specifically to your question about step size:
I think it would make sense for a plugin to just tell the user that step size is not supported if it's set and the plugin does not accept it.

jrapin: > I think it would make sense for a plugin to just tell the user that step size is not supported if it's set and the plugin does not accept it.

Then what is the point of having the same config if in practice we can't share it from one sweeper to another?

omry: 1. Learning curve is shallower for switching between them.
2. Different sweepers will never be exactly the same, not in implementation and not in supported features set (and algorithms they can use). Having a unified config when possible will make switching between them easier.

A better decision might be to switch just issue a warning about the presence of the unsupported value and ignore it.

Issue #549, opened 2020-04-20
[Feature Request] Fish shell tab completion


binshengliu:
# 🚀 Feature Request

Adding completion support for [fish shell](https://fishshell.com/).

## Motivation

**Is your feature request related to a problem? Please describe.**
Nope. This just improves the user experience.

## Pitch

**Describe the solution you'd like**

The solution uses fish's [`complete`](https://fishshell.com/docs/current/#writing-your-own-completions) command to register completion rules. When the user presses TAB, fish will call the underlying script for candidates and do the completion.
```
$ ./test.py -sc install=fish | tee /dev/tty | source
complete -c python3 -n '__fish_seen_subcommand_from ./test.py' -x -a '(/home/bsl/.miniconda3/envs/hydra38/bin/python3 /home/bsl/code/tmp/test.py -sc query=fish --_completion (commandline -cp))'
complete -c test.py -x -a '(/home/bsl/.miniconda3/envs/hydra38/bin/python3 /home/bsl/code/tmp/test.py -sc query=fish --_completion (commandline -cp))'
$ ./test.py -sc uninstall=fish | tee /dev/tty | source
complete -e -c test.py
```
A new global option `--_completion` is added following an [old convention](https://stackoverflow.com/a/20879411/955952). The current command line will be fed into hydra through this option for completion instead of using environment variables. This also leads to the interface change of `CompletionPlugin.query`, but I think it should be reasonable.

**Describe alternatives you've considered**

NA

**Are you willing to open a pull request?** (See [CONTRIBUTING](../CONTRIBUTING.md))
I did some fiddling with the code. This [branch](https://github.com/binshengliu/hydra/tree/fish-completion) passes the lint checks but I'm not sure what's happening with the tests (on the master branch it fails at `test_completion.py`). I'm okay if you are interested and want to modify the code to your standards. Just hoping this is useful and there's an auto-completion solution for fish.

## Additional context
The effect from my branch.

![Fish completion](https://user-images.githubusercontent.com/441707/79732640-50c8d380-8337-11ea-9144-a27f1cab6932.gif)

There are still three problems I don't know the answers yet.

1. `python script.py` style completion can not be uninstalled from fish. The workaround is to open a new shell.
2. Fish automatically adds a space to args with a dot(`db.` -> `db. `) which requires the user to remove a space then press TAB for further completion. This seems to be a fish's internal limitation.
3. I have no idea how to auto test these features so sorry I can't contribute test cases.

[...]

PR #550, opened 2020-04-21
support fish-completion

binshengliu:
## Motivation

As discussed in #549, adding completion for fish shell.

Yes

## Test Plan

Similar to bash completion. Already integrated to `test_shell_integration` testcase. It requires fish shell to be installed.

## Related Issues and PRs

#549


hydra/_internal/core_plugins/fish_completion.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def query(self, config_name: Optional[str], line: Optional[str] = None) -> None:
+        if not line:
+            return
+        line = self.strip_python_or_app_name(line)
+        print("\n".join(self._query(config_name=config_name, line=line)))
+
+    @staticmethod
+    def _get_exec() -> List[Tuple[str, str, str]]:
+        # Running as an installed app (setuptools entry point)
```
omry, line 61: are you handling the case for python foo.py?

hydra/_internal/hydra.py
```diff
@@ -207,6 +210,8 @@ def format_args_help(args_parser: ArgumentParser) -> str:
             if len(action.option_strings) == 0:
                 overrides = action
             else:
+                if action.help == SUPPRESS:
+                    continue
```
omry, line 214: what's this?

hydra/_internal/utils.py
```diff
@@ -239,7 +239,9 @@ def run_hydra(
                 config_name=config_name, overrides=args.overrides, cfg_type=args.cfg
             )
         elif args.shell_completion:
-            hydra.shell_completion(config_name=config_name, overrides=args.overrides)
+            hydra.shell_completion(
```
omry, line 242: Can you explain the changes you are making here?

hydra/_internal/utils.py
```diff
@@ -290,6 +292,8 @@ def get_args_parser() -> argparse.ArgumentParser:
     shell = "SHELL_NAME"
     install_cmd = 'eval "$({} -sc install={})"'.format(_get_exec_command(), shell)
     uninstall_cmd = 'eval "$({} -sc uninstall={})"'.format(_get_exec_command(), shell)
+    fish_install_cmd = "{} -sc install=fish | source".format(_get_exec_command())
+    fish_uninstall_cmd = "{} -sc uninstall=fish | source".format(_get_exec_command())
```
omry, line 296: is the eval command above not working for fish?

tests/expect/test_fish_completion.exp
```diff
@@ -0,0 +1,50 @@
...
+set timeout 4
+
+# Under pycharm, for some unknown reason readline displays
+# the suggestions in a different order, breaking the test.
+# This enables debug output from Hydra which is easier to test
+set env(HYDRA_COMP_DEBUG) 1
+set env(INPUTRC) /dev/null
+
+# Suppress bash deprecation warning on Mac
+set env(BASH_SILENCE_DEPRECATION_WARNING) 1
```
omry, line 12: not needed if we are using fish.

tests/expect/test_fish_completion.exp
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+log_user 0
+set timeout 4
+
+# Under pycharm, for some unknown reason readline displays
+# the suggestions in a different order, breaking the test.
+# This enables debug output from Hydra which is easier to test
+set env(HYDRA_COMP_DEBUG) 1
+set env(INPUTRC) /dev/null
```
omry, line 9: since you don't have an equivalent shell function that handles HYDRA_COMP_DEBUG you don't need this.
same for INPUTRC, I think this is bash specific.

tests/test_completion.py
```diff
@@ -143,6 +154,9 @@ def test_shell_integration(
                 str(num_tabs),
             ]
         )
+        if shell == "fish":
+            num_tabs -= 1
+
```
omry, line 159: add a comment explaining this.

tests/test_completion.py
```diff
@@ -110,7 +121,7 @@ def test_completion_plugin(
     @pytest.mark.parametrize(  # type: ignore
         "prog", [["python", "hydra/test_utils/completion.py"]]
     )
-    @pytest.mark.parametrize("shell", ["bash"])  # type: ignore
+    @pytest.mark.parametrize("shell", get_supported_shells())  # type: ignore
```
omry, line 124: 1. test both, but [skip](http://doc.pytest.org/en/latest/skipping.html) if fish is not installed.
2. install fish on in the [circleci config](https://github.com/facebookresearch/hydra/blob/master/.circleci/config.yml) on Linux and Mac jobs.

hydra/_internal/core_plugins/bash_completion.py
```diff
@@ -86,7 +86,7 @@ def strip_python_or_app_name(line: str) -> str:
             else:
                 raise RuntimeError("Error parsing line '{}'".format(line))
 
-    def query(self, config_name: Optional[str]) -> None:
+    def query(self, config_name: Optional[str], line: Optional[str] = None) -> None:
```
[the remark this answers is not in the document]
binshengliu: This is because the base class signature changed. The subclass needs to match the signature to pass lint checks.

hydra/_internal/core_plugins/fish_completion.py
```diff
@@ -0,0 +1,81 @@
...
+        # User scenario 2: ./script.py or src/script.py or script.py
+        name = os.path.basename(sys.argv[0])
+        cond = ""
+        output.append((name, cond, exe))
+
+        return output
+
+    @staticmethod
+    def _get_uninstall_exec() -> str:
+        name = os.path.basename(sys.argv[0])
```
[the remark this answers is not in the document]
binshengliu: Nope. It only unregisters ./foo.py as I explained in another discussion. We only have methods to remove all the completion rules of `python` command or specific options. There's no way to unregister dynamic completion rules.

hydra/_internal/core_plugins/fish_completion.py
```diff
@@ -0,0 +1,81 @@
...
+        app_args = r"^\s*[\w_\-=\./]+\s*(.*)"
+        match = re.match(python_args, line)
+        if match:
+            return match.group(1)
+        else:
+            match = re.match(app_args, line)
+            if match:
+                return match.group(1)
+            else:
+                raise RuntimeError("Error parsing line '{}'".format(line))
```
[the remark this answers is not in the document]
binshengliu: I changed the regexp a bit to also match `python2` and `python3` instead of only `python` (just noticed python2 is not supported). I wasn't sure the consideration in the original version, but this change should be generic enough to be moved in the base class.

hydra/_internal/core_plugins/fish_completion.py
binshengliu: Yes that's handled. Fish supports `conditional` completion. `foo.py` is treated as a match condition.

hydra/_internal/hydra.py
binshengliu: The `--_completion` flag should be hidden from the user. [`SUPPRESS`](https://docs.python.org/3/library/argparse.html#help) is a flag to silence options in argparse.

hydra/_internal/utils.py
binshengliu: No. Fish's eval command only works for a single line. [fish discussion](https://github.com/fish-shell/fish-shell/issues/3993)

hydra/_internal/utils.py
binshengliu: A new command line option `--_completion` is added. For `bash`, it would be `None`. For fish, its value would be the current command line until where the user presses TAB, for example `python ./foo.py db` if the user presses TAB after `db`. This modification passed `args._completion` into specific shell plugins (`hydra.shell_completion() -> plugin.query(...)`).

binshengliu:
Can you show me where it fails? I think I ran all the tests before the pr. In tests/expect/test_fish_completion.exp I'm already simulating multiple-tab. That can also test the order of the completion.

tests/test_completion.py
binshengliu: I edited the circleci conf but I honestly have zero knowledge of mac. Let's see what happens.

.circleci/config.yml
```diff
@@ -87,6 +102,7 @@ jobs:
           name: "Preparing environment"
           command: |
             sudo apt-get install -y expect
+            sudo apt-get install -y fish
```
omry, line 105:
```suggestion
            sudo apt-get install -y expect fish
```

.circleci/config.yml
```diff
@@ -7,6 +7,11 @@ jobs:
       xcode: "10.0.0"
     steps:
       - checkout
+      - run:
+          name: "Installing Fish"
+          command: |
+            brew update
+            brew install fish
```
omry, line 14: can you fold the Fish installation into "Preparing Conda Environment" and rename it to "Preparing Environment"?

hydra/_internal/core_plugins/fish_completion.py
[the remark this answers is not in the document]
omry: can you point me to some information about dynamic completion in fish?

hydra/_internal/utils.py
omry: The evaluated expression is a single line, isn't it?
Does the `| source` approach works for bash?

hydra/_internal/utils.py
omry: Does fish support registering completion functions like bash completion?
If it does, we can have the adapting logic in that shell function (it can populate $COMP_LINE) and then we wouldn't need the different interface for completion.

hydra/_internal/core_plugins/fish_completion.py
```diff
@@ -0,0 +1,59 @@
...
+    """Issue 1: python script.py style rule cannot be uninstalled
+    Issue 2: fish adds a space to "hydra. " automatically"""
+
+    def install(self) -> None:
+        output = self._get_exec()
+        script = []
+        for name, cond, exe in output:
+            args = f"'({exe} -sc query=fish --_completion (commandline -cp))'"
+            script.append(f"""complete -c {name} {cond}-x -a {args}\n""")
+        print("".join(script))
```
omry, line 22: What does the install command looks like (actual output) ?

tests/test_completion.py
```diff
@@ -131,6 +124,8 @@ def test_shell_integration(
         line: str,
         expected: List[str],
     ) -> None:
+        if not is_fish_exists():
+            pytest.skip("fish should be installed to run the tests")
```
omry, line 128: I think you are skipping both bash and fish now if fish is not installed :)

hydra/_internal/core_plugins/fish_completion.py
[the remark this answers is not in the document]
binshengliu: The dynamic completion isn't explicitly discussed in the manual. Please see a discussion here https://github.com/fish-shell/fish-shell/issues/1217#issuecomment-31441757.

The add/erase mechanism of `complete` is also relevant to this problem. It can be found [here](https://fishshell.com/docs/current/cmds/complete.html) and [here](https://fishshell.com/docs/current/#tab-completion).

hydra/_internal/core_plugins/fish_completion.py
binshengliu: Somehow I can't find the reply box for your previous comment about registering completion function. I think it's technically doable to mimic what has been done for bash. I actually thought of reusing that solution, but fish is not POSIX compliant or bash compatible. It has a different scripting syntax and I think it makes the solution more complex and less maintainable.

Sorry I'm caught up in other things today and tmr. Will come back to address your other comments after.

hydra/_internal/core_plugins/fish_completion.py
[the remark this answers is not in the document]
omry: complete --help is pretty informative. looking now.

hydra/_internal/core_plugins/fish_completion.py
omry: we should definitely not try to reuse the same function from bash, it's super specific. but the concept of having a fish specific completion function installed seems good to me.
that function would be able to bridge the gap between what Hydra is expecting and what fish is asking for.

complete --help has this, it might be what I am after.
```
       complete -c rpm -n '__fish_contains_opt -s e erase' -d nodeps 'Don't check dependencies'

       where __fish_contains_opt is a function that checks the command line
       buffer for the presence of a specified set of options.
```

hydra/_internal/core_plugins/bash_completion.py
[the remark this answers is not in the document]
omry: Okay. if we can use a completion function we will be able to get it to populate the environment variable and then we wouldn't need the interface change.

omry:
I have been doing some digging, this is what I found:

uninstall all completion for myprog:
```
complete -e --command myprog
```
register hydra_complete command as the completion function for myprog:
```
complete --command myprog --arguments '(hydra_complete (commandline -cp))' --no-files
```
Hydra complete function (need to source this):
$ cat hydra-complete.fish 
```fish
function hydra_complete
        printf "a\nb\n"
        return 1
end
```

hydra/_internal/core_plugins/fish_completion.py
binshengliu: I'm not sure I understand what you mean by the help info.

hydra/_internal/utils.py
binshengliu: The expression was multiple lines. If we use the way like bash, it would contain more lines. Bash `source` doesn't support read from stdin so this approach doesn't work for it.

hydra/_internal/core_plugins/bash_completion.py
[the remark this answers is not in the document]
binshengliu: Now I pushed some changes to use COMP_LINE. Please see the latest branch.

binshengliu:
Now we have a similar approach to the bash solution.

```
$ python ./test.py -sc install=fish
function hydra_fish_completion
    # Hydra will access COMP_LINE to generate completion candidates
    set -lx COMP_LINE (commandline -cp)

    # Find out how to call the underlying script
    set -l parts (string split -n ' ' $COMP_LINE)
    if test "$parts[1]" = "python" -o "$parts[1]" = "python3"
        set cmd "$parts[1] $parts[2]"
    else
        set cmd "$parts[1]"
    end

    # Generate candidates
    eval "$cmd -sc query=fish"
end
        
complete -c python -n '__fish_seen_subcommand_from ./test.py' -x -a '(hydra_fish_completion)'
complete -c test.py -x -a '(hydra_fish_completion)'
```

omry:
Shaping up really nicely.
A few more and requests inline.
Also, I would like to ad a test that validates that we do not get a space after tab for a thing that ends with a `.`.
This can be challenging.
Until we actually have the next version of fish released we will need to have this test disabled, but I would like it in place.

omry:
Additional checklist items:
- Update docs
- Create a news fragment

hydra/_internal/core_plugins/fish_completion.py
```diff
@@ -0,0 +1,74 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class FishCompletion(CompletionPlugin):
+    """Issue 1: python script.py style rule cannot be uninstalled
+    Issue 2: fish adds a space to "hydra. " automatically"""
+
+    def install(self) -> None:
+        script = """function hydra_fish_completion
```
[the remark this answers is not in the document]
binshengliu: It completes only when the script name matches through `__fish_seen_subcommand_from`. But I think there is indeed a potential problem here if the user switches directories and call a different script with the same name.

Checking @hydra.main may not fully solve the problem if the two scripts both use hydra. It seems to me checking the absolute path would be the safest way.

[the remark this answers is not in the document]
omry: If the other script is a Hydra script you can definitely try to activate completion on it, this is fine.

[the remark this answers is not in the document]
binshengliu: The uninstalling command you showed would remove all the completions for `python`. The complete command supports uninstalling `-s` `-l` `-o` options according to its manual, neither of which is applicable in this case.

hydra/_internal/core_plugins/fish_completion.py
```diff
@@ -0,0 +1,74 @@
...
+    @staticmethod
+    def _get_exec() -> List[Tuple[str, str]]:
+        # Running as an installed app (setuptools entry point)
+        output = []
+        # User scenario 1: python script.py
+        name = os.path.basename(sys.executable)
+        cond = f"-n '__fish_seen_subcommand_from {sys.argv[0]}' "
+        output.append((name, cond))
+
+        # User scenario 2: ./script.py or src/script.py or script.py
```
omry, line 63: Please verify that this works for the scenario of an [installed application](https://hydra.cc/docs/next/advanced/app_packaging).
Test on the example application [here](https://github.com/facebookresearch/hydra/tree/master/examples/advanced/hydra_app_example).

hydra/_internal/core_plugins/fish_completion.py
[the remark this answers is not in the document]
omry: oh, I see.
Can you ask the Fish guys how to uninstall a specific completion function for a command?

If we can't do it it's not so bad as long as we still verify that the app contains @hydra.main().

binshengliu:
Sorry for the late response. I'm a bit inundated with my own projects lately.

I only had a chance to fiddle with some test cases. Exact line end matching of regexp could test against fish's extra space after the dot. The following diff works for fish with the dot fix. I think it's also more reliable and safer than matching any substring as in the current approach.

You can note some common differences from the diff.

1. A `$` is added to `expected` to match line ending using regexp.
2. Fish requires less tabs in some cases. Unfortunately this seems inconsistent across different versions.
3. Fish adds a space to a complete completion.

I didn't have time to integrate these differences yet, but a separate test function that explicitly tracks the differences looks better to me than a lot of `if` fixes. Also I think we can just test the latest upstream version to avoid making this too complex.

```diff
diff --git a/tests/test_completion.py b/tests/test_completion.py
index b0a17768..9ea68d1e 100644
--- a/tests/test_completion.py
+++ b/tests/test_completion.py
@@ -1,6 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 import distutils.spawn
 import os
+import re
 import subprocess
 import sys
 from pathlib import Path
@@ -52,15 +53,15 @@ base_completion_list: List[str] = [
         ("", 2, base_completion_list),
         (" ", 2, base_completion_list),
         ("dict", 2, ["dict.", "dict_prefix="]),
-        ("dict.", 3, ["dict.key1=", "dict.key2=", "dict.key3="]),
+        ("dict.", 2, ["dict.key1=", "dict.key2=", "dict.key3="]),
         ("dict.key", 2, ["dict.key1=", "dict.key2=", "dict.key3="]),
-        ("dict.key1=", 2, ["dict.key1=val1"]),
+        ("dict.key1=", 2, ["dict.key1=val1 "]),
         ("dict.key3=", 2, ["dict.key3="]),  # no value because dict.key3 is ???
         ("list", 2, ["list.", "list_prefix="]),
         ("list.", 2, ["list.0=", "list.1=", "list.2="]),
         (
             "hydra/",
-            3,
+            2,
             [
                 "hydra/help=",
                 "hydra/hydra_help=",
@@ -72,14 +73,14 @@ base_completion_list: List[str] = [
             ],
         ),
         ("test_hydra/lau", 2, ["test_hydra/launcher="]),
-        ("test_hydra/launcher=", 2, ["test_hydra/launcher=fairtask"]),
-        ("test_hydra/launcher=fa", 2, ["test_hydra/launcher=fairtask"]),
+        ("test_hydra/launcher=", 2, ["test_hydra/launcher=fairtask "]),
+        ("test_hydra/launcher=fa", 2, ["test_hydra/launcher=fairtask "]),
         # loading groups
         ("gro", 2, ["group="]),
-        ("group=di", 2, ["group=dict"]),
+        ("group=di", 2, ["group=dict "]),
         (
             "group=dict ",
-            3,
+            2,
             [
                 "dict.",
                 "dict_prefix=",
@@ -93,9 +94,9 @@ base_completion_list: List[str] = [
                 "toys.",
             ],
         ),
-        ("group=", 2, ["group=dict", "group=list"]),
-        ("group=dict group.dict=", 2, ["group.dict=true"]),
-        ("group=dict group=", 2, ["group=dict", "group=list"]),
+        ("group=", 2, ["group=dict ", "group=list "]),
+        ("group=dict group.dict=", 2, ["group.dict=true "]),
+        ("group=dict group=", 2, ["group=dict ", "group=list "]),
     ],
 )
 class TestCompletion:
@@ -144,6 +145,9 @@ class TestCompletion:
         # Fish starts completion with one tab compared to bash's two tabs.
         if shell == "fish":
             num_tabs -= 1
+            # Escape dot for regexp matching of expect. Add $ for exact string end match
+            # (would fail if fish adds an extra space).
+            expected = [re.escape(x) + "$" for x in expected]
 
         cmd.extend(
             [
```
```diff
diff --git a/tests/expect/test_fish_completion.exp b/tests/expect/test_fish_completion.exp
index d23da099..94212802 100644
--- a/tests/expect/test_fish_completion.exp
+++ b/tests/expect/test_fish_completion.exp
@@ -35,7 +36,7 @@ puts "\nMatching:\n"
 foreach arg [lrange $argv 3 end] {
     send "\t"
     expect {
-       "$arg" {puts "matched '$arg' to '$expect_out(0,string)'\nbuffer: '$expect_out(buffer)'";continue}
+       -re "$arg" {puts "matched '$arg' to '$expect_out(0,string)'\nbuffer: '$expect_out(buffer)'";continue}
         timeout {puts "Error matching $arg";exit 1}
     }
 }
```

omry:
Gotcha, don't feel guilty! just be back to finish the job. I think it would be great to have fish completion support :).

It's hard to see what the changes are in this form, feel free to update the PR.

I agree that it's best to test a single fish version (latest).
the completion installation can verify if the version fish version is supported and abort installation with a warning otherwise.

Do you know when the fish changes will be available for normal consumption?

.circleci/config.yml
```diff
@@ -92,6 +92,8 @@ jobs:
       - run:
           name: "Preparing environment"
           command: |
+            sudo apt-add-repository ppa:fish-shell/release-3
```
omry: Please add a comment here explaining why we need this.

hydra/_internal/core_plugins/fish_completion.py
```diff
@@ -22,8 +19,14 @@ def install(self) -> None:
     set -l parts (string split -n ' ' $COMP_LINE)
     if test "$parts[1]" = "python" -o "$parts[1]" = "python3"
         set cmd "$parts[1] $parts[2]"
+        set script "$parts[2]"
     else
         set cmd "$parts[1]"
+        set script "$parts[1]"
+    end
+
+    if not grep -q "@hydra.main" $script
```
omry, line 28: this should only be used for "python foo.py" mode.
if you are installing an app with entry point scripts you will end up with an executable (.exe on Windows).

tests/expect/test_fish_completion.exp
```diff
@@ -16,17 +16,11 @@ set len [expr [string length $in_line] - 1]
...
 
 send "[lindex $argv 0] -sc install=fish | source\r"
 expect >
 
-set command "[lindex $argv 0] $line"
-while {$num_tabs > 0} {
-    incr num_tabs -1
-    append command "\t"
-}
-
```
omry: this makes sense to me.
the whole thing with multiple tabs is very bash specific.

tests/expect/test_fish_completion.exp
```diff
@@ -35,7 +29,7 @@ puts "\nMatching:\n"
 foreach arg [lrange $argv 3 end] {
     send "\t"
     expect {
-	"$arg" {puts "matched '$arg' to '$expect_out(0,string)'\nbuffer: '$expect_out(buffer)'";continue}
+	-re "$arg" {puts "matched '$arg' to '$expect_out(0,string)'\nbuffer: '$expect_out(buffer)'";continue}
```
omry, line 32: I can
can't quiet remember what -re is doing.
please explain.

tests/test_completion.py
```diff
@@ -152,7 +153,12 @@ def test_shell_integration(
 
         # Fish starts completion with one tab compared to bash's two tabs.
```
omry, line 154: comment outdated

tests/test_completion.py
```diff
@@ -152,7 +153,12 @@ def test_shell_integration(
...
         if shell == "fish":
-            num_tabs -= 1
+            # Fish will add a space to an unambiguous completion.
+            expected = [x + " " if re.match(r".*=\w+$", x) else x for x in expected]
+
+            # Exactly match token end. See
+            # https://github.com/fish-shell/fish-shell/issues/6928
+            expected = [re.escape(x) + "$" for x in expected]
```
omry, line 161: can we do this unconditionally?

binshengliu: For bash as well? I don't think the current expect script for bash supports doing this way.

tests/expect/test_fish_completion.exp
binshengliu: Matching regexp. "$arg" would be regexp as we want to do exact line end match.

hydra/_internal/core_plugins/fish_completion.py
binshengliu: Nice catch.

.circleci/config.yml
binshengliu: I think I parsed their git log incorrectly. The space fix is in master but not the latest release yet. We can test against a nightly build now and use the official release once a new version is out.

tests/expect/test_fish_completion.exp
omry: are you escaping the inputs?
expected strings have ., you should re.escape() them before passing them otherwise the . would interact with the regex matching.

tests/test_completion.py
omry: Yes, I am trying to minimize the things that are making fish special. your suggesting to test against eol makes sense for bash as well, with a bit of luck it will just pass if you do it there and add the -re to the expect match.

.circleci/config.yml
omry: ah, it did seem a bit too fantastic to me :)
can we automatically install nightly?

omry: if not, we can skip add a skip for now and re-enable it once they have a proper release.

binshengliu: Indeed managing their releases is a bit overkill. I got the code to run tests only with fish > 3.1.2 or git build after 3.1.2. The latter is for local tests.

tests/expect/test_fish_completion.exp
binshengliu: yes they are escaped.

binshengliu:
Can you point me how you manage your doc and news? There seem to be multiple copies of the doc. What should I do with news?

tests/test_completion.py
```diff
@@ -8,6 +8,7 @@
 from typing import List
 
 import pytest
+from packaging import version
```
omry, line 11: that's a tricky one. Packaging is available because of the pytest dependency.
```
$ pipdeptree  -r -p packaging
packaging==20.3
  - pytest==5.4.1 [requires: packaging]
```

The pytest dependency is a test only one, so this will not work at runtime.
It's fine, but I was surprised to see you use it here without having to add the dependency.

hydra/_internal/utils.py
```diff
@@ -300,10 +302,17 @@ def get_args_parser() -> argparse.ArgumentParser:
...
+    {}
+
+    Fish shell uninstall:
+    {}
 """.format(
-            install_cmd, uninstall_cmd
+            install_cmd, uninstall_cmd, fish_install_cmd, fish_uninstall_cmd
         ),
     )
+
```
omry, line 315: Please clear this from the diff and create a new issue to improve the completion plugin interface to support per-plugin help.

omry:
As for CI errors, please rebase:
I already fixed those.

tests/test_completion.py
binshengliu: Hmm, I thought it was a built-in module. I'll add that dependency in dev requirements.

hydra/_internal/utils.py
binshengliu: I'm not sure I know what you mean by clearing it without modifying the interface. I can extend that help interface as part of this pr.

omry: That works too.
How 
```python
class CompletionPlugin(Plugin):

    @abstractmethod
    def help(self, command: str):
        """
        :param command: "install" or "uninstall"
        :return: command the user can run to install or uninstall this shell completion on the appropriate shell
        """
        ...

```

tests/test_completion.py
omry: sounds good.

omry:
> 
> 
> Can you point me how you manage your doc and news? There seem to be multiple copies of the doc. What should I do with news?

The news snippets are documented in the contributing guide.
The docs - update the main copy (website/docs).
You can see what it looks like by running the web server locally (follow the README in website) and switching to the `next` version from the versions link at the top bar.

binshengliu:
As for the uninstall cmd, I think the manual is very clear about what it can do now. AFAIK typical fish completion plugins install their registering command as a fish script to be run at fish startup. If the user doesn't want the completion they just remove the script. I haven't seen erase command being used in any applications so far. But sure we can ask them for an answer regarding the dynamic completion.

hydra/_internal/utils.py
binshengliu: Cool.

omry:
> As for the uninstall cmd, I think the manual is very clear about what it can do now. AFAIK typical fish completion plugins install their registering command as a fish script to be run at fish startup. If the user doesn't want the completion they just remove the script. I haven't seen erase command being used in any applications so far. But sure we can ask them for an answer regarding the dynamic completion.

bash completion allows deletion of individual completions, it seems sensible to have a reverse command for each of the commands registering a completion with zsh as well.
If it's not supported we can just make it unsupported and tell the user he need to restart the shell, but I would like to confirm with the fish devs.

binshengliu:
A few notes here.

The plugin instantiation seems depending on some logic at later stages (ConfigLoader) so I changed `help` and `provides` into static to be available to generate help info.

And I had a closer look at bash's uninstall methods.

```
$ python test.py -sc uninstall=bash
unset hydra_bash_completion
complete -r python
```

Fish supports the same function to `complete -r python` by `complete -e -c python`. We can do the same thing but this is what I've always been trying to avoid. It's too aggressive that it also removes all the original python completions (-v, -h, ...) already available in the system.

Unsetting the completion function is a really neat trick and does the job, so I suggest we also reset the completion function to empty to achieve uninstalling effect for fish.

```
$ python test.py -sc uninstall=fish
complete -e -c test.py
function hydra_fish_completion
end
```

tests/test_hydra.py
```diff
@@ -497,17 +497,13 @@ def test_sweep_complex_defaults(
 --run,-r : Run a job
 --multirun,-m : Run multiple jobs with the configured launcher
 --shell_completion,-sc : Install or Uninstall shell completion:
-    Install:
-    eval "$(python examples/tutorial/2_config_file/my_app.py -sc install=SHELL_NAME)"
+    Bash:
+    eval "$(python examples/tutorial/2_config_file/my_app.py -sc install=bash)"
+    eval "$(python examples/tutorial/2_config_file/my_app.py -sc uninstall=bash)"
```
omry, line 502:
```suggestion
    Bash - Install:
    eval "$(python examples/tutorial/2_config_file/my_app.py -sc install=bash)"
    Bash - Uninstall:
    eval "$(python examples/tutorial/2_config_file/my_app.py -sc uninstall=bash)"
```

omry: We can also potentially introduce a shell specific help, but this could be a bit dicey:
```
$ python foo.py --help -sc
Supported shells: bash, fish
For shell specific help, use : --help -sc NAME 

$ python foo.py --help -sc bash
Bash - Install:
  eval "$(python examples/tutorial/2_config_file/my_app.py -sc install=bash)"
Bash - Uninstall:
  eval "$(python examples/tutorial/2_config_file/my_app.py -sc uninstall=bash)"
```

omry:
Un-setting the completion function should work.
Do you know if it will do a reasonable job with Bash as well? if so we can do the same thing there.

tests/test_hydra.py
binshengliu: That looks quite extensible but I'm not sure if this improves user experiences atm. As a user, I'd prefer using one command (python foo.py --hydra-help) to get everything I need in one page.

binshengliu:
I guess not for bash but I'm not completely sure. It seems bash only supports one rule for a command, so the existing one is always overwritten by a new registration. Fish supports multiple rules so the new rule doesn't affect others.

```
$ complete | grep python
complete -o bashdefault -o default -F _fzf_path_completion python
$ eval "$(python test.py -sc install=bash)"
$ complete | grep python
complete -o default -o nospace -F hydra_bash_completion python
```

It seems remembering the old rule using an env variable would work. [complete manual](https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html)
`If the -p option is supplied, or if no options are supplied, existing completion specifications are printed in a way that allows them to be reused as input.`

I pushed a suggestion for this idea but you may want to do more tweaking.

tests/test_hydra.py
omry: Okay. at this point I think it's okay but eventually too much of the help screen would be dedicated to this.
I can live what your current proposal, but update it to both the shell and the commands in the header as I suggested.

omry:
I see.
it's a pity that fish does not support uninstalling an individual completion.
for now your suggestion for using an empty function is acceptable.

hydra/_internal/core_plugins/bash_completion.py
```diff
@@ -16,7 +16,11 @@
 class BashCompletion(CompletionPlugin):
     # TODO: detect python with path like /foo/bar/python
     def install(self) -> None:
-        script = """hydra_bash_completion()
+        # Record the old rule for uninstalling
+        script = (
+            f"export _HYDRA_OLD_COMP=$(complete -p {self._get_exec()} 2> /dev/null)\n"
+        )
+        script += """hydra_bash_completion()
```
omry, line 23: since you are improving this, can you add a test as well?
you can verify state by grepping the output of complete.

omry:
rebase to fix windows CI issue.

hydra/_internal/utils.py
```diff
@@ -286,6 +286,23 @@ def _get_exec_command() -> str:
...
 
 
+def _get_completion_help() -> str:
+    from hydra.core.plugins import Plugins
+    from hydra.plugins.completion_plugin import CompletionPlugin
+
+    completion_plugins = Plugins.instance().discover(CompletionPlugin)
+    completion_info: List[str] = []
+    for cls in completion_plugins:
+        plugin_cls = cast(Type[CompletionPlugin], cls)
```
omry, line 296: A perceptually cleaner way to do this is to use assert:
```suggestion
    for plugin_cls in completion_plugins:
        assert isinstance(plugin_cls, CompletionPlugin)
```

hydra/_internal/core_plugins/bash_completion.py
binshengliu: Can you suggest a change? I don't get it what to test?
The output is like
```
$ complete -p python
complete -o bashdefault -o default -F _fzf_path_completion python
```
or if the completion doesn't exist
```
$ complete -p notexist
bash: complete: notexist: no completion specification
```

omry: This is what I had in mind:

You preserve the old completion and restore it on uninstall.
test:
0. clear all completion
1. configure some completion for python (not Hydra).
2. install Hydra bash completion
3. uninstall Hydra bash completion
4. verify your the old completion is restored.

binshengliu:
OK, the install/uninstall tests for both bash and fish are added. If possible I'm hoping to finalize this asap as I don't have too much time on it now.

omry:
First of all, I want to thank you for your patience and time investment.
This is a great contribution. I have some Hydra stickers I would love to send you. reach out to me in a DM on the chat if you are interested.

We are really close.
1. I added a TODO suggestion as a reminder for me to support the end of line $ for bash too.
2. update the docs in website/docs, particular, the [tab completion page](https://hydra.cc/docs/next/tutorials/basic/tab_completion) in the basic tutorial is saying only bash is supported.
3. add a new file fragment. basically `news/ISSUE_NUM.feature` with a line about this new feature.
see [this](https://hydra.cc/docs/next/development/contributing#news-entries) for more info about news entries.

binshengliu:
Not a problem. It's fun to learn and fiddle with something new. I also appreciate your patience and guidance. The doc change should be straightforward. I'll finish asap.

omry:
Looking good.
please fix the lint error and I will land it.

binshengliu:
Can you have a look at the lint error? I didn't touch this file.
```
nox > flake8 --config .flake8
.\hydra\test_utils\test_utils.py:314:23: F541 f-string is missing placeholders
```

omry:
Please rebase and try again.
not sure why that passed CI before.

omry:
Awesome, thanks!

Issue #554, opened 2020-04-26
[Ax] Fix OmegaConf warnings related to the future compact key support


shagunsodhani:
Related to https://github.com/omry/omegaconf/issues/152

shagunsodhani:
Closed in https://github.com/shagunsodhani/hydra/commit/6927d43f9c1b0c1d0241468cb3eb4e43cb4af9c6

PR #555, opened 2020-04-26
Fixing warnings related to OmegaConf 2.0.0 compact key

shagunsodhani:
## Motivation

[Ax] Fix warnings related to the compact key support 

Yes

## Test Plan

Will add some tests

## Related Issues and PRs

https://github.com/facebookresearch/hydra/issues/554

### DIFF
--- a/plugins/hydra_ax_sweeper/example/conf/config.yaml
+++ b/plugins/hydra_ax_sweeper/example/conf/config.yaml
@@ -25,9 +25,10 @@ hydra:
           max_epochs_without_improvement: 10
 
         params:
-          banana.x:
-            type: range
-            bounds: [-5, 5]
-          banana.y:
-            type: range
-            bounds: [-5, 10.1]
+          banana:
+            x:
+              type: range
+              bounds: [-5, 5]
+            y:
+              type: range
+              bounds: [-5, 10.1]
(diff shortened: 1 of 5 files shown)


omry:
Next time it's best to discuss user facing changes before you implement them to be sure we reach consensus.

Here are the options I see:
1. What you did:
```yaml
params:
  banana:
    x:
      type: range
      bounds: [-5, 5]
    y:
      type: range
      bounds: [-5, 5]
```

2. Using a list, eliminating the need for a unique key:
```yaml
params:
- key: banana.x
  type: range
  bounds: [-5, 5]
- key: banana.y
  type: range
  bounds: [-5, 5]
```

3. allowing the user to define simple unique aliases (no .)
```yaml
params:
  x:
    key: banana.x
    type: range
    bounds: [-5, 5]
  y:
    key: banana.y
    type: range
    bounds: [-5, 5]
```

Before we discuss the props and cons, let me know if you can think of any other option.

omry:
Okay, after discussing, we decided on option 4 (new).
We will use a different separator in the config:
```yaml
# config:
banana:
 x: 10

# parameters description:
params:
  banana_x:
    type: range
    bounds: [-5, 5]
```

This brings up an issue of how to handle keys that has `_`, the solution we decided on is to allow escaping of the _ in the key:
```yaml
# config:
banana:
 foo_bar: 10

# parameters description:
params:
  banana_foo\_bar:
    type: range
    bounds: [-5, 5]
```

cc @jrapin: fyi.
Don't implement this in the NeverGrad plugin.
@shagunsodhani will create the initial implementation in Ax and then will then create the shared core opt package where we can start sharing the implementation between the two plugins (also switching to the agreed upon command line and changing the config to follow the same kind of changes).

shagunsodhani:
Closed in favor of https://github.com/facebookresearch/hydra/pull/567

Issue #577, opened 2020-05-07
Fix Ax sweeper to pass common sweeper tests


omry:
#576 introduces common sweeper tests that should pass for all Sweeper.
Currently those tests are failing for the Ax sweeper, please re-enable the tests (commented out in #576) and fix the Ax sweeper such that they pass.

omry:
Closing for now. we can revisit what improving testing for optimizing sweepers once they have a common library.

PR #587, opened 2020-05-14
wip - show possible way of integrating the sweeper suite

shagunsodhani:
## Motivation

WIP. Related to https://github.com/facebookresearch/hydra/issues/577

Yes

## Test Plan

WIP

## Related Issues and PRs

https://github.com/facebookresearch/hydra/issues/577

### DIFF
--- a/hydra/test_utils/launcher_common_tests.py
+++ b/hydra/test_utils/launcher_common_tests.py
@@ -4,28 +4,22 @@ Common test functions testing launchers
 """
 import copy
 from pathlib import Path
-from typing import Any, Callable, List, Optional, Set
+from typing import Any, List, Optional, Set
 
 import pytest
-from omegaconf import DictConfig, OmegaConf
-
 from hydra import TaskFunction
 from hydra.test_utils.test_utils import (
     TSweepRunner,
     integration_test,
     verify_dir_outputs,
 )
-
-ValidateSweepType = Callable[[TSweepRunner], None]
+from omegaconf import DictConfig, OmegaConf
 
 
 @pytest.mark.usefixtures("restore_singletons")
 class LauncherTestSuite:
     def task_function(self, cfg: DictConfig) -> Any:
         return 100
-
-    def validate_sweep_1_job(self, sweep: TSweepRunner):
-        return validate_sweep_1_job(sweep)
 
     def test_sweep_1_job(
         self, sweep_runner: TSweepRunner, launcher_name: str, overrides: List[str],
@@ -36,7 +30,6 @@ class LauncherTestSuite:
             + overrides,
             strict=False,
             task_function=self.task_function,
-            function_to_validate_sweep=self.validate_sweep_1_job,
         )
 
     def test_sweep_2_jobs(
@@ -227,7 +220,6 @@ def sweep_1_job(
     overrides: List[str],
     strict: bool,
     task_function: Optional[TaskFunction],
-    function_to_validate_sweep: ValidateSweepType,
 ) -> None:
     """
     Runs a sweep with one job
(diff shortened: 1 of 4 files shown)


shagunsodhani:
Things to consider:

(i) The Sweeper Suite expects some non-None return value. Ax plugin doesnt return any value. Since the output formats are different, I will have to override how the output of the sweep is verified.
(ii) This PR is a WIP and does the override for only 1 test, to show how it would look like.

shagunsodhani:
> > (i) The Sweeper Suite expects some non-None return value. Ax plugin doesnt return any value. Since the output formats are different, I will have to override how the output of the sweep is verified.
> 
> Can the Ax plugin be modified to return the standard expected output?

Should be doable. Infact I am not sure why we didnt do it originally. Is the format of "standard expected output` specified somewhere? Adding a related issue: https://github.com/facebookresearch/hydra/issues/468

omry:
This is purely for testing as far as I remember.
no return value, no testing.

It seems likely that exact verification can be plugin specific, but at the very least they should return something helpful.
This task is about making them line up with the basic sweeper test verification. if that does not make sense we can create something more specialized that can hopefully cover all optimizing sweepers.

shagunsodhani:
Some points to highlight between the HPO sweepers and basic sweeper (as I was integrating the test suite):

* Basic sweeper returns the output of all the combinations of the input parameters. By output, I am referring to the [JobReturn](https://github.com/facebookresearch/hydra/blob/995782be780dccadccbf2680696698404ada1693/hydra/core/utils.py#L164) object. For Ax, the total number of combinations can be very large and in practice we use early stopping. Even then, the number of combinations that Ax sweeps over could be large (say 100). I am proposing that Ax sweeper (and maybe all HPO sweepers) return only the best JobReturn object. The sweeper plugin would have to maintain the required state internally.

* `sweep_1_job`, `sweep_2_jobs` probably do not make sense for HPO as we do not have a syntax for: 

> Optimize quadratic function for x=-1:1, 2:3 y=-1:1,2:3 ie consider the following four combinations: 
* x=-1:1 y=-1:1
* x=-1:1 y=2:3
* x=2:3 y=-1:1
* x=2:3 y=2:3

ie, we can not specify that the plugin should use two different ranges (or choices...) of any parameter (say x).

* `not_sweeping_hydra_overrides` should be doable.

* `sweep_two_config_groups` eg `optimizers=adam,nesterov` In the current setup, these arguments will be passed to the HPO plugins as well. In case of Ax, it will be interpreted as a choice parameter

omry:
The intention here is to ensure that some common things are tested for all plugins.
If the optimizing sweepers are too different we can introduce a common test for them in the hydra-opt-core package we discussed.

shagunsodhani:
I added implemented more tests. Regarding `test_sweep_2_jobs` and `test_sweep_2_optimizers`, I dont know if they make sense for HPO (described in ). Regarding `test_sweep_over_unspecified_mandatory_default` and `test_sweep_and_override`, I am not clear if they make sense for HPO (need to think more).

We need to agree on what tests should be implemented by the HPOs.

What is missing:

1.  `JobReturn` object should return the best params. Currently, it returns the last params.
2.  Once point 1 is finished, we do not need to write the results on filesystem and can use the `JobReturn` object to validate them.

hydra/test_utils/launcher_common_tests.py
```diff
@@ -7,14 +7,13 @@
...
 import pytest
-from omegaconf import DictConfig, OmegaConf
-
 from hydra import TaskFunction
 from hydra.test_utils.test_utils import (
     TSweepRunner,
     integration_test,
     verify_dir_outputs,
 )
+from omegaconf import DictConfig, OmegaConf
```
omry, line 16: odd, was this not sorted?

shagunsodhani: iirc it looks like this after sorting :(

omry: check the contributing document to see how to run isort for plugins through nox.

omry:
this is not going anywhere right? close?

shagunsodhani:
Probably better to make a fresh attempt

PR #634, opened 2020-06-01
[Submitit Plugin] Prevent default retry on timeout

jrapin:
## Motivation

The current default configuration was retrying once when timed-out. This is an unexpected behavior in my opinion, jobs should be retried on timeouts only if the user explicitly ask for it (cc @alexholdenmiller ;) )

Yes

## Test Plan

N/A

### DIFF
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/conf/hydra/launcher/submitit.yaml
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/conf/hydra/launcher/submitit.yaml
@@ -10,7 +10,7 @@ params:
     # slurm queue parameters
     slurm:
       nodes: 1
-      num_gpus: 1
+      gpus_per_node: 1
       ntasks_per_node: 1
       mem: ${hydra.launcher.mem_limit}GB
       cpus_per_task: 10
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/conf/hydra/launcher/submitit.yaml
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/conf/hydra/launcher/submitit.yaml
@@ -29,7 +29,9 @@ params:
       timeout_min: 60
     # auto queue parameters
     auto:
-      max_num_timeout: 1
+      slurm_max_num_timeout: 0
+      gpus_per_node: 1
+      timeout_min: 60
 
 # variables used by queues above
 mem_limit: 24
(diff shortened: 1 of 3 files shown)


jrapin:
Additional observation:
This using the slurm executor directly. Was there any reason? I would advise using auto executor whenever possible instead.

omry:
Can you also fix the doc page (it has the previous default):

https://hydra.cc/docs/next/plugins/submitit_launcher

re: autoexecutor:
I think I am not convinced that it's the right approach for the Hydra plugin. In my opinion using config groups to configure the desired executor is more flexible and also making less assumptions.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/conf/hydra/launcher/submitit.yaml
```diff
@@ -10,7 +10,7 @@ params:
     # slurm queue parameters
     slurm:
       nodes: 1
-      num_gpus: 1
```
jrapin, line 13: this is deprecated now

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/conf/hydra/launcher/submitit.yaml
```diff
@@ -21,15 +21,17 @@ params:
...
     # local queue parameters
     local:
       gpus_per_node: 1
       tasks_per_node: 1
       timeout_min: 60
     # auto queue parameters
     auto:
-      max_num_timeout: 1
+      slurm_max_num_timeout: 0
+      gpus_per_node: 1
```
jrapin, line 33: I've added these so that it's equivalent to the local queue

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -99,16 +99,12 @@ def launch(
         queue_parameters = self.queue_parameters.copy()
         OmegaConf.set_struct(queue_parameters, True)
         if self.queue == "auto":
-            max_num_timeout = self.queue_parameters.auto.max_num_timeout
-            with open_dict(queue_parameters):
```
jrapin, line 103: This raised an error when trying on slurm.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/conf/hydra/launcher/submitit.yaml
```diff
@@ -21,15 +21,17 @@ params:
...
+      max_num_timeout: 0
     # local queue parameters
     local:
       gpus_per_node: 1
       tasks_per_node: 1
       timeout_min: 60
     # auto queue parameters
     auto:
-      max_num_timeout: 1
+      slurm_max_num_timeout: 0
```
jrapin, line 32: max_num_timeout was deprecated for auto executor, since it is currently slurm specific

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
omry: Thanks.
We need some proper testing for this plugin.

Issue #623, opened 2020-05-29
Move submitit launcher config to structured config


jieru-hu:
a follow up on https://github.com/facebookresearch/hydra/pull/619/files

PR #637, opened 2020-06-01
Move submitit launcher to structured config.

jieru-hu:
## Motivation

#623 

Yes/No

## Test Plan

## Related Issues and PRs


plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -0,0 +1,75 @@
...
+
+@dataclass
+class SlurmQueueConf:
+    nodes: int = 1
+    gpus_per_node: int = 1
+    ntasks_per_node: int = 1
+    mem: str = "${hydra.launcher.mem_limit}GB"
+    cpus_per_task: int = 10
+    time: int = 60
+    partition: str = "dev"
```
omry, line 24: I am actually not sure that this makes sense for the opensource plugin version.
We should probably make FAIR specific.

@jrapin, what would be a good default value for the partition in the open source plugin (The FAIR cluster can get different default values through the fair-cluster-defaults plugin).

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -0,0 +1,75 @@
...
+    signal_delay_s: int = 120
+    job_name: str = "${hydra.job.name}"
```
omry, line 26: try with `${hydra:job.name}`.
This is now the recommended form for Hydra config interpolation. It will also work inside the user's config.

website/docs/plugins/submitit_launcher.md
```diff
@@ -29,45 +29,59 @@ defaults:
 Note that this plugin expects a valid environment in the target host. usually this means a shared file system between
 the launching host and the target host.
 
-Submitit supports 3 kind of queues: auto, local and slurm. Its config looks like this
+Submitit supports 3 types of queues: auto, local and slurm. Its config looks like this
```
omry, line 32: Yeah, I am starting to think that maybe yaml is a better form to document the config despite it missing the types.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jieru-hu: set it to ```dev``` to be consistent with submitit's example https://github.com/facebookincubator/submitit/blob/master/docs/examples.md

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -0,0 +1,74 @@
...
+    folder: str = "${hydra.sweep.dir}/.${hydra.launcher.params.queue}"
+
+    queue_parameters: QueueParams = QueueParams()
+
+
+@dataclass
+class SubmititLauncherConf(ObjectConf):
+    cls: str = "hydra_plugins.hydra_submitit_launcher.submitit_launcher.SubmititLauncher"
+    params: SubmititConf = SubmititConf()
+    mem_limit: int = 2
```
jieru-hu, line 66: set this to 2GB to be consistent with submitit's example https://github.com/facebookincubator/submitit

website/docs/plugins/submitit_launcher.md
jieru-hu: agreed

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jieru-hu: per offline discussion, keeping it as ```hydra.``` for now.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -0,0 +1,74 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
omry, line 1: Can you document the fields here (comments)?

jieru-hu: Added reference to submitit. I thought about adding it line by line in hydra but it seems repetitive.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -0,0 +1,78 @@
...
+    timeout_min: int = 60
+
+
+@dataclass
+class AutoQueueConf:
+    # check the following for more info on slurm_max_num_timeout
+    # https://github.com/facebookincubator/submitit/blob/master/docs/checkpointing.md
+    slurm_max_num_timeout: int = 0
+    gpus_per_node: int = 1
+    timeout_min: int = 60
```
jrapin, line 47: In practice there is much more to it that only those parameters, should they *all* be referenced?

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jrapin: > set it to dev to be consistent with submitit's example

At some point we may remove the default since even dev probably does not exist everywhere... also using dev is not optimal as a default for FAIR :s no perfect solution here I think, dev is good enough for now :s

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
omry: It would be awesome if you can send a PR completing it. This plugin is based on the somewhat outdated internal submitit plugin.

PR #638, opened 2020-06-01
Update hydra-core requirements spec

willfrey:
## Motivation

The requirements for hydra-core in plugins/**/setup.py were malformed. This fixes them so that the plugins are installable from PyPI with the 1.0.0rc1 version of hydra-core.

Yes.

## Test Plan

I manually tested building hydra-core with versions 1.0.0rc1, 1.0.0, 1.0.1rc1, 1.0.1 and installed hydra-ax-sweeper from pre-built wheels using only a local PyPI server. For each version listed, I had only that version available on my local PyPI server. And in all cases, that version was installed without an issue.

I also tested building hydra-core with versions 1.1.0rc1 and 1.1.0. In both cases, installation failed when installing hydra-px-sweeper, as one would expect.

### DIFF
--- a/plugins/hydra_ax_sweeper/setup.py
+++ b/plugins/hydra_ax_sweeper/setup.py
@@ -22,6 +22,6 @@ with open("README.md", "r") as fh:
             "Operating System :: POSIX :: Linux",
             "Development Status :: 4 - Beta",
         ],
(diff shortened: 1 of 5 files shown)


omry:
awesome, thanks.

PR #640, opened 2020-06-01
re-released all plugins with proper dependency on hydra-core

omry:


plugins/hydra_submitit_launcher/setup.py
```diff
@@ -5,7 +5,7 @@
     LONG_DESC = fh.read()
     setup(
         name="hydra-submitit-launcher",
-        version="1.0.0rc1",
+        version="1.0.0rc333
```
jieru-hu, line 8: why 333?

plugins/hydra_nevergrad_sweeper/setup.py
```diff
@@ -6,7 +6,7 @@
     LONG_DESC = fh.read()
     setup(
         name="hydra-nevergrad-sweeper",
-        version="1.0.0rc2",
+        version="1.0.0rc4",
```
jieru-hu, line 9: why the rc version increments by 2?

plugins/hydra_submitit_launcher/setup.py
omry: well shit, good catch.

plugins/hydra_nevergrad_sweeper/setup.py
omry: I am keeping the repo version on the next version, not on the released version.

PR #651, opened 2020-06-05
[submitit] Add AutoExecutor parameters

jrapin:
## Motivation

Add parameters for AutoExecutor

Yes

## Test Plan

Manually tested on SLURM cluster with auto executor

## Related Issues and PRs

### DIFF
--- a/plugins/hydra_submitit_launcher/example/config.yaml
+++ b/plugins/hydra_submitit_launcher/example/config.yaml
@@ -1,4 +1,11 @@
 defaults:
   - hydra/launcher: submitit
+hydra:
+  launcher:
+    params:
+      queue: auto
+      queue_parameters:
+        auto:
+          cluster: local
 
 task: 1
--- a/plugins/hydra_submitit_launcher/example/config.yaml
+++ b/plugins/hydra_submitit_launcher/example/config.yaml
@@ -1,11 +1,4 @@
 defaults:
   - hydra/launcher: submitit
-hydra:
-  launcher:
-    params:
-      queue: auto
-      queue_parameters:
-        auto:
-          cluster: local
 
 task: 1
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
@@ -95,7 +95,7 @@ class AutoQueueConf:
     slurm_signal_delay_s: int = 30
     # slurm partition to use on the cluster
     # (depending on the cluster configuration, default "dev" may not exist)
-    partition: str = "dev"
+    slurm_partition: str = "dev"
 
 
 @dataclass
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
@@ -28,11 +28,10 @@ class SlurmQueueConf:
     ntasks_per_node: int = 1
     # number of nodes to use for the job
     nodes: int = 1
-    # memory to reserve for the job on each node
+    # memory to reserve for the job on each node, in GB
     mem: str = "${hydra.launcher.mem_limit}GB"
     # slurm partition to use on the cluster
-    # (depending on the cluster configuration, default "dev" may not exist)
-    partition: str = "dev"
+    partition: Optional[str] = None
     # USR1 signal delay before timeout
     signal_delay_s: int = 120
     # name of the job
@@ -94,8 +93,7 @@ class AutoQueueConf:
     # USR1 signal delay before timeout for the slurm queue
     slurm_signal_delay_s: int = 30
     # slurm partition to use on the cluster
-    # (depending on the cluster configuration, default "dev" may not exist)
-    slurm_partition: str = "dev"
+    slurm_partition: Optional[str] = None
 
 
 @dataclass
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
@@ -116,6 +116,7 @@ class SubmititConf:
 class SubmititLauncherConf(ObjectConf):
     cls: str = "hydra_plugins.hydra_submitit_launcher.submitit_launcher.SubmititLauncher"
     params: SubmititConf = SubmititConf()
+    # memory to reserve for the job on each node, in GB
     mem_limit: int = 2
 
 
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
@@ -65,7 +65,7 @@ class AutoQueueConf:
 
     # cluster to use (currently either "slurm" or "local" are supported,
     # None defaults to an available cluster)
-    cluster: Optional[str] = None
+    cluster: str = "slurm"
 
     # maximum time for the job in minutes
     timeout_min: int = 60
@@ -105,7 +105,7 @@ class QueueParams:
 
 @dataclass
 class SubmititConf:
-    queue: QueueType = QueueType.local
+    queue: QueueType = QueueType.slurm
 
     folder: str = "${hydra.sweep.dir}/.${hydra.launcher.params.queue}"
 
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
@@ -1,6 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 from dataclasses import dataclass
 from enum import Enum
+from typing import Optional
 
 from hydra.core.config_store import ConfigStore
 from hydra.types import ObjectConf
@@ -16,35 +17,85 @@ class QueueType(Enum):
 class SlurmQueueConf:
     # Params are used to configure sbatch, for more info check:
     # https://github.com/facebookincubator/submitit/blob/master/submitit/slurm/slurm.py
-    nodes: int = 1
-    gpus_per_node: int = 1
-    ntasks_per_node: int = 1
-    mem: str = "${hydra.launcher.mem_limit}GB"
-    cpus_per_task: int = 10
+
+    # maximum time for the job in minutes
     time: int = 60
+    # number of cpus to use for each task
+    cpus_per_task: int = 10
+    # number of gpus to use on each node
+    gpus_per_node: int = 1
+    # number of tasks to spawn on each node
+    ntasks_per_node: int = 1
+    # number of nodes to use for the job
+    nodes: int = 1
+    # memory to reserve for the job on each node
+    mem: str = "${hydra.launcher.mem_limit}GB"
+    # slurm partition to use on the cluster
+    # (depending on the cluster configuration, default "dev" may not exist)
     partition: str = "dev"
+    # USR1 signal delay before timeout
     signal_delay_s: int = 120
+    # name of the job
     job_name: str = "${hydra.job.name}"
     # Maximum number of retries on job timeout.
     # Change this only after you confirmed your code can handle re-submission
     # by properly resuming from the latest stored checkpoint.
+    # check the following for more info on slurm_max_num_timeout
+    # https://github.com/facebookincubator/submitit/blob/master/docs/checkpointing.md
     max_num_timeout: int = 0
 
 
 @dataclass
 class LocalQueueConf:
-    gpus_per_node: int = 1
-    tasks_per_node: int = 1
+    # local executor mocks the behavior of slurm locally
+
+    # maximum time for the job in minutes
     timeout_min: int = 60
+    # number of gpus to use on each node
+    gpus_per_node: int = 1
+    # number of tasks to spawn on each node (only one node available in local executor)
+    tasks_per_node: int = 1
 
 
 @dataclass
 class AutoQueueConf:
+    # auto executor automatically identifies and uses available cluster
+    # Currently this is only slurm, but local executor can be manually forced
+    # instead.
+    # Most parameters are shared between clusters, some can be cluster specific
+
+    # cluster to use (currently either "slurm" or "local" are supported,
+    # None defaults to an available cluster)
+    cluster: Optional[str] = None
+
+    # maximum time for the job in minutes
+    timeout_min: int = 60
+    # number of cpus to use for each task
+    cpus_per_task: int = 1
+    # number of gpus to use on each node
+    gpus_per_node: int = 0
+    # number of tasks to spawn on each node
+    tasks_per_node: int = 1
+    # memory to reserve for the job on each node (in GB)
+    mem_gb: int = 4
+    # number of nodes to use for the job
+    nodes: int = 1
+    # name of the job
+    name: str = "${hydra.job.name}"
+
+    # following parameters are SLURM specific
+
+    # Maximum number of retries on job timeout.
+    # Change this only after you confirmed your code can handle re-submission
+    # by properly resuming from the latest stored checkpoint.
     # check the following for more info on slurm_max_num_timeout
     # https://github.com/facebookincubator/submitit/blob/master/docs/checkpointing.md
     slurm_max_num_timeout: int = 0
-    gpus_per_node: int = 1
-    timeout_min: int = 60
+    # USR1 signal delay before timeout for the slurm queue
+    slurm_signal_delay_s: int = 30
+    # slurm partition to use on the cluster
+    # (depending on the cluster configuration, default "dev" may not exist)
+    partition: str = "dev"
 
 
 @dataclass
--- a/plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py
+++ b/plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py
@@ -39,7 +39,6 @@ class TestSubmititLauncher(LauncherTestSuite):
                 ],
                 "hydra": {
                     "launcher": {
-                        "cls": "hydra_plugins.hydra_submitit_launcher.submitit_launcher.SubmititLauncher",
                         "params": {
                             "queue": "local",
                             "folder": "${hydra.sweep.dir}/.${hydra.launcher.params.queue}",
@@ -66,7 +65,6 @@ class TestSubmititLauncher(LauncherTestSuite):
                 ],
                 "hydra": {
                     "launcher": {
-                        "cls": "hydra_plugins.hydra_submitit_launcher.submitit_launcher.SubmititLauncher",
                         "params": {
                             "queue": "auto",
                             "folder": "${hydra.sweep.dir}/.${hydra.launcher.params.queue}",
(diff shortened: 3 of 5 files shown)


plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -16,35 +17,85 @@ class QueueType(Enum):
...
+    # number of tasks to spawn on each node
     ntasks_per_node: int = 1
+    # number of nodes to use for the job
+    nodes: int = 1
+    # memory to reserve for the job on each node
     mem: str = "${hydra.launcher.mem_limit}GB"
-    cpus_per_task: int = 10
-    time: int = 60
+    # slurm partition to use on the cluster
+    # (depending on the cluster configuration, default "dev" may not exist)
```
omry, line 34: Is there a choice that always works? what happens if you don't specify the partition?

We can get a different default on the FAIR cluster (or other environments quiet easily through a ConfigSearchPathPlugin).

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -16,35 +17,85 @@ class QueueType(Enum):
...
+    # number of cpus to use for each task
+    cpus_per_task: int = 10
+    # number of gpus to use on each node
     gpus_per_node: int = 1
+    # number of tasks to spawn on each node
     ntasks_per_node: int = 1
+    # number of nodes to use for the job
+    nodes: int = 1
+    # memory to reserve for the job on each node
     mem: str = "${hydra.launcher.mem_limit}GB"
```
omry, line 32: Can you put a comment on mem_limit definition in the launcher that it's in GB?

plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py
```diff
@@ -55,7 +55,35 @@ class TestSubmititLauncher(LauncherTestSuite):
                 },
             },
             ["-m"],
-        )
+        ),
+        # auto queue
+        (
+            {
+                "defaults": [
+                    {"hydra/launcher": "submitit"},
```
omry, line 63: Seems to be worth adding a submit_local config option that would do the right thing.

plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py
```diff
@@ -55,7 +55,35 @@ class TestSubmititLauncher(LauncherTestSuite):
...
+                    {"hydra/hydra_logging": "hydra_debug"},
+                    {"hydra/job_logging": "disabled"},
+                ],
+                "hydra": {
+                    "launcher": {
+                        "cls": "hydra_plugins.hydra_submitit_launcher.submitit_launcher.SubmititLauncher",
```
omry, line 69: cls is probably not needed here (and not in where you copied it from).

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jrapin: 1. I have no idea where this mem_limit is supposed to come from, grepping mem_limit yields nothing in hydra.
2. Is there a way to get this value as an integer and not a string? That would be useful to set it in auto executor as well, since in that case it is not a string.

plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py
jrapin: Not sure what you mean here

plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py
jrapin: done

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jrapin: > Is there a choice that always works? what happens if you don't specify the partition?

Good point, I though that there was a default value in submitit but there is none, so it falls back to the default partition of the cluster, better initialize it as None then

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
omry: 1. https://github.com/facebookresearch/hydra/blob/979024ea459a72334bb733a033514c45d44e39f2/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py#L70

2. Historically the submitit mem param was in the form xxxGB, what is the recommended form and type for mem?

plugins/hydra_submitit_launcher/example/config.yaml
```diff
@@ -1,4 +1,11 @@
 defaults:
   - hydra/launcher: submitit
+hydra:
+  launcher:
+    params:
+      queue: auto
+      queue_parameters:
+        auto:
+          cluster: local
```
omry, line 10: This seems like a very verbose example.
What is the objective here? Is the current example not running locally by default?

I would prefer to keep the basic example using local explicitly.
We also need actual slurm testing here and a dedicated slurm example, but this is beyond the scope of this diff.

plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py
omry: The current design of this plugin is that one config can do all three modes by selecting the mode with queue_type.

An alternative design is that we have 3 different config group options:
```
hydra/launcher:
  submitit_slurm
  submitit_local
  submitit_auto
```

and have the user select the one they need. it should come with the basic defaults  for that mode.
This is what we ended up with for the Ray launcher (WIP in PR) from @jieru-hu.
This can be a followup diff, it is a user facing change and we should think about it carefully.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jrapin: 1. Oh I did not realize that would be here thanks
2. the slurm executor accepts the slurm `mem` parameter, so a string which ca be xxxGB or xxxMB etc. We wanted to simplify and further type it in the auto executor, hence it is an integer, and always in GB. In my opinion the recommended form is using auto executor :D

plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py
jrapin: I wouldn't be the best person for this, it's up to you but probably better not mix it with this diff

plugins/hydra_submitit_launcher/example/config.yaml
jrapin: It was using the local executor, not through the auto executor though. Ends up being the same but I feel that having the whole hierarchy explicitely written helps a lot when starting from nothing, it probably took me 15min to get there

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -16,35 +17,85 @@ class QueueType(Enum):
...
     # check the following for more info on slurm_max_num_timeout
     # https://github.com/facebookincubator/submitit/blob/master/docs/checkpointing.md
     slurm_max_num_timeout: int = 0
-    gpus_per_node: int = 1
-    timeout_min: int = 60
+    # USR1 signal delay before timeout for the slurm queue
+    slurm_signal_delay_s: int = 30
+    # slurm partition to use on the cluster
+    # (depending on the cluster configuration, default "dev" may not exist)
+    slurm_partition: str = "dev"
```
omry: leftover dev.
(check the code and docs to see there are no others?)

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
omry: This config is for the slurm queue so it makes sense that it passes the expected string (xxxGB).

Are you suggesting that we remove support for the local and slurm queues in favor of auto?
In my opinion you will discover that having a single queue to configure many others will become unsustainable if you add additional backends.

Hydra can better handle this kind of variability through config group options.

jrapin: > Are you suggesting that we remove support for the local and slurm queues in favor of auto?

That's what I have stating from the start :D 

> you will discover that having a single queue to configure many others will become unsustainable if you add additional backends.

I'd say it the other way round, if we had many backend and one would require writting nearly similar configurations for each one, it can be error prone. the aim with auto executor is to share the parameters which are more widely used (currently: gpus_per_nodes, nodes, tasks_per_nodes, name, cpus per tasks, time out). Each executor may have a different way of handling this independently (for now each specific executor has been sticking close to the cluster parameter name), but the auto executor allows unification.
At some point we could support 2 clusters, say "slurm" and "cluster2", most of the parameters we would use would be the ones mentioned above. Only 2 things were different if we wanted to run a job: slurm_partition/cluster2_hostgroup (same purpose, each with cluster specific options), and the `folder` for the xp (because the storage structure was different).

So the same code would run on both clusters without requiring duplicated options.

If ever we need to override one shared option, we could with something like `slurm_gpus_per_node` for instance. Not sure it currently works (I'll crosscheck) but that is one of the things @gwenzek changes have recently enabled.
I know that it looks similar to having different configurations as hydra enables, the thing is, in my opinion with the executors, 99% of times you would want the *equivalent* options on the other cluster, which auto executor can convert automatically

omry: > That's what I have stating from the start :D

Not really, you have been saying people should use auto. not to remove slurm and local.

> I'd say it the other way round, if we had many backend and one would require writting nearly similar configurations for each one, it can be error prone

This can be addressed through interpolation. but we need to decide on the exact recommended structure. Once we have a standard individual launchers can translate from the standard to the what is expected by the system it's using. This also makes it easy for people to switch between different launchers (fairtask).

Here is a proposal:
1. Hydra adds job specific resources it the core, containing the common things: cpus, memory, gpus.
2. Submitit creates a SLRUM config node that can potentially be shared with other SLURM launchers.

I am going to redefine the names under the assumption that:
* A node is a host running multiple tasks
* A task is a specific run with resources requirements

Job specific resources in Hydra core config (a job here represent a single invocation of the user function, this is existing Hydra terminology):

These can be shared by all non trivial launchers (submitit, fairtask, ray, joblib):

```yaml
hydra:
  job:
    # Job requested resources.
    # Not all launchers supports these.
    resourecs:
      cpu_cores: 1
      max_mem_gb: 10
      num_gpus: 0
```

Common SLURM config.
Some things are inheriting the values from hydra.job.resources.
If we end up using this for other SLRUM launchers (fairtask), we can promote it out of the submitit plugin.

slurm_common.yaml
```yaml
hydra:
  slurm:
    # number of nodes to allocate for the job
    num_nodes: 1
    # using job to align with Hydra terminology, we can use task too.
    # Number of user function runs on the each node
    jobs_per_node : 1
    gpus_per_job: ${hydra.job.resoures.num_gpus}
    cpu_cores_per_job: ${hydra.job.resoures.cpu_cores}
    job_timeout_minutes: 1
```
hydra/launcher/submitit_slurm.yaml:
```yaml
hydra:
  launcher:
    cls: SubmititSlurmLauncher
    params:
      num_nodes: ${hydra.slurm.num_nodes}
      jobs_per_node : ${hydra.slurm.jobs_per_node}
      gpus_per_job: ${hydra.slurm.gpus_per_job}
      cpu_cores_per_job: ${hydra.slurm.cpu_cores_per_job}
      job_timeout_minutes: ${hydra.slurm.job_timeout_minutes}
```

hydra/launcher/submitit_local.yaml:
```yaml
hydra:
  launcher:
    cls: SubmititLocalLauncher
    params:
      # I don't think the local launcher actually supports these things, but let's pretend.
      # Since local is not SLURM, this can use directly from hydra.job.resources:
      num_gpus: ${hydra.job.resoures.num_gpus}
      cpu_cores: ${hydra.job.resoures.cpu_cores}
```

Thoughts?

jrapin: I was considering something somehow equivalent:
1. Only use AutoExecutor, since it encapsulate everything
2. Move all the shared config (gpus_per_node, mem_gb etc) to one base config (this basically looks like your hydra-scope config
3. Having separate config classes for each cluster (that would basically only be slurm right now) with cluster specific options and *possibly* some overrides of the basic config. In practice I would definitely not like keeping all options there, so users should be able to add parameters that are not present in it.

Considering your proposition, the names are incorrect in my opinion. A job has the similar definition in both hydra and slurm, you're missing the distinction of multi-tasks and multi-node jobs. Currently that is:

```yaml
hydra:
  job:
    # Job requested resources.
    # Not all launchers supports these.
    resources:
      nodes: 1
      tasks_per_node: 1
      cpus_per_task: 1
      mem_gb: 10
      gpus_per_node: 0
      timeout_min: 10
```
`nodes` and `tasks_per_node` are necessary to run multi-node and multi-task jobs so the distinction is important and the need to define cpus and gpus in consequence. The only element I would rename (but for some minor name clarifications) is `gpus_per_node` -> `gpus_per_task`. We have not done it yet for historical reasons and to avoid introducing too many changes with the open source release, but it's planned.

omry: > Having separate config classes for each cluster

This becomes trivial with Structured Configs.

> Considering your proposition, the names are incorrect in my opinion. A job has the similar definition in both hydra and slurm, you're missing the distinction of multi-tasks and multi-node jobs

Hydra does not support multi-task (what is that) or multi-node jobs.
Those are properties of the Submitit (and potentially other SLURM based launchers).

plugins/hydra_submitit_launcher/example/config.yaml
omry: Specific plugin configuration discoverability is not great right now.
You can get the Hydra config (which contains them using `--cfg hydra`.

plugins/hydra_submitit_launcher/example/config.yaml
```diff
@@ -1,4 +1,11 @@
...
```
omry, line 9: Please remove this from the example.
You can add another example showing how to to use the auto queue if you want.

The default queue should be slurm like in the [original plugin](https://github.com/fairinternal/hydra-fair-plugins/blob/master/plugins/hydra-submitit/hydra_plugins/submitit/conf/hydra/launcher/submitit.yaml#L6) (not local as is the case now).
please change that too (it's in config.py).

The test is currently overriding the queue to local so it should continue to work.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jrapin: > In my proposal there is also no Hydra support for cpu, mem and gpus

I don't get what you mean, that looks exactly like it does.

How do you make interpolation work with dataclasses when you need for instance the multiplication of 2 parameters as an integer as well?

omry: > I don't get what you mean, that looks exactly like it does.

Where does it look like that?
If the answer is in the above snippet, this is not something in Hydra now.
and even if it gets in, it will only mean that a job can declare it's resource requirements. different launchers would still need to respect that.

I think details like supporting multiple jobs on the same node (multitask) or not on different nodes (multi-node?) are details that are more specific to SLRUM which are not as generic.

> How do you make interpolation work with dataclasses when you need for instance the multiplication of 2 parameters as an integer as well?

Currently OmegaConf does not support arithmetic operations.
This is something I will consider in a future version ([task](https://github.com/omry/omegaconf/issues/91)).

In this case, this kind of logic can be handled by the Launcher itself before it's calling submitit.

jrapin: > In this case, this kind of logic can be handled by the Launcher itself before it's calling submitit.

I would not be the one doing that, I hate duplicated work. In my opinion after getting more understanding of the plugin, this is done and should stay in the submitit package, and this plugin should only use the AutoExecutor. I'll finish this basic PR without changing the behavior for now anyway, let's chat about this offline

omry: I didn't plan for these changes in this diff. Changes to Hydra are frozen except bug fixes until I start working on 1.1.

(We can of course freely change the plugin).
I agree that we should continue discussing this offline.

website/docs/plugins/submitit_launcher.md
```diff
@@ -39,33 +39,85 @@ class QueueType(Enum):
...
+
+    # maximum time for the job in minutes
+    time: int = 60
+    # number of cpus to use for each task
+    cpus_per_task: int = 10
+    # number of gpus to use on each node
     gpus_per_node: int = 1
+    # number of tasks to spawn on each node
     ntasks_per_node: int = 1
+    # number of nodes to use for the job
```
omry, line 53: Curious about these. I never used them in submitit:

for multitasking, does submitit itself launches the exact same job multiple times on the node?
how can the user control the differences between the jobs?
for example, if each job need a different config overrides?

Hydra 1.0 adds the ability to control the job environment [variables from the config](https://hydra.cc/docs/next/configure_hydra/job#hydrajobenv_set) which can be handy in some of these cases, for example to map the DDP rank_id to the hydra.job.num.

jrapin: > for multitasking, does submitit itself launches the exact same job multiple times on the node?

No `ntasks_per_node` and `nodes` are features of slurm, so we just forward the parameter to sbatch.

But be careful, it does not spawn multiple jobs, it is still a *unique job* (with a unique job id), but split on `nodes` machines, and on each machine the same code runs in `ntasks_per_node` processes/tasks. Each of the process is given a local and a global task rank which is used for synchronization. Environment variables (and submitit helpers) give you access to this.

In practice it's used for multi-GPUs jobs since one process/task will be responsible for one GPU (hence `gpus_per_node` will probably evolve to `gpus_per_task`)

> how can the user control the differences between the jobs?

I think the last 2 paragraph above explains it. There is only one job, running on several processes, and synchronization is done through environment variables (set by slurm in this case) which can be accessed through `submit.helpers.JobEnvironment` to unify between the different clusters (if available...)

jrapin:
> see my comments. if you are not sure what they are I can state the requests again.

Not sure what you mean, I think all your comments were addressed (I just did not comment back, I thought you had seen since you had accepted indeed...)

omry:
re-reviewing, I think I was looking at an old version of some file.

omry:
yes, looks good. merged it in.
sorry about that.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -16,35 +17,83 @@ class QueueType(Enum):
...
+    time: int = 60
+    # number of cpus to use for each task
+    cpus_per_task: int = 10
+    # number of gpus to use on each node
     gpus_per_node: int = 1
+    # number of tasks to spawn on each node
     ntasks_per_node: int = 1
+    # number of nodes to use for the job
+    nodes: int = 1
+    # memory to reserve for the job on each node, in GB
```
omry, line 31: I think at this point we should just inline the memory.
the interpolation here is not helping.

PR #670, opened 2020-06-11
Use structured config in the Ax plugin.

### LINKED ISSUE #469: Use structured config in the Ax plugin

shagunsodhani:
Fixes #469

## Motivation

Use structured config in the Ax plugin.

Yes

## Test Plan

Passes plugin tests

## Related Issues and PRs

#469

### DIFF
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
@@ -13,6 +13,7 @@ from hydra.types import TaskFunction
 from omegaconf import DictConfig, OmegaConf
 
 from ._earlystopper import EarlyStopper
+from .config import AxConfig, ClientConfig, ExperimentConfig
 
 log = logging.getLogger(__name__)
 
@@ -124,12 +125,16 @@ def get_one_batch_of_trials(
 class CoreAxSweeper:
     """Class to interface with the Ax Platform"""
 
-    def __init__(self, ax_config: DictConfig, max_batch_size: Optional[int]):
+    def __init__(self, ax_config: AxConfig, max_batch_size: Optional[int]):
         self.launcher: Optional[Launcher] = None
         self.job_results = None
-        self.experiment = ax_config.experiment
-        self.early_stopper = EarlyStopper(**ax_config.early_stop)
-        self.ax_client_config = ax_config.client
+        self.experiment: ExperimentConfig = ax_config.experiment
+        self.early_stopper: EarlyStopper = EarlyStopper(
+            max_epochs_without_improvement=ax_config.early_stop.max_epochs_without_improvement,
+            epsilon=ax_config.early_stop.epsilon,
+            minimize=ax_config.early_stop.minimize,
+        )
+        self.ax_client_config: ClientConfig = ax_config.client
         self.max_trials = ax_config.max_trials
         self.ax_params: DictConfig = OmegaConf.create({})
         if hasattr(ax_config, "params"):
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
@@ -8,7 +8,6 @@ from hydra.types import MISSING, ObjectConf
 
 @dataclass
 class EarlyStopConfig:
-
     minimize: bool = MISSING
 
     max_epochs_without_improvement: int = 10
@@ -19,7 +18,6 @@ class EarlyStopConfig:
 
 @dataclass
 class ExperimentConfig:
-
     # Experiment name
     name: Optional[str] = None
 
@@ -49,13 +47,9 @@ class AxConfig:
 
     # max_trials is application-specific. Tune it for your use case
     max_trials: int = 10
-
     early_stop: EarlyStopConfig = EarlyStopConfig()
-
     experiment: ExperimentConfig = ExperimentConfig()
-
     client: ClientConfig = ClientConfig()
-
     params: Dict[str, Any] = field(default_factory=dict)
 
 
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
@@ -1,0 +1,77 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional
+
+from hydra.core.config_store import ConfigStore
+from hydra.types import MISSING, ObjectConf
+
+
+@dataclass
+class EarlyStopConfig:
+
+    minimize: bool = MISSING
+
+    max_epochs_without_improvement: int = 10
+
+    # An improvement larger than epsilon is considered significant
+    epsilon: float = 0.00001
+
+
+@dataclass
+class ExperimentConfig:
+
+    # Experiment name
+    name: Optional[str] = None
+
+    # Name of metric to optimize or null if you only return one metric
+    objective_name: str = "objective"
+
+    # Defaults to minimize, set to false to maximize
+    minimize: bool = True
+
+    # For the remaining paramaters, refer the Ax documentation: https://ax.dev/api/core.html#experiment
+    parameter_constraints: Optional[List[str]] = None
+    outcome_constraints: Optional[List[str]] = None
+    status_quo: Optional[Dict[str, Any]] = None
+
+
+@dataclass
+class ClientConfig:
+
+    verbose_logging: bool = False
+
+    # set random seed here to make Ax results reproducible
+    random_seed: Optional[int] = None
+
+
+@dataclass
+class AxConfig:
+
+    # max_trials is application-specific. Tune it for your use case
+    max_trials: int = 10
+
+    early_stop: EarlyStopConfig = EarlyStopConfig()
+
+    experiment: ExperimentConfig = ExperimentConfig()
+
+    client: ClientConfig = ClientConfig()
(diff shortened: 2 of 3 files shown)


plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
```diff
@@ -0,0 +1,77 @@
...
+    # max_trials is application-specific. Tune it for your use case
+    max_trials: int = 10
+
+    early_stop: EarlyStopConfig = EarlyStopConfig()
+
+    experiment: ExperimentConfig = ExperimentConfig()
+
+    client: ClientConfig = ClientConfig()
+
+    params: Dict[str, Any] = field(default_factory=dict)
```
omry, line 59: Be more conservative with new lines between members. add them to separate distinct groups.
If there is only one group don't add newlines in between everything.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+    max_epochs_without_improvement: int = 10
+
+    # An improvement larger than epsilon is considered significant
+    epsilon: float = 0.00001
+
+
+@dataclass
+class ExperimentConfig:
+
```
omry, line 22:
```suggestion
```

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
```diff
@@ -0,0 +1,77 @@
...
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional
+
+from hydra.core.config_store import ConfigStore
+from hydra.types import MISSING, ObjectConf
+
+
+@dataclass
+class EarlyStopConfig:
+
```
omry, line 11:
```suggestion
```

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -124,12 +125,16 @@ def get_one_batch_of_trials(
...
-        self.ax_client_config = ax_config.client
+        self.experiment: ExperimentConfig = ax_config.experiment
+        self.early_stopper: EarlyStopper = EarlyStopper(
+            max_epochs_without_improvement=ax_config.early_stop.max_epochs_without_improvement,
+            epsilon=ax_config.early_stop.epsilon,
+            minimize=ax_config.early_stop.minimize,
+        )
+        self.ax_client_config: ClientConfig = ax_config.client
         self.max_trials = ax_config.max_trials
         self.ax_params: DictConfig = OmegaConf.create({})
```
omry, line 139: can you explain this part?

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+@dataclass
+class ExperimentConfig:
+
+    # Experiment name
+    name: Optional[str] = None
+
+    # Name of metric to optimize or null if you only return one metric
+    objective_name: str = "objective"
```
omry, line 27: seems like a good candidate for Enum

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # Defaults to minimize, set to false to maximize
+    minimize: bool = True
+
+    # For the remaining paramaters, refer the Ax documentation: https://ax.dev/api/core.html#experiment
+    parameter_constraints: Optional[List[str]] = None
+    outcome_constraints: Optional[List[str]] = None
+    status_quo: Optional[Dict[str, Any]] = None
```
omry, line 35: are these optional or required?

Generally I see very little usage of MISSING.

shagunsodhani:
> A few minor comments and a question.
> Overall looks good.
> 
> How much effort this was? did it flow or did you hit many problems?

Overall, it was very smooth. The only thing I was unsure of was when we assign values in config.py, do they override the values in config.yaml or not. I need to dig a bit deeper and will update the documentation once I am done.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
shagunsodhani: You mean why am I passing them explicitly instead of unpacking?

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
shagunsodhani: I am not sure if this should be enum. Looking at the [documentation](https://ax.dev/api/service.html#ax.service.ax_client.AxClient.create_experiment), it seems that the value can be any arbitrary string.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
shagunsodhani: These are optional as the Ax API assigns a default value (None) to them.

shagunsodhani:
> You should probably delete
> hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/conf/hydra/sweeper/ax.yaml
> 
> and remove the searchpath plugin.

Question: In the config file, we do not specify all the possible config params that can be passed to Ax Client. The user can specify them in the config and then will be passed on but they are not mentioned by default. If we are moving to structured configs, I would have to specify all of them in the config.py correct? (otherwise config.py will complain if the user updates config.yaml)

Regarding removing config.yaml, I agree it is redundant (and extra work to keep it in sync with config.py) but for an end user, it may be difficult to create config.yaml by looking at config.py. So I am inclined towards keeping config.yaml (atleast as a reference, even if we do not actually use it).

What happens if config.yaml specifies one value (ie one default value) and config.py specifies another default value? Whose value takes precedence? My understanding is config.yaml but want to double check

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
shagunsodhani: MISSING is used sparsely so that user has to override only the most important things (IMO wether to minimize or maximize). I can use it more often

omry: Nope, this sounds right.

omry:
> Question: In the config file, we do not specify all the possible config params that can be passed to Ax Client. The user can specify them in the config and then will be passed on but they are not mentioned by default. If we are moving to structured configs, I would have to specify all of them in the config.py correct? (otherwise config.py will complain if the user updates config.yaml)

You may be confusing two config files here:
The user cannot put anything in the config file packages with the plugin. they can put it in their own plugin.

When you say "we do not specify all the possible config params that can be passed to Ax Client", this is true both both config.py and ax.yaml.
Please be more specific, I don't understand what yo will have to specify in the config.py if we drop the file that you are not already specifying.

> Regarding removing config.yaml, I agree it is redundant (and extra work to keep it in sync with config.py) but for an end user, it may be difficult to create config.yaml by looking at config.py. So I am inclined towards keeping config.yaml (atleast as a reference, even if we do not actually use it).

This is a general problem with structured configs.
I am thinking of providing a facility to get a subsection of the config in --cfg job|hydra.
for example:
```yaml
$ python foo.py --cfg hydra --package hydra.launcher
hydra:
  launcher:
    ...
```
This was a user will be able to easily see the yaml config file corresponding with a specific package in the config which will help the user creating the config file they are using to configure this and other plugins.

> What happens if config.yaml specifies one value (ie one default value) and config.py specifies another default value? Whose value takes precedence? My understanding is config.yaml but want to double check

Your understanding is correct. the process is documented [here](https://hydra.cc/docs/next/tutorials/structured_config/schema) (although it may not be 100% clear that this is what is happening right now).

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
omry: no, this:
```
self.ax_params: DictConfig = OmegaConf.create({})
if hasattr(ax_config, "params"):
  self.ax_params.update(ax_config.params)
  self.ax_params.update(ax_config.params)
```

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
omry: Ah, I see.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
shagunsodhani: Ohh this is for the case the when the user does not specify the `params` in the config.yaml (so `ax_config.params` is `None`)

shagunsodhani:
I think I was just confusing a few things. Now its clear :)

omry:
Ah, landed prematurely.
can you have a followup diff cleaning the existing config and searchpath plugin?

shagunsodhani:
Ouch okay :) I thought we are keeping the config.yaml Pushing the changes

omry:
Opened this:

https://github.com/facebookresearch/hydra/issues/674

shagunsodhani:
https://github.com/facebookresearch/hydra/pull/675

One change is that I have assigned the default value of `minimize=True` in `config.py` (as was on `config.yaml`)

PR #671, opened 2020-06-12
added fb specific docs

omry:


omry:
@jieru-hu  FYI: moving the fbcode docs here.

PR #675, opened 2020-06-13
[Ax Plugin] Delete conf and search path plugin

shagunsodhani:
## Motivation

Related to #469 

Yes

## Test Plan

Passes tests

## Related Issues and PRs

#469 #670


omry:
As far as I can see you are not deleting the config file.

shagunsodhani:
ouch my bad - thank you for spotting it

omry:
So you see, removing the config file actually exposed that you didn't set the minimize flag to True.

Speaking of which:
This IS an opportunity for a an Enum:

```python
class Optimize(Enum):
  minimize = 1
  maximize = 2
```

but we can stick to the Ax API if we want.

shagunsodhani:
> So you see, removing the config file actually exposed that you didn't set the minimize flag to True.
> 
> Speaking of which:
> This IS an opportunity for a an Enum:
> 
> ```python
> class Optimize(Enum):
>   minimize = 1
>   maximize = 2
> ```
> 
> but we can stick to the Ax API if we want.

Lets stick to Ax API for now.

shagunsodhani:
pop()

Issue #676, opened 2020-06-13
Create unit tests examples with compose


omry:
For:
- pytest
- unittest

Issue #591, opened 2020-05-17
[Bug] Running Hydra with Pytest yields `NotImplementedError` error


vict0rsch:
# 🐛 Bug

Running Hydra with Pytest yields `NotImplementedError` error

## To reproduce

```
directory structure:
	hydra/
		conf/
			__init__.py
			config.yaml
	tests/
		test_hydra.py
```

```
# test_hydra.py 
import unittest
from hydra.experimental import compose, initialize


from pathlib import Path


class FullUnitTest(unittest.TestCase):

    def test_hydra_is_loaded(self):

        conf_path = Path(__file__).parent.parent / "hydra" / "conf"
        conf_path = conf_path.resolve()
        print()
        print(">>>>>>>", conf_path)
        print(">>>>>>>", conf_path.exists())
        print(">>>>>>>", str(conf_path))
        initialize(config_dir=str(conf_path))
        conf = compose(str(conf_path / "config.yaml"))
        print(conf.pretty())

```

** Stack trace/error message **

```
$ pytest -s
============================================================= test session starts ==============================================================
platform darwin -- Python 3.7.4, pytest-5.4.2, py-1.8.1, pluggy-0.13.1
rootdir: /Users/victor/Documents/Experiments/python-exps/tests
collected 1 item                                                                                                                               

test_hydra.py 
>>>>>>> /Users/victor/Documents/Experiments/python-exps/hydra/conf
>>>>>>> True
>>>>>>> /Users/victor/Documents/Experiments/python-exps/hydra/conf
F

=================================================================== FAILURES ===================================================================
______________________________________________________ FullUnitTest.test_hydra_is_loaded _______________________________________________________

is_pkg = True, filename = 'test_hydra//Users/victor/Documents/Experiments/python-exps/hydra/conf'

    @staticmethod
    def _exists(is_pkg, filename):
        if is_pkg:
            module_name, resource_name = ConfigLoader._split_module_and_resource(
                filename
            )
            try:
>               ret = resource_exists(module_name, resource_name)

../exp-env/lib/python3.7/site-packages/hydra/_internal/config_loader.py:365: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[...]

PR #679, opened 2020-06-15
Add pytest example tutorial

packerliu:
## Motivation

add example pytest for hydra.cc

Yes

## Test Plan

No, this is adding example tutorial to add pytest.  Maybe add additional pytest on top of this later.

## Related Issues and PRs

issue #591, issue #676


omry:
Thanks, I have a pending change that is creating an example with pytest directly (not unittest) and is also simplifying some other things (for example, your example will fail if you have more than one call to initialize()).

I will keep this open until I am able to land the other example (pending a response on [this](https://gitlab.com/python-devs/importlib_resources/-/issues/103)).
My example is relying on switching the config source in Hydra from a pkg_resource based one to importlib_resources based one - which will become standard in Python 3.9 and is the long term solution).

I want two of examples:
1. One using simple pytest:
2. One using unittest.

omry:
Thanks for the pull request.
I created a complete unit tests example in https://github.com/facebookresearch/hydra/pull/685

Issue #566, opened 2020-05-04
Recursive instantiation


omry:
this would go well with a change to the config definition of an instantiated object:
### Current
```yaml
model:
  cls: foo.bar.ResNet
  params:
    num_layers: 50
```
### New
```yaml
model:
  _target_: model.ResNet
  num_layers: 50
```

Once this is implemented, we can describe recursive instantiation like:
```yaml
trainer:
  _target_: trainer.Trainer
  batch_size: 32
  model:
    _target_: model.ResNet
    num_layers: 50
```
An remaining open question is how to support passthrough for nested objects.

Currently, passthrough can override individual values:

```python
trainer = instantiate(cfg.trainer, batch_size=64)
```

a possible approach is for nested objects is to use a dictionary:

```python
trainer = instantiate(cfg.trainer, batch_size=64, overrides={"trainer.model.num_layers": 50})
```

shagunsodhani:
Is the New style intended to replace the current one?

omry:
Not entirely. If you don't own the the dataclass you will not be able to add the _type_ field to it.

shagunsodhani:
Okay my original concern was if we could keep around the `params` namespace but if both the styles can be used, that request is moot.

omry:
we may want to deprecate it in favor of 
```
args:
 - positional_1
 - positional_1
kwargs:
  named_1 : 1
  named_2 : 2
```

shagunsodhani:
> we may want to deprecate it in favor of
> 
> ```
> args:
>  - positional_1
>  - positional_1
> kwargs:
>   named_1 : 1
>   named_2 : 2
> ```

Looks good!

Hulk89:
I'm currently use recursive instantiating like this.

## config.yaml
```yaml
string: hello hydra
parent_module:
  _target_: modules.a.A
  __args:
    - ${child_module}
child_module:
  _target_: modules.b.B
  value: ${string}
```

## main.py
```python
import hydra
from omegaconf import DictConfig, OmegaConf

def instantiate(conf, *args, **kwargs):
    if hasattr(conf, "_target_"):
        if '__args' in conf:
            args = list(args) + [instantiate(i) for i in conf.__args]
            primitive_conf = OmegaConf.to_container(conf, resolve=True)
            del primitive_conf['__args']
            conf = DictConfig(primitive_conf)
    return hydra.utils.instantiate(conf, *args, **kwargs)


@hydra.main(config_path='conf', config_name='config')
[...]

PR #682, opened 2020-06-15
[Submitit] Restructure the plugin configuration

jrapin:
## Motivation

Only use AutoExecutor for submitit plugin, since it is the recommended method

## Todo

- [x] clarify parameter names (timeout_min -> timeout_minute(s)? gpus_per_node -> gpus_per_task (forward compatibility). TBD
- [x] clean DictConf instantiation in init
- [x] can we get rid of executor parameter (local/slurm)?
- [x] manual test on slurm
- [x] update documentation

Yes

## Test Plan

To do


plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -7,35 +7,53 @@
...
+class ExecutorName(Enum):
     local = "local"
     slurm = "slurm"
 
 
 @dataclass
-class SlurmQueueConf:
-    # Params are used to configure sbatch, for more info check:
-    # https://github.com/facebookincubator/submitit/blob/master/submitit/slurm/slurm.py
+class BaseSubmititConf:
```
jrapin, line 16: All other configurations derive from this one.
It does not set any value

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -7,35 +7,53 @@
...
-    nodes: int = 1
-    # memory to reserve for the job on each node, in GB
-    mem: str = "${hydra.launcher.mem_limit}GB"
+    nodes: Optional[int] = None
+    # name of the job
+    name: Optional[str] = None
+
+
+@dataclass
+class SubmititSlurmConf(BaseSubmititConf):
```
jrapin, line 37: submitit configuration derives from the base configuration so that it can override any default value
and adds a few slurm specific parameters

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -45,27 +63,12 @@ class SlurmQueueConf:
...
-    tasks_per_node: int = 1
-
-
-@dataclass
-class AutoQueueConf:
-    # auto executor automatically identifies and uses available cluster
-    # Currently this is only slurm, but local executor can be manually forced
-    # instead.
-    # Most parameters are shared between clusters, some can be cluster specific
+class SubmititConf(BaseSubmititConf):
```
jrapin, line 66: This is the main configuration, it provides all the default values, and can be overriden by executor specific configs

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -82,42 +85,16 @@ class AutoQueueConf:
...
-class SubmititConf:
-    queue: QueueType = QueueType.slurm
-
-    folder: str = "${hydra.sweep.dir}/.${hydra.launcher.params.queue}"
-
-    queue_parameters: QueueParams = QueueParams()
+    # additional specific parameters and overrides are
+    # provided here
+    local_executor: BaseSubmititConf = BaseSubmititConf()
+    slurm_executor: SubmititSlurmConf = SubmititSlurmConf()
```
jrapin, line 91: Here are the configs for
- overriding the default/shared values above in a given executor
- adding cluster specific parameters

Since local executor does not have anything more, it just have a plain BaseSubmititConf for overriding the shared configuration

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -45,27 +63,12 @@ class SlurmQueueConf:
...
+    folder: str = "${hydra.sweep.dir}/.${hydra.launcher.params.queue}"
 
-    # cluster to use (currently either "slurm" or "local" are supported,
+    # executor to use (currently either "slurm" or "local" are supported,
     # None defaults to an available cluster)
-    cluster: str = "slurm"
+    executor: ExecutorName = ExecutorName.slurm
 
     # maximum time for the job in minutes
     timeout_min: int = 60
```
jrapin, line 74: by default all the values here are shared for all clusters/executors. But they can always be bypassed

omry:
ask for review when ready.

jrapin:
> ask for review when ready.

I wouldnt call it `ready` but I need help sorting this out. I couldn't make the config group work.

First thing to deal with in my opinion is what the structure of the conf should look like:
for now its:
```
folder
queue
params:
    gpus_per_task
     etc
```

I tried to change it to something like the following, removing one level
```
folder
queue
gpus_per_task
etc
```
but that makes the init of the plugin a bit more complicated.

When that is dealt with, the second question will of course be "how to change from local to slurm"

omry:
The plugin is instantiated using hydra.util.instantiate.
The structure it expects is [ObjectConf](https://github.com/facebookresearch/hydra/blob/master/hydra/types.py#L10-L17).
This is where the first params is coming from. This is not flexible.

There is an open rfc to eliminate that and support recursive instantiation (https://github.com/facebookresearch/hydra/issues/566).
For now, keep the first params.

Try something like this (note that I renamed *Conf to *Params to reflect that I want to use them as the params object.

```python
@dataclass
class BaseParams:
    # Where does this belong? maybe in base? 
    # if not move it to one of the the two specific configs
    folder: str = "${hydra.sweep.dir}/.${hydra.launcher.params.executor}"
    ...

@dataclass
class SlurmParams(BaseParams):
   ...

@dataclass
class LocalParams(BaseParams):
    ...


@dataclass
class SlurmConf:
    params: SlurmParams = SlurmParams()

@dataclass
class LocalConf:
    params: LocalParams = LocalParams()

# finally, register two different choices:
ConfigStore.instance().store(
    group="hydra/launcher",
    name="submitit_local",
    node=LocalConf,
    provider="submitit_launcher",
)

ConfigStore.instance().store(
    group="hydra/launcher",
    name="submitit_slurm",
    node=SlurmConf,
    provider="submitit_launcher",
)
```

There is a similar example [here](https://github.com/facebookresearch/hydra/blob/master/examples/tutorials/structured_configs/3_config_groups/my_app_with_inheritance.py) (without the ObjectConf).

omry:
An alternative style which might be easier to read is this:
(Not sure which one I prefer yet).
I [do this in Hydra's config](https://github.com/facebookresearch/hydra/blob/master/hydra/conf/__init__.py#L87-L99) when it's not getting too deep.


```python
@dataclass
class BaseParams:
  ...

@dataclass
class SlurmConf:
  @dataclass
  class SlurmParams(BaseParams):
     ...

  params: SlurmParams = SlurmParams()

@dataclass
class LocalConf:
  @dataclass
  class LocalParams(BaseParams):
      ...

  params: LocalParams = LocalParams()
```

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -7,65 +7,18 @@
...
-class AutoQueueConf:
-    # auto executor automatically identifies and uses available cluster
-    # Currently this is only slurm, but local executor can be manually forced
-    # instead.
-    # Most parameters are shared between clusters, some can be cluster specific
-
-    # cluster to use (currently either "slurm" or "local" are supported,
-    # None defaults to an available cluster)
-    cluster: str = "slurm"
+    executor: ExecutorName = ExecutorName.slurm  # can we get rid of this?
```
jrapin, line 20: Is this needed here? Is there another way to get this information since we now call `submitit_local` or `submitit_slurm`. That would prevent a user from setting it to the wrong value.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -33,10 +35,12 @@ def manipulate_search_path(self, search_path: ConfigSearchPath) -> None:
...
-    def __init__(self, queue: str, folder: str, queue_parameters: DictConfig) -> None:
-        self.queue = queue
-        self.queue_parameters = queue_parameters
-        self.folder = folder
+    def __init__(self, **params: Any) -> None:
+        self.params = OmegaConf.structured(
+            LocalParams if params["executor"].value == "local" else SlurmParams
+        )
+        for key, val in params.items():
+            OmegaConf.update(self.params, key, val if key != "executor" else val.value)
```
jrapin, line 43: How can this be simplified?

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -96,36 +100,40 @@ def launch(
...
-                x: y
-                for x, y in queue_parameters[self.queue.value].items()
-                if x not in init_parameters
+                f"{exec_name}_{x}": y
+                for x, y in params.items()
+                if x in specific_init_keys
             }
         )
+        init_keys = specific_init_keys | set(init_renamer)  # used config keys
+        executor = submitit.AutoExecutor(**init_params)
```
jrapin, line 121: Now everything goes through AutoExecutor, whatever the choice of the executor. This makes it more maintainable.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -82,47 +35,65 @@ class AutoQueueConf:
     # name of the job
     name: str = "${hydra.job.name}"
 
-    # following parameters are SLURM specific
 
+@dataclass
+class SlurmParams(BaseParams):
+    """Slurm configuration overrides and specific parameters
+    """
```
jrapin, line 42: for slurm, some names will have changed now: `ntasks_per_node` -> `tasks_per_node` etc so that both slurm and local use the same terminology (the one used in AutoExecutor)

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -7,65 +7,18 @@
...
-    # Most parameters are shared between clusters, some can be cluster specific
-
-    # cluster to use (currently either "slurm" or "local" are supported,
-    # None defaults to an available cluster)
-    cluster: str = "slurm"
+    executor: ExecutorName = ExecutorName.slurm  # can we get rid of this?
+    submitit_folder: str = "${hydra.sweep.dir}/.${hydra.launcher.params.executor.value}"
 
     # maximum time for the job in minutes
     timeout_min: int = 60
```
jrapin, line 24: we'll need to decide if we update some of the names for more clarity (timeout_min-> timeout_minute(s)?) or forward compatibility (gpus_per_node-> gpus_per_task?)

plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py
```diff
@@ -33,51 +34,12 @@ class TestSubmititLauncher(LauncherTestSuite):
...
-                                "auto": {
-                                    "cluster": "local",
-                                    "gpus_per_node": 0,
-                                    "tasks_per_node": 1,
-                                    "timeout_min": 1,
-                                }
-                            },
-                        },
-                    }
+                    "launcher": {"params": {"gpus_per_node": 0, "timeout_min": 1}},
```
jrapin, line 42: actually this is not useful, that's just here to prove that it works

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
omry: for me timeout_min is just as clear as timeout_minutes.

As for future compatibility: your call.
you will be able to transition from gpus_per_node-> gpus_per_task by manually deprecating it:

1. add support for both and a comment that gpus_per_node is deprecated in the config.
2. if the user provide gpus_per_node, issue a UserWarning and convert it to gpus_per_task, if the user provides gpus_per_task  use that.
3. after a few months we can remove gpus_per_node completely.

This strategy may be too painful to implement for a plugin, but this is what I am trying to do with Hydra itself.
The alternative is to just have a release that is breaking it at once, which will have users coming with pitchforks asking about how to fix it.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
omry: Right now you are using this as the flag to gate slurm or local initialization logic.
An alternative I think will be much cleaner is to use a different class. I will comment below.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -63,26 +66,34 @@ class SlurmSubmititConf(BaseSubmititConf):
...
-    params: LocalSubmititConf = LocalSubmititConf()
+class SlurmConf(ObjectConf):
+    cls: str = "hydra_plugins.hydra_submitit_launcher.submitit_launcher.SubmititLauncher"
+    params: SlurmParams = SlurmParams()
 
 
 @dataclass
-class SubmititLauncherConf(ObjectConf):
+class LocalConf(ObjectConf):
     cls: str = "hydra_plugins.hydra_submitit_launcher.submitit_launcher.SubmititLauncher"
```
omry, line 81: You can define a launcher class for each mode and have and have the custom logic in it.
They can share a superclass or otherwise reuse most of the logic through other means.
```python
  cls: str = "hydra_plugins.hydra_submitit_launcher.submitit_launcher.SubmititLocalLauncher"
```
This will also allow you to named parameters in the `__init__` of the class:
```
class SubmititLauncher(Launcher):
    def __init__(self, 
      partition: Optional[str],
      comment: Optional[str],
      constraint: Optional[str],
      exclude: Optional[str],
      ...
) -> None:
        # forward directly to the implementation
        # (my preference would be to SlurmExecutor and LocalExecutor at this point).
        ...
```

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
omry: As you can tell, it takes some trickery to prepare the correct dictionary for the AutoExecutor.
The reason is that the input config is no longer lining up with AutoExecutor parameters.

If you used SlurmExecutor and LocalExecutor directly, with corresponding SubmititSlurmLauncher and SubmititLocalLauncher as I suggested above I think you would be able to more easily match the init signature of each to what you want to pass the respective submitit launcher.

You can still use auto, but at least in this case the logic in each of the launchers will need to initialize a known mode which will simplify things

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -94,33 +100,40 @@ def launch(
...
+
+        # specify resources/parameters
+        baseparams = set(dataclasses.asdict(BaseParams()).keys())
+        print(baseparams)
```
omry: don't forget the prints.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
jrapin: > If you used SlurmExecutor and LocalExecutor directly, with corresponding SubmititSlurmLauncher and SubmititLocalLauncher as I suggested above I think you would be able to more easily match the init signature of each to what you want to pass the respective submitit launcher.

I *won't* call `SlurmExecutor` and `LocalExecutor` directly, that would make the plugin harder to maintain. The alignment will actually be much simpler to maintain. Also, there's a PR on its way in submitit that would make it even easier.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jrapin: > for me timeout_min is just as clear as timeout_minutes.

Then lets keep it that way

> you will be able to transition from gpus_per_node-> gpus_per_task by manually deprecating it:

If that kind of transition is fine for you, since it's basically what will happen in the submitit package it will be definitely easy to handle here. So let's keep it to what's in the package then.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jrapin: I'd rather not double down the name of each parameter both in the config and in the init. Naming them here would just add code for nothing in my opinion (we don't need type hints here since it's never called directly by a user, and all the checks can be efficiently performed through the config). So is there a shortcut to initialize the omega conf from a dictionary? I couldnt find any without requiring a for loop

omry: You asked how to simplify. this is my advice.
I don't think you will need to double much:
between config inheritance, and passing the actually needed parameters into each init there will be no duplication in my opinion.

regardless, it's your call.
Changing submitit can also help.

jrapin: but is there a better way to build the OmegaConf structure from a dict. Currently I am doing:
```python
        self.params = OmegaConf.structured(LocalParams)
        for key, val in params.items():
            OmegaConf.update(self.params, key, val)
```
bur I would think I am missing something (and it would not work with Enum, but fortunately there is no more Enum now).
Otherwise I'll use the Config directly but I prefer the flexibility of OmegaConf

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
omry: absolutely.
we can support both temporarily and issue a warning if someone uses the old one until we are comfortable nuking it.
Now is a chance to do some serious clean up but normally we will want to maintain compatibility.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -34,13 +34,16 @@ def manipulate_search_path(self, search_path: ConfigSearchPath) -> None:
...
-class SubmititLauncher(Launcher):
+class LocalSubmititLauncher(Launcher):
+
+    _EXECUTOR = "local"
+
     def __init__(self, **params: Any) -> None:
         self.params = OmegaConf.structured(
-            LocalParams if params["executor"].value == "local" else SlurmParams
+            LocalParams if self._EXECUTOR == "local" else SlurmParams
         )
```
omry: I suggest adding a base class to express that the slurm launcher is not REALLY a local launcher.

```python
# your current local:
class BaseSubmititLauncher(Launcher):
  def _get_params_class() -> Type[BaseParams]:
     raise NotImplementedError

  ... # the rest is the same.

class LocalSubmititLauncher(BaseSubmititLauncher)
  def _get_params_class() -> Type[BaseParams]:
     return LocalParams 

class SlurmSubmititLauncher(BaseSubmititLauncher)
  def _get_params_class() -> Type[BaseParams]:
     return SlurmParams
```

You can go back to a DictConfig with something like this (and now you don't need the update loop)
```python
self.params = OmegaConf.structured(self._get_params_class(**params))
```

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -34,13 +34,16 @@ def manipulate_search_path(self, search_path: ConfigSearchPath) -> None:
...
         for key, val in params.items():
-            OmegaConf.update(self.params, key, val if key != "executor" else val.value)
+            OmegaConf.update(self.params, key, val)
```
omry: I am surprised that worked in the previous version when you use the actual object and not a DictConfig.

regardless, you can get rid of this loop with the example I gave you above.

plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py
omry: how is this proving it works? I don't think it's validating that the configs are actually making it all the way in.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -100,37 +103,36 @@ def launch(
...
             if x not in init_keys
         }
         executor.update_parameters(**params)
 
         log.info(
-            f"Submitit '{exec_name}' sweep output dir : "
+            f"Submitit '{self._EXECUTOR}' sweep output dir : "
             f"{self.config.hydra.sweep.dir}"
         )
         sweep_dir = Path(str(self.config.hydra.sweep.dir))
```
omry: I hope this part can be simplified once you make some changes to submitit.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
omry:
```
self.params = OmegaConf.structured(LocalParams(**params))
```

Keep in mind that dataclasses have a problem that causes them to not do what you expect if you instantiate:

```python

@dataclass
class A:
  a : int = 10

@dataclass
class B(A):
  a: int = 20

@dataclass
class C(A):
  a = 30

print(A())
print(B())
print(C())
```

```
$ python 1.py 
A(a=10)
B(a=20)
C(a=10)
```

If this is causing issues we can think of something.

omry: The general answer to:
> but is there a better way to build the OmegaConf structure from a dict

is:
```python
OmegaConf.create({})
```

In this example you want to get the type safety from the dataclass.
Here is some context:
https://omegaconf.readthedocs.io/en/latest/structured_config.html#merging-with-other-configs

You can also try what I suggested, which is possibly a better match here.

plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py
jrapin: it's not validating anything, it basically proves that hydra works and accepts the parameters, nothing more :D
I like keeping this around so I can try with other overrides easily if need be

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
jrapin: It would get only marginally simpler: getting rid of the need for baseparams.
Then again it's not that complicated, and that should be robust.

jrapin:
@omry it's now manually tested and working on SLURM cluster.
In my opinion it seems ready, let me know if it requires more updates.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -1,71 +1,17 @@
...
-class AutoQueueConf:
-    # auto executor automatically identifies and uses available cluster
-    # Currently this is only slurm, but local executor can be manually forced
-    # instead.
-    # Most parameters are shared between clusters, some can be cluster specific
-
-    # cluster to use (currently either "slurm" or "local" are supported,
-    # None defaults to an available cluster)
-    cluster: str = "slurm"
+    submitit_folder: str = "${hydra.sweep.dir}/.submitit/%j"
```
omry, line 14: Let's talk about this for a moment:
With Hydra, each job is getting a working directory inside the sweep dir.

If we can keep all the files for the job in that directory it would be really good (users will have an easier time debugging issues for a particular job).
This may take some submitit changes, but is something to consider for the future.

Right now, if a job is failing the user will need to:
1. Look at the log file in the job output directory.
2. Look in the submitit output folder, figure out which files belong to the job that caused the issue somehow and check the output/error files.

If we managed to use the working directory here it would be better.
I suspect currently this is not possible without changes to submitit.

An alternative solution to this problem is to symlink the stdout and stderr from the submitit folder into the job working directory. But this can also be tricky.

This is not a feature request for this PR, but something to think about to improve the user experience.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -96,36 +101,33 @@ def launch(
...
-            folder=self.folder,
-            **{
-                x: y
-                for x, y in queue_parameters[self.queue.value].items()
-                if x in init_parameters
-            },
-        )
-        executor.update_parameters(
+        # build executor
+        init_params = dict(folder=params.submitit_folder)
```
omry, line 107: nit:
style wise, I have grown to prefer maps:
```suggestion
        init_params = {"folder" : params.submitit_folder}
```

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -96,36 +101,33 @@ def launch(
...
-                x: y
-                for x, y in queue_parameters[self.queue.value].items()
-                if x not in init_parameters
+                f"{self._EXECUTOR}_{x}": y
+                for x, y in params.items()
+                if x in specific_init_keys
             }
         )
+        init_keys = specific_init_keys | {"submitit_folder"}
+        executor = submitit.AutoExecutor(cluster=self._EXECUTOR, **init_params)
```
omry, line 117: this line is confusing to me:

The auto executor is expecting the strings "local" or "slurm" as options for cluster?

website/docs/plugins/submitit_launcher.md
```diff
@@ -23,73 +23,20 @@ Once installed, add `hydra/launcher=submitit` to your command line. Alternativel
...
 defaults:
-  - hydra/launcher: submitit
+  - hydra/launcher: submitit_slurm
 ```
 
 Note that this plugin expects a valid environment in the target host. usually this means a shared file system between
 the launching host and the target host.
 
-Submitit supports 3 types of queues: auto, local and slurm. Its config looks like this
+Submitit actually implements 2 different launchers: `submitit_slurm` to run on a SLURM cluster, and `local` to test locally.
```
omry, line 32: What do you think about the following?
1. Include links to config.py with dataclasses instead of copying them in.
2. include the outputs of
```
$ python foo.py hydra/launcher=submitit_slurm --cfg hydra -p hydra.launcher
$ python foo.py hydra/launcher=submitit_local-cfg hydra -p hydra.launcher
```
When describing each of the two launchers.
I am still trying to discover what is the best way to document plugins with structured configs, and I think I like the above proposal.

omry: Ah, I saw that you already added similar output.
Also since the primary use case is SLURM we can focus on it.

This will output the current slurm parameters:
```
python foo.py hydra/launcher=submitit_slurm --cfg hydra -p hydra.launcher
```

and I recommend that instead of inlining the dataclasses we can just give a link.

Also, an example of how to override the launcher parameters from the command line would be helpful:

```
python foo.py hydra/launcher=submitit_slurm hydra.launcher.params.partition=prod
```

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
omry: ok.
We can potentially simplify by moving the @dataclass into submitit later. definitely good enough for now.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jrapin: > If we can keep all the files for the job in that directory it would be really good

Indeed (for now this is the same behavior as before, but I agree it is not convenient)

> I suspect currently this is not possible without changes to submitit.

It would actually be possible if not using arrays, but that's not a good option. With arrays on the other hand we can't specify the exact folder of each run individually, it must be deduced from the parameters of the array. Hacking something is somehow possible but would require a lot of (mostly unhelpful) changes in submitit  and would not be robust anyway.

The other option would be that hydra uses the folder provided by submitit (using the job_id instead of 0, 1, 2 etc). But that's probably not something you'd want either, right?

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
jrapin: I used to always use maps, but recently changed my mind since now I find it's just more verbose (always requiring a string...). I'm fine with what you prefer anyway

website/docs/plugins/submitit_launcher.md
jrapin: > Ah, I saw that you already added similar output.

Yes I find them the most interesting since they make it easier to discover settings than just stating the config classes.

> and I recommend that instead of inlining the dataclasses we can just give a link.

I have shared feelings about this, since the interest of adding the config files there came from the description of each parameters. Then again, since it often ends up being outdated if not careful, it's probably better to just have the link.

So I just:
- removed the config classes
- added a link the the config classes
- added the discovery of submitit_local parameters
- added a commandline override example

jrapin:
ci/circleci: py36_win seems broken :s

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
omry: The user can customize the sweep subdirectory through Hydra.
So yeah, I can't really use the submitit generated dir.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -82,47 +28,65 @@ class AutoQueueConf:
...
-    params: SubmititConf = SubmititConf()
-    # memory to reserve for the job on each node, in GB
-    mem_limit: int = 2
+# finally, register two different choices:
+ConfigStore.instance().store(
+    group="hydra/launcher",
+    name="submitit_local",
+    node=LocalConf,
+    provider="submitit_launcher",
+)
```
omry, line 84: I think we can lose LocalConf and SlurmConf class definitions without losing functionality or clarity.
```
ConfigStore.instance().store(..., node=ObjectConf(cls: str ="XYZClass", params = XYZParams()))
```

omry:
I increased the no-output timeout for windows. can you rebase and try again?
also see my comment about simplifying config.py a bit.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jrapin: I may have done something wrong but I get:
omegaconf.errors.UnsupportedValueType: Value 'LocalParams' is not a supported primitive type

omry: Is LocalParama a @dataclass?

jrapin: its a BaseParams, which is a dataclass, does it need a decorator as well? I'm not really familiar with dataclasses

omry: Try to create a minimal example showing the problem with OmegaConf directly.
this seems to work and is very close to what you are doing.

```python
In [1]: from dataclasses import dataclass
In [9]: from typing import Any

In [2]: from omegaconf import OmegaConf

In [3]: @dataclass
   ...: class Foo:
   ...:     num : int = 10
   ...: 

In [4]: @dataclass
   ...: class LocalFoo(Foo):
   ...:     name: str = "name"
   ...: 

In [5]: OmegaConf.structured(LocalFoo)
Out[5]: {'num': 10, 'name': 'name'}

In [6]: OmegaConf.structured(LocalFoo())
Out[6]: {'num': 10, 'name': 'name'}

In [7]: OmegaConf.structured({"add-hoc":LocalFoo()})
Out[7]: {'add-hoc': {'num': 10, 'name': 'name'}}



In [10]: @dataclass
    ...: class ObjectConf:
    ...:     cls: str = "abc"
    ...:     params: Any = None
    ...: 

In [11]: OmegaConf.structured(ObjectConf(cls="xyz", params=LocalFoo()))
Out[11]: {'cls': 'xyz', 'params': {'num': 10, 'name': 'name'}}
```

omry: Example from someone else (he didn't complain so I think it's working for him):

https://github.com/anthonytec2/pytorch-lightning/pull/1/files/6b8d535b6eef983184b95beb3aec7838f6df41ab..09c393501a2c7ba451f2fc15e8d91b608fe2e770

omry:
going to land it, I am having some test failures with the existing submitit plugin after making some core changes and it will be best to fix them on top of your changes.
I will try to simplify the config as I suggested.

omry:
One thing I noticed that is missing is a news fragment. especially for breaking changes like this one we should add them.
I will do it for this one.

website/docs/plugins/submitit_launcher.md
omry: At some point I will support field description in the help or --cfg. when that happens we will be able a better description here.

jrapin:
> One thing I noticed that is missing is a news fragment. especially for breaking changes like this one we should add them.
> I will do it for this one.

Thanks for this and for merging ;)

PR #686, opened 2020-06-16
save conf before task execution

### LINKED ISSUE #678: Save .hydra/*.yaml before task execution
# 🐛 Bug

Saving the configurations is a really cool idea that helps me check what I actually did. I noticed the save timing is delayed after task execution in the `1.0.0rc1`. I have encountered some scenarios that I think it would be better if the configs are saved earlier.

For example,

1. Tasks may exit accidentally due to bugs. Knowing detailed configurations would be helpful.
2. For time-consuming tasks like ML training which may last for several days, it would be helpful to be able to check the config.yaml. This helps me a lot when I have multiple tasks running but I can't remember the real difference. I do `diff config.yaml`.
3. Some tasks don't need to be finished. The full training may be designed to run 100 epochs but it also happens it's aborted/killed in the middle.

## To reproduce

Default behavior in `1.0.0rc1`.

NA

NA

## Expected Behavior
NA

## System information
- **Hydra Version** :  1.0.0rc1
- **Python version** : NA
- **Virtual environment type and version** : NA
- **Operating system** : NA

binshengliu:
## Motivation

Save conf before task execution which was changed in 98128a4d.

Yes

## Test Plan

NA

## Related Issues and PRs

Fix #678

### DIFF
--- a/hydra/core/utils.py
+++ b/hydra/core/utils.py
@@ -112,6 +112,12 @@ def run_job(
         hydra_cfg = OmegaConf.masked_copy(config, "hydra")
         assert isinstance(hydra_cfg, DictConfig)
 
+        if config.hydra.output_subdir is not None:
+            hydra_output = Path(config.hydra.output_subdir)
+            _save_config(task_cfg, "config.yaml", hydra_output)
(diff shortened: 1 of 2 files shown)


omry:
Thanks!

PR #730, opened 2020-06-29
Upgrade Docusaurus v2 + try to fix Prism

slorber:
## Motivation

Attempt to fix highlighting for python

see https://github.com/facebook/docusaurus/issues/3004

Yes

## Test Plan

No

---

http://localhost:3000/docs/next/tutorials/structured_config/minimal_example

image

http://localhost:3000/docs/next/tutorials/structured_config/nesting

image

(the imports have the class "keyword", it's just not styled by the theme, so you could improve that).

### DIFF
--- a/website/sidebars.js
+++ b/website/sidebars.js
@@ -5,7 +5,7 @@ function FBInternalOnly(elements) {
 }
 
 module.exports = {
-    Docs: {
+    docs: {
         About: [
             'intro',
             'terminology',
--- a/website/package.json
+++ b/website/package.json
@@ -10,12 +10,15 @@
     "deploy": "docusaurus deploy"
   },
   "dependencies": {
-    "@docusaurus/core": "^2.0.0-alpha.56",
-    "@docusaurus/preset-classic": "^2.0.0-alpha.56",
+    "@docusaurus/core": "^2.0.0-alpha.58",
+    "@docusaurus/preset-classic": "^2.0.0-alpha.58",
     "classnames": "^2.2.6",
     "docusaurus-plugin-internaldocs-fb": "^0.3.3",
     "react": "^16.8.4",
     "react-dom": "^16.8.4"
+  },
+  "resolutions": {
+    "prism-react-renderer": "1.1.0"
   },
   "browserslist": {
     "production": [
--- a/website/versioned_sidebars/version-0.11-sidebars.json
+++ b/website/versioned_sidebars/version-0.11-sidebars.json
@@ -1,5 +1,5 @@
 {
-  "version-0.11/Docs": [
+  "version-0.11/docs": [
     {
       "type": "category",
       "label": "About",
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -972,10 +972,10 @@
   resolved "https://registry.yarnpkg.com/@csstools/convert-colors/-/convert-colors-1.4.0.tgz#ad495dc41b12e75d588c6db8b9834f08fa131eb7"
   integrity sha512-5a6wqoJV/xEdbRNKVo6I4hO3VjyDq//8q2f9I6PBAvMesJHFauXDorcNCsr9RzvsZnaWi5NYCcfyqP1QeFHFbw==
 
-"@docusaurus/core@^2.0.0-alpha.56":
-  version "2.0.0-alpha.56"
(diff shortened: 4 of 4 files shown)


website/sidebars.js
```diff
@@ -5,7 +5,7 @@ function FBInternalOnly(elements) {
 }
 
 module.exports = {
-    Docs: {
+    docs: {
```
slorber, line 8: The docs plugin expect a sidebar with name 'docs', not 'Docs', this is why your site breaks on upgrade. We should make this more robust in the future.

slorber:
Note, it's actually upgrading the dependency from 1.1.0 to 1.1.1 that seems to break the highlighting, so I forced the former using:

```
  "resolutions": {
    "prism-react-renderer": "1.1.0"
  },
```

You likely upgraded without knowing as it's a minor semver range

omry:
Thanks for digging!
I tried on a clean website just now (npx @docusaurus/init@next init my-website classic):

It seems to work with Docs or with docs.
Any idea why the difference?

website/sidebars.js
omry: > Note, it's actually upgrading the dependency from 1.1.0 to 1.1.1 that seems to break the highlighting, so I forced the former using:
> 
> ```
>   "resolutions": {
>     "prism-react-renderer": "1.1.0"
>   },
> ```
> 
> You likely upgraded without knowing as it's a minor semver range

Seems likely, thanks again. I will file an issues once I confirm this locally.

omry:
Ah, docs plugin is actually what I am using for versioning.

Related : in [this](https://v2.docusaurus.io/docs/versioning/), the npm command is not working.

website/sidebars.js
omry: Ah, saw that you already did. Thanks!

slorber:
> It seems to work with Docs or with docs.
> Any idea why the difference?

We want to support multiple instances of docs plugin at the same time in the future. This part is not the most robust or easy to understand with things hardcoded here and there. You should rather use `docs`. I can't tell you yet why but will know more soon :)

> Related : in this, the npm command is not working.

image

It seems to work for me (using "docs" at least)

omry:
Is it the same "Docs" from the top bar?
image

slorber:
This navbar is just a link with a label and a target, that you can configure in your site config:

image

omry:
Ah, cool.
While testing locally, I noticed this change:
image

In my case I would actually not want to discourage users that are browsing the next version. (I am in pre-release stage and I did not cut a version of the docs yet).
Do you know if this something I can play with in the docs plugin?

slorber:
image

You don't need to match the sidebar to the label of this link.


Actually the sidebar attribute name "docs" or "Docs" is more like an idea than a label displayed anywhere.

For example, take a look at Docusaurus 2 website, using 2 different sidebars:
- https://v2.docusaurus.io/docs/design-principles
- https://v2.docusaurus.io/docs/team/

omry:
> For example, take a look at Docusaurus 2 website, using 2 different sidebars:
> 
>     * https://v2.docusaurus.io/docs/design-principles
> 
>     * https://v2.docusaurus.io/docs/team/

Holy shit, I never thought you could have two different sidebars. this is awesome.

slorber:
Unfortunately, this banner is something we added recently and I think there's currently no way to disable it. If you would open an issue so we don't forget. (related https://github.com/facebook/docusaurus/pull/2916)

In a more general way, there's also a feature that has been quite requested to make "next" the "current version". 

Yes, 2 different sidebars, and soon we also plan to work on multiple instances (like for example ios sdk vs android sdk, when both don't have different versioning lifecycles)

omry:
> Unfortunately, this banner is something we added recently and I think there's currently no way to disable it. If you would open an issue so we don't forget. (related [facebook/docusaurus#2916](https://github.com/facebook/docusaurus/pull/2916))

https://github.com/facebook/docusaurus/issues/3013

> In a more general way, there's also a feature that has been quite requested to make "next" the "current version".

This has not been a big deal to me, when I want people to start using the next version in earnest I will cut a version.
I know that I can change the default version by ordering versions.json.
Maybe an easy solution for this request is to support it there:

```json
[
  "0.11", 
  "next"
]
```
VS:
```json
[
  "next",
  "0.11"
]
```

> Yes, 2 different sidebars, and soon we also plan to work on multiple instances (like for example ios sdk vs android sdk, when both don't have different versioning lifecycles).

very nice.

Issue #731, opened 2020-06-29
Clean up instantiate


omry:
instantiate for classes only
call for callables.
### instantiate
```
foo:
  type: ...
```
```python
object = instantiate(cfg.foo)
```

### call
```
foo:
  callable: ...
```
```python
ret = call(cfg.foo)
```

for backward compatibility, instantiate and call should still support both inputs but issue deprecation warning if the wrong input is specified.

shagunsodhani:
Based on the meeting today, the new goals are:

- `type`/`callable` key will not be introduced.
- `call` and `instantiate` would be aliases.
- Add support for return `None` when config is `None`.
- Improve the documentation:
  -  Split the documentation into two parts: (i) a short description and a longer, more technical version

omry:
Can you create a new task? It seems like it would be better than to repurpose this task.

shagunsodhani:
Closing in favor of https://github.com/facebookresearch/hydra/issues/764

PR #737, opened 2020-07-01
Differentiate between instantiate and call methods.

shagunsodhani:
## Motivation

Differentiate between `instantiate` and `call` methods.

Yes

## Test Plan

Added tests

## Related Issues and PRs

https://github.com/facebookresearch/hydra/issues/731

### DIFF
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -525,8 +525,8 @@ def _get_cls_name(config: Union[ObjectConf, DictConfig]) -> str:
     def _warn(field: str) -> None:
         if isinstance(config, DictConfig):
             warnings.warn(
-                f"""Config key '{config._get_full_key(field)}' is deprecated for `instantiate` since Hydra 1.0
-                 and will be removed in Hydra 1.1.
+                f"""Config key '{config._get_full_key(field)}' is deprecated for `instantiate` since Hydra 1.0 \
+and will be removed in Hydra 1.1.
 Use 'type' instead of '{field}'.""",
                 category=UserWarning,
             )
@@ -568,8 +568,8 @@ def _get_callable_name(config: Union[ObjectConf, DictConfig]) -> str:
     def _warn(field: str) -> None:
         if isinstance(config, DictConfig):
             warnings.warn(
-                f"""Config key '{config._get_full_key(field)}' is deprecated for `call` since Hydra 1.0
-                 and will be removed in Hydra 1.1.
+                f"""Config key '{config._get_full_key(field)}' is deprecated for `call` since Hydra 1.0 \
+and will be removed in Hydra 1.1.
 Use 'callable' instead of '{field}'.""",
                 category=UserWarning,
             )
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -584,4 +584,4 @@ Use '{key_name}' instead of '{field}'.""",
     elif _has_field(config=config, field="class"):
         return _get_name(field="class")
     else:
-        raise ValueError("Input config does not have a {key_name} field")
+        raise ValueError(f"Input config does not have a `{key_name}` field")
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -525,14 +525,15 @@ def _get_cls_name(config: Union[ObjectConf, DictConfig]) -> str:
     def _warn(field: str) -> None:
         if isinstance(config, DictConfig):
             warnings.warn(
-                f"""Config key '{config._get_full_key(field)}' is deprecated since Hydra 1.0 and will be removed in Hydra 1.1.
+                f"""Config key '{config._get_full_key(field)}' is deprecated for `instantiate` since Hydra 1.0
+                 and will be removed in Hydra 1.1.
 Use 'type' instead of '{field}'.""",
                 category=UserWarning,
             )
         else:
             warnings.warn(
                 f"""
-ObjectConf field '{field}' is deprecated since Hydra 1.0 and will be removed in Hydra 1.1.
+ObjectConf field '{field}' is deprecated for `instantiate` since Hydra 1.0 and will be removed in Hydra 1.1.
 Use 'type' instead of '{field}'.""",
                 category=UserWarning,
             )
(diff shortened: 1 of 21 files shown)


shagunsodhani:
@omry I could use some feedback on the PR. The current status is as follows:

1. Fails on lint tests.
2. Using `type` to denote the class will likely be a problem eventually (given its a Python keyword and used in Python files when using Structured Configs). 
3. Need to update the documentation. Will do that once point 2 is fixed.

omry:
I will review soon, hopefully tomorrow.
in the mean time, can you give a proper title to the PR?

hydra/_internal/utils.py
```diff
@@ -504,19 +504,36 @@ def _get_kwargs(config: Union[ObjectConf, DictConfig], **kwargs: Any) -> Any:
     return final_kwargs
 
 
+def _has_field(config: Union[ObjectConf, DictConfig], field: str) -> bool:
```
omry, line 507: It's hard to review when you mix your changes with refactoring like moving this function as is.
Try to separate cleanup to a different commit labeled as such to make the review easier.

hydra/_internal/utils.py
```diff
@@ -525,32 +542,60 @@ def _getcls(field: str) -> str:
...
         return _getcls(field="cls")
-    elif _has_field(field="class"):
+    elif _has_field(config=config, field="class"):
         return _getcls(field="class")
     else:
-        raise ValueError("Input config does not have a target field")
+        raise ValueError("Input config does not have a type field")
+
+
+def _get_callable_name(config: Union[ObjectConf, DictConfig]) -> str:
```
omry, line 564: this function is a copy of the other one with some minor changes. try to reduce this duplication by passing additional parameters. to one function.

hydra/_internal/utils.py
shagunsodhani: My bad. Will be more careful next time.

hydra/_internal/utils.py
shagunsodhani: Okay!

shagunsodhani:
> Nice.
> 
> I am thinking to add another mode that officially supports either type or callable as the same field.
> could still be target.
> 
> The advantage is that it simplifies the logic if the code does not know or care if it's a callable or a type.
> 
> we can make it explicit:
> `instantiate_or_call(cfg)`
> Where the it accepts without complains either type, callable or target.

Sounds good.

> Also, you didn't update the documentation and you didn't create a news fragment.

Yeah this is pending on alternative name of `type`. I am proposing to use `cls_name` or `cls_type` (or its modifications with `class`).

omry:
let's go with type.
we can always deprecate it if we really get into trouble.

omry:
Take a look at [this](https://github.com/anthonytec2/pytorch-lightning/pull/2/files/17450c54837b535757f5d89ba2ba63f116278a83..64714cbef1d8d6eb377c84d85ad207fa228b8582#diff-3598bc6c2f51dfaaf178b169ea9b5af4R33-R35).

Any thoughts about making instantiate(None) to return None to simplify that?

omry:
cc @anthonytec2

shagunsodhani:
> this

Yes! This sounds quite useful. Should I include it in the next PR ?

omry:
or this one. you still need to make some changes here I think.

shagunsodhani:
> or this one. you still need to make some changes here I think.

Right my bad :)

shagunsodhani:
@omry I want to confirm if I should be updating all the plugins to use `type` or `callable` in place of `target`? Using `target` is supported.

Todo:

1. Take another pass with the documentation and examples (might have missed some)
2. Add support for handling null value
3. Update documentation and add news fragment.

hydra/_internal/utils.py
```diff
@@ -522,86 +522,67 @@ def _has_field(config: Union[ObjectConf, DictConfig], field: str) -> bool:
...
+    """
+
+    if should_get_cls:
+        fn_name = "instantiate"
+        key_name = "type"
+        key_name_to_depriciate = "callable"
+    else:
+        fn_name = "call"
+        key_name = "callable"
+        key_name_to_depriciate = "type"
```
omry, line 547: deprecate

hydra/_internal/utils.py
```diff
@@ -522,86 +522,67 @@ def _has_field(config: Union[ObjectConf, DictConfig], field: str) -> bool:
...
-        return _getcls(field="class")
-    else:
-        raise ValueError("Input config does not have a type field")
 
+def _get_cls_or_callable_name(
+    config: Union[ObjectConf, DictConfig], should_get_cls: bool
+) -> str:
+    """Since the logic for getting class or callable name is very similar,
+    we use this wrapper function to avoid code-duplication. if `should_get_cls` is set to True,
+    the function returns the name of class, else it returns the name of callable.
```
omry, line 537: no need for the comment to justify the function.
just describe what it does.

hydra/types.py
```diff
@@ -14,6 +14,8 @@
 class ObjectConf(Dict[str, Any]):
     # class, class method or function name
     target: str = MISSING
+    callable: str = MISSING
+    type: str = MISSING
```
omry, line 18: I don't like this.
let's chat tomorrow.

website/docs/patterns/objects.md
```diff
@@ -4,19 +4,30 @@ title: Creating objects and calling functions
...
     :param config: An ObjectConf or DictConfig describing what to call and what params to use
     :param args: optional positional parameters pass-through
     :param kwargs: optional named parameters pass-through
-    :return: the return value from the specified class or method
+    :return: the return value from the specified method
     """
 ```
 
-### ObjectConf definition
+### ObjectConf definition for hydra.utils.instantiate
```
omry, line 30: ObjectConf here does not line up with the abomination I don't like above.

shagunsodhani:
> Take a look at [this](https://github.com/anthonytec2/pytorch-lightning/pull/2/files/17450c54837b535757f5d89ba2ba63f116278a83..64714cbef1d8d6eb377c84d85ad207fa228b8582#diff-3598bc6c2f51dfaaf178b169ea9b5af4R33-R35).
> 
> Any thoughts about making instantiate(None) to return None to simplify that?
> 
> `config.yaml`:
> 
> ```yaml
> foo: null
> ```
> 
> ```python
> cfg = OmegaConf.create("config.yaml")
> assert instantiate(cfg.foo) == None
> ```
> 
> `Structured Configs`:
> 
> ```python
> @dataclass
> class Config:
>   foo: Optional[ObjectConf] = None
> 
> cfg = OmegaConf.structured(Config)
> assert instantiate(cfg.foo) == None
> ```

https://github.com/facebookresearch/hydra/pull/737/commits/fea610e6807d7a62aa2b0c5c0ed2cfb4fd61a5eb adds support for this - for both `call` and `instantiate`. I do not feel strongly about supporting both `call` and `instantiate` and we may decided to support just `instantiate`.

shagunsodhani:
Closing in light of recent discussions (https://github.com/facebookresearch/hydra/issues/731#issuecomment-656426570)

PR #518, opened 2020-04-08
Ray launcher WIP 

jieru-hu:
Update 10/14
Follow up items:
- [x] add custom resolver to get lib versions.
- [x] clean up orphaned instances. 

Update 09/30

A few items I want to follow up separately (will create PR for these)
1. cannot run the plugin in python 3.6 (cloudpickle issue, same error message as #428, failed run here https://app.circleci.com/pipelines/github/jieru-hu/hydra/611/workflows/0d99e3b0-1442-446f-857b-f476a7707b6d/jobs/6648 ), I tried a few things (updating cloudpickle version etc but was not able to resolve it.)
2. nightly builds test AMIs (right now the AMI is set as an env variable which is annoying everytime when we need to update the AMI id, I want to automate this.)
3. doc update. The doc needs to be refreshed a bit. I think it might be easier to create a separate PR for this.

Update 09/28

Summary of the changes:
#### 2. Changes to integration test:
The goal is "No outbound traffic for the test instances." The barrier is the `pip install` and `conda create` we need to run while setting up the instance which requires us to open 443 to all outbound traffic. 
To get around this: for `conda` and dependency packages needed for starting the cluster, I created a base AMI that has everything pre-installed. 2) for Hydra related packages, build the wheels at test time and install the wheels on the instance.
The upside is we achieve "no outbound traffic for the instance", the downside is that means we need update AMI when dependencies changes. To help with that I created a script 
(`create_ami.py`) to automate building the AMI.

It would be good to build nightly AMIs and wheels, that's something I want to work on soon. 

<details><summary>output from running `create_ami.py`</summary>

```
$ AWS_PROFILE=jieru python create_ami.py 
2020-09-28 16:23:56.268051 - Running: aws ec2 authorize-security-group-egress --group-id sg-0a1 --ip-permissions IpProtocol=tcp,FromPort=443,ToPort=443,IpRanges=[{CidrIp=0.0.0.0/0}]
2020-09-28 16:23:57.464861 - 
2020-09-28 16:23:57.487688 - Running: ray up /var/folders/n_/9qzct77j68j6n9lh0lw3vjqcn96zxl/T/tmpjw6lihef.yaml -y
2020-09-28 16:25:57.400029 - 2020-09-28 16:23:58,540    INFO cli_logger.py:388 -- Using cached config at /var/folders/n_/9qzct77j68j6n9lh0lw3vjqcn96zxl/T/ray-config-d951a214f8602b878335411b5df6e84af463922b

[...]

PR #739, opened 2020-07-02
Changes to integration tests for support launching to remote machines

jieru-hu:
## Motivation

This is the integration tests change part from https://github.com/facebookresearch/hydra/pull/518. Per our offline discussion, creating a new PR here so its easier to review the changes out of the context of ray launcher.

Yes

## Test Plan

## Related Issues and PRs


omry:
Please describe what this is actually doing in the PR description.

jieru-hu:
> Please describe what this is actually doing and why it is useful to achieve the task of launching to remove machines in the PR description.

Sorry my bad. I've added more context to the PR description! :) @omry

omry:
>  Add sweep dir override & clean up
Before the change the tests will override the sweep_dir with a tmp dir. However this won’t work when the remote machines are different OS from the machines where the test is launched. Add sweep_dir override to make it a relative dir rather than a remote dir helps solve this.

Can you override the sweep dir in the overrides instead?
"hydra.sweep.dir=multirun"

> Make task_function return function instead. This allows the subclass to make task_function as None. This helps solve cloudpicking issues with RAY and also make the task function more flexible.

Can you give more details about the cloudpickling issue?

> In IntegrationTestSuite
>    Add launcher specific overrides. This allows launcher specific overrides which can handy when certain overrides doesn’t apply to all other launchers.

Individual launchers are already able to pass additional overrides. From the example launcher - extra_flags below.
Note that this makes the entire task_launcher_cfg parameter redundant and I intend to clean it up.
In the example below the extra flags is ["-m"].

Try making it: ["-m", "hydra.sweep.dir=."]
```
@pytest.mark.parametrize(
    "task_launcher_cfg, extra_flags",
    [
        (
            {
                "defaults": [
                    {"hydra/launcher": None},
                    {"hydra/hydra_logging": "hydra_debug"},
                    {"hydra/job_logging": "disabled"},
                ],
                "hydra": {
                    "launcher": {
                        "target": "hydra_plugins.example_launcher_plugin.example_launcher.ExampleLauncher",
                        "params": {"foo": 10, "bar": "abcde"},
                    }
                },
            },
            ["-m"],
        )
    ],
)
class TestExampleLauncherIntegration(IntegrationTestSuite):
    """
    Run this launcher through the integration test suite.
    """
    pass
```


>    Add pre_output_validation : a function that is highly customizable for the launcher specific classes to process the output before validation. In a remote launcher’s case, this method is overriden in subclasses to down the output file to local dir for validation to proceed.

That's not a good name. it does not validate anything.
How about `post_run_callback()` ?

>    Add orig_working_dir : this was added due to the same reason as adding sweep_dir, setting working dir to be tmp dir won’t play well when test machine and remote machines are on different OS

With what I said above, reconsider this change for the same reasons :)

jieru-hu:
Thx for the review @omry ! I've made some progress on addressing your comments. Let me know what you think :)

As for the pickling issue, if in task_function we do not return a function but instead just do ```return 100``` I got this:

```
E                       ray.exceptions.RayTaskError(ModuleNotFoundError): ray::hydra_plugins.hydra_ray_launcher._launcher_util._run_job() (pid=23990, ip=10.30.99.17)
E                         File "python/ray/_raylet.pyx", line 427, in ray._raylet.execute_task
E                         File "python/ray/_raylet.pyx", line 444, in ray._raylet.execute_task
E                         File "/Users/jieru/opt/anaconda3/envs/pytest38/lib/python3.8/site-packages/ray/serialization.py", line 317, in deserialize_objects
E                           self._deserialize_object(data, metadata, object_id))
E                         File "/Users/jieru/opt/anaconda3/envs/pytest38/lib/python3.8/site-packages/ray/serialization.py", line 257, in _deserialize_object
E                           return self._deserialize_msgpack_data(data, metadata)
E                         File "/Users/jieru/opt/anaconda3/envs/pytest38/lib/python3.8/site-packages/ray/serialization.py", line 238, in _deserialize_msgpack_data
E                           python_objects = self._deserialize_pickle5_data(pickle5_data)
E                         File "/Users/jieru/opt/anaconda3/envs/pytest38/lib/python3.8/site-packages/ray/serialization.py", line 226, in _deserialize_pickle5_data
E                           obj = pickle.loads(in_band)
E                       ModuleNotFoundError: No module named 'hydra_ray_launcher'

/Users/jieru/opt/anaconda3/envs/pytest38/lib/python3.8/site-packages/ray/worker.py:1515: RayTaskError(ModuleNotFoundError)

```

hydra/extra/pytest_plugin.py
```diff
@@ -51,6 +51,11 @@ def _(
         sweep.config_name = config_name
         sweep.strict = strict
         sweep.overrides = overrides or []
+        sweep_dir_overrides = [r for r in sweep.overrides if "hydra.sweep.dir=" in r]
```
omry, line 54: I think it's best if you compose the config and get the run dir fro mhydra.sweep.dir in the resulting config.
this kind of logic is fragile and hacky.

hydra/test_utils/launcher_common_tests.py
```diff
@@ -288,9 +289,10 @@ def sweep_2_jobs(
     with sweep:
         assert sweep.temp_dir is not None
         assert sweep.returns is not None
-        temp_dir = Path(sweep.temp_dir)
 
-        multirun_cfg_path = Path(temp_dir) / "multirun.yaml"
+        config_dir = Path(sweep.temp_dir)
```
omry, line 293: what is this fixing?

hydra/test_utils/launcher_common_tests.py
```diff
@@ -346,6 +348,40 @@ def sweep_two_config_groups(
 
 @pytest.mark.usefixtures("hydra_restore_singletons")
 class IntegrationTestSuite:
+    working_dir_key = "tests_working_dir"
+    output_dir_key = "tests_output_dir"
+
+    def filter_overrides(
```
omry, line 354: You are adding a whole lot of new logic here and I have no idea what it's supposed to do.
1. What overrides is this filtering?
2. What does this achieve?

hydra/test_utils/launcher_common_tests.py
```diff
@@ -388,13 +424,16 @@ def test_custom_task_name(
...
+        output_tmp_dir, overrides = self.filter_tmp_dir(overrides)
         integration_test(
             tmpdir=tmpdir,
             task_config=cfg,
             overrides=overrides,
             prints="HydraConfig.get().job.name",
             expected_outputs=expected_name,
             filename=filename,
+            output_tmp_dir=output_tmp_dir,
+            post_run_callback=self.post_run_callback(),
```
omry, line 436: this is calling the function, you probably want to pass it instead.

hydra/test_utils/launcher_common_tests.py
```diff
@@ -388,13 +424,16 @@ def test_custom_task_name(
         task_config = OmegaConf.create(task_config or {})  # type: ignore
         cfg = OmegaConf.merge(task_launcher_cfg, task_config)
         assert isinstance(cfg, DictConfig)
+        output_tmp_dir, overrides = self.filter_tmp_dir(overrides)
```
omry, line 427: You are going through a lot of trouble to get a output temp dir, but the function is taking tmpdir as a parameter.
what is the difference?

hydra/test_utils/test_utils.py
```diff
@@ -129,10 +133,15 @@ def __call__(self, cfg: DictConfig) -> Any:
...
-        self.temp_dir = tempfile.mkdtemp()
+
         overrides = copy.deepcopy(self.overrides)
         assert overrides is not None
-        overrides.append(f"hydra.sweep.dir={self.temp_dir}")
+        if self.sweep_dir:
+            self.temp_dir = self.sweep_dir
+        else:
+            self.temp_dir = tempfile.mkdtemp()
+            overrides.append(f"hydra.sweep.dir={self.temp_dir}")
```
omry, line 143: There has to be a better way to do what you want.

Explain the problem you are solving again?

hydra/test_utils/test_utils.py
```diff
@@ -306,7 +328,10 @@ def experiment(cfg):
             modified_env.update(env_override)
         subprocess.check_call(cmd, env=modified_env)
 
-        with open(output_file, "r") as f:
+        if post_run_callback:
+            post_run_callback(**locals())
```
omry, line 332: make the callback well defined and accept specific things. passing **locals()  is fragile and unclear.

hydra/test_utils/test_utils.py
jieru-hu: ```temp_dir``` is local to the  OS of the machine running the tests, a temp dir on MAC OS looks like ```/var/folders/n_/9qzct77j68j6n9lh0lw3vjqcn96zxl/T/tmpz_w5rxac```, when running on remote ubuntu AWS instances, Hydra will try to create a dir looks like this, which won't work. So I want to manually pass in a ```sweep_dir``` in this case.

hydra/test_utils/launcher_common_tests.py
jieru-hu: nothing 😂 a legacy change which i need to revert.

hydra/test_utils/launcher_common_tests.py
jieru-hu: Same issue with the tmp dir is OS dependent. A temp dir that works on MAC OS doesnt work on Ubuntu :(

hydra/test_utils/launcher_common_tests.py
jieru-hu: I want to override the tests working dir and tests output dir somehow. 

Right now I'm passing the override from here https://github.com/facebookresearch/hydra/pull/518/files#diff-b44e4fd91c6f6ff3f7dc895d9a6a1672R121-R122

So in the tests I want to get the dir overrides value but also remove them from the application overrides. Not sure if there's a better way to do this (previous I was doing method overriding). Any suggestion?

hydra/extra/pytest_plugin.py
jieru-hu: Thanks! Didn't think of that. will address this once we agree this is the best way to approach the issue.

hydra/test_utils/test_utils.py
omry: okay.
should we push down the creation of the working directory for the job itself? It should be creating it if it's not there anyway.

If the actual directory is not there it will try to create it. we just need to choose a directory that can work remotely. a safe bet is `${env:HOME}/tmp/${now:...}`.

I will take a closer look tomorrow, but I think the integration test is taking a tempdir from the caller. can you caller override that tmpdir through the overrides somehow instead of opening a whole new channel through everything to pass it?

hydra/test_utils/launcher_common_tests.py
omry: Hydra provices the ability to override EVERYTHING in the configuration through the command line overrides.
you can pass an override like : hydra.run.dir=/foo/bar and that will be come the run dir for the job, no?

hydra/extra/pytest_plugin.py
jieru-hu: forgot to address this one, will fix this in a follow up commit

hydra/test_utils/launcher_common_tests.py
jieru-hu: Thx. I've taken another pass and try to override everything instead for the Integration tests.

hydra/test_utils/launcher_common_tests.py
jieru-hu: ```self.post_run_callback()``` returns another function here, so calling it here

hydra/test_utils/test_utils.py
jieru-hu: true, i will make this one better.

hydra/test_utils/launcher_common_tests.py
```diff
@@ -346,6 +349,15 @@ def sweep_two_config_groups(
...
 class IntegrationTestSuite:
+
+    def post_run_callback(
+        self, **kwargs: Any
+    ) -> Optional[Callable[[Dict[Any, Any]], None]]:
+        """
+        This will be called after running test app and before validating output file from test.
+        """
+        return None
+
```
omry, line 360: I would like this function to have a clear signature. this is an API we expect other plugins to use.

hydra/test_utils/launcher_common_tests.py
```diff
@@ -489,22 +502,25 @@ def test_custom_sweeper_run_workdir(
         expected_dir: str,
         task_launcher_cfg: DictConfig,
         extra_flags: List[str],
+        expected_working_dir: Optional[Path]
```
omry, line 505: Please explain what expected_working_dir is any why it's needed.
I see it's local to this test which is good, why was it not required before?

hydra/extra/pytest_plugin.py
omry: roger.

hydra/test_utils/launcher_common_tests.py
```diff
@@ -346,6 +349,15 @@ def sweep_two_config_groups(
 
 @pytest.mark.usefixtures("hydra_restore_singletons")
 class IntegrationTestSuite:
+
+    def post_run_callback(
+        self, **kwargs: Any
+    ) -> Optional[Callable[[Dict[Any, Any]], None]]:
+        """
+        This will be called after running test app and before validating output file from test.
```
omry, line 357: You can give some context in the comment:
"Typical use case is to download remote working directory locally to allow for validation of remotely launched jobs".

hydra/test_utils/launcher_common_tests.py
omry: Please do.

hydra/test_utils/launcher_common_tests.py
```diff
@@ -538,12 +555,16 @@ def test_to_absolute_path_multirun(
             overrides=overrides,
             prints="hydra.utils.to_absolute_path('/foo/bar')",
             expected_outputs=path,
+            post_run_callback=self.post_run_callback(),
         )
-        outputs = [str(tmpdir / "foo/bar"), str(tmpdir / expected_dir)]
+        working_dir = expected_working_dir if expected_working_dir else tmpdir
```
omry, line 560: you have an allocated LOCAL temp directory. can you download your files into that instead of juggling between the two output directories?

omry:
I am not seeing any test that is using the post_run_callback.
add it to be sure it's actually going to be useful for your use case in the Ray plugin.

hydra/test_utils/launcher_common_tests.py
jieru-hu: I've removed params and replaced it with a method instead. (I've realized that adding a params impacts all the existing tests and no easy way to get around that in pytest. )

Also I've renamed this to ```test_app_working_dir```. The test app working dir is the same on local machines as the tests working dir, but is different when tests kick off application on remote machines. which is why this was added.

hydra/test_utils/launcher_common_tests.py
jieru-hu: I think  the naming of the variable is confusing. I've renamed it as previously mentioned. Pls let me know if it still doesnt make sense.

hydra/test_utils/launcher_common_tests.py
omry: I am having a hard time following the last sentence as it talks about the history if this PR, which is not really relevant right now.

What is this new test testing?

hydra/test_utils/launcher_common_tests.py
```diff
@@ -489,22 +502,25 @@ def test_custom_sweeper_run_workdir(
...
             tmpdir=tmpdir,
             task_config=cfg,
             overrides=overrides,
             prints="os.getcwd()",
-            expected_outputs=str(tmpdir / expected_dir),
+            expected_outputs=expected_outputs,
+            post_run_callback=self.post_run_callback(),
         )
 
     def test_get_orig_dir_multirun(
```
omry, line 522: It looks like you are changing the semantics of this test (it's testing the output working directory, and you are providing different working directory as expected output).

Please explain why.

hydra/test_utils/launcher_common_tests.py
omry: This test is testing that the hydra.utils.to_absolute_path() is working correctly.
It's intended to be executed and tested in the job.

Is this change changing the expected output to that of the remote run?

hydra/test_utils/launcher_common_tests.py
jieru-hu: sorry for the confusion. The PR is updated now. I didnt rename the test, but just renames the variable's name to ```test_app_working_dir```

hydra/test_utils/launcher_common_tests.py
jieru-hu: added.

hydra/extra/pytest_plugin.py
jieru-hu: I didn't end using the ```compose``` API but used the utils from hydra internal to get the overrides value.

```compose``` API doesn't apply to the use case here since here I want to know if the tests has overriden ```sweep.dir```, if not, create a ```tmp_dir```

File: hydra/test_utils/launcher_common_tests.py (hunk 4 above)
Comment (jieru-hu) [reply hydra/test_utils/launcher_common_tests.py]: fixed this.

File: hydra/test_utils/launcher_common_tests.py (hunk 8 above)
Comment (jieru-hu) [reply hydra/test_utils/launcher_common_tests.py]: added signature.

File: hydra/test_utils/launcher_common_tests.py (hunk 11 above)
Comment (jieru-hu) [reply hydra/test_utils/launcher_common_tests.py]: yes, changing this to the remote working dir of the test application.

File: hydra/test_utils/test_utils.py (hunk 6 above)
Comment (jieru-hu) [reply hydra/test_utils/test_utils.py:152]: I thought I responded to this one, but somehow it got lost 🤔

There's 2 test suites:  ```LauncherTestSuite``` and the ```IntegrationTestSuite```. I've fixed the ```IntegrationTestSuite``` with the tmpdir. Pushing the dir creation to the the jobs is a great ideas. I will make the changes now.

File: hydra/extra/pytest_plugin.py [hunk 13]
```diff
@@ -42,6 +44,7 @@ def _(
         config_name: Optional[str],
         overrides: Optional[List[str]],
         strict: Optional[bool] = None,
+        temp_dir: Optional[Path] = None,
```
omry, line 47: I am tempted to start [requiring named parameters](https://stackoverflow.com/questions/2965271/forced-naming-of-parameters-in-python
) in places like this.

It's a breaking change initially but it will make it much easier to remove things later, for example strict will go away here in the next major version.

hydra/test_utils/launcher_common_tests.py
```diff
@@ -346,6 +391,20 @@ def sweep_two_config_groups(
 
 @pytest.mark.usefixtures("hydra_restore_singletons")
 class IntegrationTestSuite:
+    def post_run_callback(self, tmp_dir: str) -> None:
```
omry, line 394: Since this is happening after the run, we should have the JobReturn. it seems much better to pass that than "tmp_dir".
This is what JobReturn contains:
```python
@dataclass
class JobReturn:
    overrides: Optional[Sequence[str]] = None
    return_value: Any = None
    cfg: Optional[DictConfig] = None
    hydra_cfg: Optional[DictConfig] = None
    working_dir: Optional[str] = None
    task_name: Optional[str] = None
```

hydra/test_utils/launcher_common_tests.py
```diff
@@ -346,6 +391,20 @@ def sweep_two_config_groups(
...
+        download remote working directory locally to allow for validation of remotely launched jobs.
+        """
+        return None
+
+    def get_test_app_working_dir(self) -> Optional[Path]:
+        """
+        By default test applications working dir is tmpdir, override this method if that's not the case.
+        This could be helpful when the tests kick off applications on remote machines.
+        """
+        return None
```
omry, line 406: Still feels a bit leaky (as in leaky abstraction), but this is much better.

hydra/test_utils/launcher_common_tests.py
```diff
@@ -495,16 +556,23 @@ def test_custom_sweeper_run_workdir(
         task_config = OmegaConf.create(task_config or {})  # type: ignore
         cfg = OmegaConf.merge(task_launcher_cfg, task_config)
         assert isinstance(cfg, DictConfig)
+        expected_outputs = (
+            str(self.get_test_app_working_dir() / expected_dir)  # type: ignore
```
omry, line 560: why the type ignore?

hydra/test_utils/launcher_common_tests.py
```diff
@@ -538,12 +607,23 @@ def test_to_absolute_path_multirun(
             overrides=overrides,
             prints="hydra.utils.to_absolute_path('/foo/bar')",
             expected_outputs=path,
+            post_run_callback=self.post_run_callback,
         )
-        outputs = [str(tmpdir / "foo/bar"), str(tmpdir / expected_dir)]
+        working_dir = (
+            self.get_test_app_working_dir()
+            if self.get_test_app_working_dir()
```
omry, line 614: is get_app_working_dir() specific to multirun?

hydra/test_utils/test_utils.py
```diff
@@ -283,19 +295,28 @@ def experiment(cfg):
             f.write("# @package _global_\n")
             OmegaConf.save(task_config, f)
         config_name = "config_name='config'"
-    output_file = str(tmpdir / "output.txt")
+
+    script_output_file = str(tmpdir / "output.txt")
+
+    # You can have our test script write to an arbitrary file
+    # if so, you are in charge of copying the arbitrary file to the validation dir.
+    validation_file = str(tmpdir / "output.txt")
```
omry, line 303: this is not really arbitrary. it's output.txt in the job output directory. this is 100% controlled by this function.

hydra/test_utils/launcher_common_tests.py
jieru-hu: nice, yea, I wasn't sure what's best to put in the signature, this looks much better.

hydra/test_utils/test_utils.py
jieru-hu: removed, this variable was not needed anymore.

hydra/test_utils/launcher_common_tests.py
jieru-hu: mypy complains since ```self.get_test_app_working_dir()``` returns an ```Optional[Path]```, but we check in the if else so it should be fine.


```
hydra/test_utils/launcher_common_tests.py:553: error: Unsupported left operand type for / ("None")
hydra/test_utils/launcher_common_tests.py:553: note: Left operand is of type "Optional[Path]"

```

hydra/test_utils/launcher_common_tests.py
jieru-hu: I think we've resolved this one. Pls let me know otherwise :)

hydra/test_utils/launcher_common_tests.py
jieru-hu: No,  I don't think so. But launcher always run in multi run mode, right?

hydra/test_utils/launcher_common_tests.py
jieru-hu: Ok, I realized that I could just download the output from as part of running aws launcher so I've removed this completely.

This was not possible a few iterations ago because of reasons that I no longer recall 🙃 but refactoring helped make it possible.

hydra/test_utils/launcher_common_tests.py
omry: okay, is this actually hydra.run.dir in the config?

hydra/test_utils/launcher_common_tests.py
omry: mypy is smart enough to know if the value cannot be None.
in this case you are calling a function which may return something else the second time.
put it into a temp variable and mypy will know it can't be None.

hydra/test_utils/launcher_common_tests.py
omry: great.

hydra/test_utils/launcher_common_tests.py
jieru-hu: no. ```hydra.runtime.cwd``` here is the tests' working dir (local machine dir, in this case the ```temp_dir```), but ```get_test_app_working_dir ``` is where the test app is actually kicked off, this is different on a remote machine. (in Ray's case, this value would be "/home/ubuntu" but in the local machine case this would be "/tmp/hydra_test/" or something like this.  )

hydra/test_utils/launcher_common_tests.py
jieru-hu: thx! fixed.

jieru-hu:
squashed all the commits and rebased on master again.

omry:
The eagle has landed!

PR #741, opened 2020-07-06
[RQ plugin] Compose sweep config in launch function

jan-matthis:
## Motivation

This PR moves composition of the sweep config from `execute_job` into `launch`: `launch` is executed on the submitting node while `execute_job` will be forked processes on the worker. This way, config files only need to be accessibly by the submitting node which will handle the composition.

Yes

### DIFF
--- a/plugins/hydra_rq_launcher/setup.py
+++ b/plugins/hydra_rq_launcher/setup.py
@@ -6,7 +6,7 @@ with open("README.md", "r") as fh:
     LONG_DESC = fh.read()
     setup(
         name="hydra-rq-launcher",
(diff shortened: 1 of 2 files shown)


plugins/hydra_rq_launcher/setup.py
```diff
@@ -6,7 +6,7 @@
     LONG_DESC = fh.read()
     setup(
         name="hydra-rq-launcher",
-        version="1.0.0rc1",
+        version="1.0.0rc2",
```
omry, line 9: As a rule, I try to always have the repo in the next version. (in other words, it should have already been at rc2).

PR #744, opened 2020-07-07
[RQ launcher] Clear environment variable cache

jan-matthis:
# Motivation

This PR adds a condition to clear environment variables from the cache of the sweep config. If, say, `${env:HOME}` was part of `hydra.run.sweep.dir`, it would be set to the `$HOME` dir of the headnode otherwise, and resolution would fail on the worker.

Yes

## Related Issues and PRs

https://github.com/facebookresearch/hydra/pull/741 changed the RQ launcher plugin such that it composes the sweep configuration in the launch function (e.g., on a headnode) rather than inside a worker process (e.g., remote).

### DIFF
--- a/plugins/hydra_rq_launcher/hydra_plugins/hydra_rq_launcher/_core.py
+++ b/plugins/hydra_rq_launcher/hydra_plugins/hydra_rq_launcher/_core.py
@@ -113,6 +113,12 @@ def launch(
             sweep_config.hydra.job.id = enqueue_keywords["job_id"]
             sweep_config.hydra.job.num = initial_job_idx + idx
 
+        # Remove environment variables from cache, e.g., used when resolving sweep_dir
+        config_cache = OmegaConf.get_cache(sweep_config)
+        if "env" in config_cache:
+            del config_cache["env"]
+            OmegaConf.set_cache(sweep_config, config_cache)
+
         job = queue.enqueue(
             execute_job,
             sweep_config=sweep_config,


omry:
Please describe the actual use case you had that triggered the issue.
This could be a fundamental problem effecting other plugins as well.

jan-matthis:
> Please describe the actual use case you had that triggered the issue.

Say you are launching `hydra_app hydra.sweep.dir='${env:HOME}/project_dir' hydra/launcher=rq -m`. The launcher will create a sweep dir in the main process:
```
sweep_dir = Path(str(launcher.config.hydra.sweep.dir))
sweep_dir.mkdir(parents=True, exist_ok=True)
```

Following this, `OmegaConf.get_cache(launcher.config)` will contain an entry for `env.HOME` (this it had to be resolved to create the dir). When the sweep config is composed, the cache of the master config (`launcher.config`) is copied: https://github.com/facebookresearch/hydra/blob/master/hydra/_internal/config_loader_impl.py#L326, which will include the `$HOME` environment variable of the submitting/host node. 

This PR circumvents the issue by clearing env from the cache. Thinking about it, other plugins that do remote invocation can equally be affected by this. An alternative would be to e.g. clear `env` in `config_loader_impl.py`.

omry:
Exactly. other plugins would be effected too.
We should probably only copy specific things in the cache (such as now function) instead in the place where you pointed out.

Please send another PR, with proper tests for this.

PR #745, opened 2020-07-07
[RQ launcher] Change defaults for timeouts

jan-matthis:
## Motivation

This PR changes the default configuration of the RQ launcher plugin regarding timeouts: Python RQ has four different timeout values, referred to job_timeout (maximum runtime of a job), ttl (maximum wait time in queue), result_ttl (how long results are kept), and failure_ttl (how long failed jobs are kept). 

Having low timeout values might lead to unexpected behaviour, e.g. jobs being interrupted unexpectedly, or results disappearing. RQ only allows setting timeouts to `None` for a subset of timeout values. This PR changes the default config so that all timeout values have a very high (100 days) default applied. 

Yes

### DIFF
--- a/plugins/hydra_rq_launcher/hydra_plugins/hydra_rq_launcher/_core.py
+++ b/plugins/hydra_rq_launcher/hydra_plugins/hydra_rq_launcher/_core.py
@@ -101,6 +101,12 @@ def launch(
         description = " ".join(filter_overrides(overrides))
 
         enqueue_keywords = OmegaConf.to_container(rq_cfg.enqueue, resolve=True)
+        if enqueue_keywords["job_timeout"] is None:
+            enqueue_keywords["job_timeout"] = -1
+        if enqueue_keywords["result_ttl"] is None:
+            enqueue_keywords["result_ttl"] = -1
+        if enqueue_keywords["failure_ttl"] is None:
+            enqueue_keywords["failure_ttl"] = -1
         if enqueue_keywords["job_id"] is None:
             enqueue_keywords["job_id"] = str(uuid.uuid4())
         if enqueue_keywords["description"] is None:
--- a/plugins/hydra_rq_launcher/hydra_plugins/hydra_rq_launcher/config.py
+++ b/plugins/hydra_rq_launcher/hydra_plugins/hydra_rq_launcher/config.py
@@ -23,14 +23,14 @@ class RedisConf:
 
 @dataclass
 class EnqueueConf:
-    # maximum runtime of the job before it's interrupted and marked as failed (in sec)
-    job_timeout: Optional[int] = None
-    # maximum queued time before the job before is discarded (in sec)
-    ttl: Optional[int] = None
-    # how long successful jobs and their results are kept (in sec), default: 10 days
-    result_ttl: int = 864000
-    # specifies how long failed jobs are kept (in sec), default: 100 days
-    failure_ttl: int = 8640000
+    # maximum runtime of the job before it's killed (e.g. "1d" for 1 day, units: d/h/m/s), default: no limit
+    job_timeout: Optional[str] = None
+    # maximum queued time before the job before is discarded (e.g. "1d" for 1 day, units: d/h/m/s), default: no limit
+    ttl: Optional[str] = None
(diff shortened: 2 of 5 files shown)


plugins/hydra_rq_launcher/hydra_plugins/hydra_rq_launcher/config.py
```diff
@@ -23,12 +23,12 @@ class RedisConf:
...
 @dataclass
 class EnqueueConf:
-    # maximum runtime of the job before it's interrupted and marked as failed (in sec)
-    job_timeout: Optional[int] = None
-    # maximum queued time before the job before is discarded (in sec)
-    ttl: Optional[int] = None
-    # how long successful jobs and their results are kept (in sec), default: 10 days
-    result_ttl: int = 864000
+    # maximum runtime of the job before it's interrupted (in sec), default: 100 days
+    job_timeout: int = 8640000
```
omry, line 27: By interrupted you mean "murdered with a sharp signal"?

100 days seems large enough to indicate that you are actually just trying to disable it. Is there no way to disable it in Redis?

jan-matthis: Exactly, the process gets killed and the job moves to the failed queue. 

No, it's not possible to disable it in Python RQ -- I was tricked by this as well. While it's possible to pass `job_timeout=None`, this will internally default to a `180` second timeout. I think it's much more preferable to default to not enforcing time limits which this PR effectively does

omry: You think 100 days as infinity, but it's not.
I actually had ML jobs that took over 3 months to complete.

Can you file an issue against Redis queue to allow them to accept no-timeout?

jan-matthis: Found a way around, `-1` should work for inf `job_timeout`. Does that sound good?

omry: Much better.
Be sure to add a comment in the dataclass.

jan-matthis: Looking into it, I figured that instead of passing integers, strings with units can be passed, e.g., `24h`. I think this is much easier to use and understand than using integers with special meanings (such as -1 for infinite, and expressing all timeouts by converting to seconds). 

I have updated the code in the latest commit to use 1000 days (`1000d`) as default timeout for all four values.

omry: I still prefer -1 for infinite or another constant if they support it.
if they take a string or an int, use Any as the type and do validation that i's a string or an int in the code itself. (as of now OmegaConf does not support [Union[int,str]](url)).

jan-matthis: In the latest commit, types are `Optional[str]` defaulting to `None`. Internally, `None` is converted such that it means no limit (the internal conversion necessary to archive that differs between the values, but it's just internals this way that a user doesn't need to worry about).

jan-matthis: In addition, I updated the example to show how a default timeout value can be overridden

plugins/hydra_rq_launcher/hydra_plugins/hydra_rq_launcher/_core.py
```diff
@@ -101,6 +101,12 @@ def launch(
         description = " ".join(filter_overrides(overrides))
 
         enqueue_keywords = OmegaConf.to_container(rq_cfg.enqueue, resolve=True)
+        if enqueue_keywords["job_timeout"] is None:
+            enqueue_keywords["job_timeout"] = -1
+        if enqueue_keywords["result_ttl"] is None:
+            enqueue_keywords["result_ttl"] = -1
+        if enqueue_keywords["failure_ttl"] is None:
+            enqueue_keywords["failure_ttl"] = -1
         if enqueue_keywords["job_id"] is None:
```
omry, line 110: That's a bit clumsy, but I get why you are doing it.

PR #748, opened 2020-07-08
Introducing a DSL for the command line override syntax

### LINKED ISSUE #573: Multirun: allow commas in config values set from command-line
Passing `--multirun foo=bar,baz` will result in a sweep over the two specified values for `key`. However, the comma might be used as a part of an actual string value. A simple fix would be to allow introduce an escape character for value splitting, so that e.g. `--multirun foo=bar/,baz` results in a single "bar,baz" value for "foo".

### LINKED ISSUE #738: Strict definition for command line values and sweeping
[Document](https://docs.google.com/document/d/1OELdhAe8jWfNLioqjfRaMcmh56vnzIwUNA0Id9aRIOQ/edit)

omry:
## Docs
[Hydra Override Syntax](https://docs.google.com/document/d/1OELdhAe8jWfNLioqjfRaMcmh56vnzIwUNA0Id9aRIOQ/edit#heading=h.tw9op6ul34nk) 

[Hydra Extended Override Syntax](https://docs.google.com/document/d/1JDZGHKk4PrZHqsTTS6ao-DQOu2eVD4ULR6uAxVUR-WI/edit#) describes the design of a followup extension to the syntax to enable more powerful sweeping using a well defined syntax.

Comments welcome on both

## Description
Closes #738
Closes #573 

### Must
- [x] Fix setup.py to build antlr grammar automatically on install, sdist, and develop.
- [x] Update Example Sweeper plugin
- [x] Update Ax Sweeper plugin (#752)
- [x] Update Nevergrad Sweeper plugin (#752)
- [x] Add tests for list and dict command line overrides
- [x] Fix CI on Windows (issues with isort not skipping generated parser code there)

### Features (could be deferred)
- [x] Support scientific floating point notation
- [x] Support type casting through CLI (#752)
- [x] Define sweep filters and ordering syntax in Override DSL (globing? regex? random shuffle?)(#752)
- [x] Define syntax for range sweep (integers, floats, with and without step size). (#752)

### Docs
- [x] Docs update
- [x] News fragments


romesc-CMU:
Just want to say, I've been testing this out with multirun, and it's very nice! Every arbitrarily complex thing I've passed in the CLI so far has worked nicely.

omry:
Thanks for playing with this!
I really like it too.
Take a look at the extended syntax (second document in the description) to see what extensions are planned.

PR #749, opened 2020-07-08
Only copy timestamp to sweep config cache

jan-matthis:
## Motivation

Changes `load_sweep_config` to only copy timestamp information from the master config cache. Previously, the full cache was copied, which could lead to unexpected resolution e.g. of environment variables on remote hosts.

Yes

## Test Plan

TBD

## Related Issues and PRs

https://github.com/facebookresearch/hydra/pull/744

### DIFF
--- a/tests/test_config_loader.py
+++ b/tests/test_config_loader.py
@@ -1,4 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+import os
 import re
 from dataclasses import dataclass
 from typing import Any, List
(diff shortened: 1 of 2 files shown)


tests/test_config_loader.py
```diff
@@ -446,6 +446,34 @@ def test_assign_null(
         assert cfg == {"normal_yaml_config": True, "abc": None}
         assert len(recwarn) == 0
 
+    def test_sweep_config_cache(self, path: str) -> None:
```
omry, line 449: That's a good start. but you need to test two things:
1. that `now` is cached.
2. that HOME environment variable is not.

I I suspect you are only testing one of the two.

jan-matthis: That‘s taken care of here: `assert "now" in sweep_cfg_cache.keys() and len(sweep_cfg_cache.keys()) == 1`. Nothing except the now key must be in the cache in the test.

omry: well, test it functionally:
monkeypatch the environment variable after you access it the first time and see that you get the new value.

jan-matthis: Alright, this is addressed in the latest commit

tests/test_config_loader.py
```diff
@@ -446,6 +447,38 @@ def test_assign_null(
...
+
+        sweep_cfg = config_loader.load_sweep_config(
+            master_config=master_cfg,
+            sweep_overrides=["time='${now:%H-%M-%S}'", "home='${env:HOME}'"],
+        )
+
+        sweep_cfg_cache = OmegaConf.get_cache(sweep_cfg)
+        assert len(sweep_cfg_cache.keys()) == 1 and "now" in sweep_cfg_cache.keys()
+        assert sweep_cfg_cache["now"] == master_cfg_cache["now"]
+        os.environ["HOME"] = "/another/home/dir/"
```
omry, line 479: This is now changing the HOME environment variable for the rest of the tests.

Use [monkeypatch](https://docs.pytest.org/en/stable/monkeypatch.html#monkeypatching-environment-variables) to undo it automatically.

jan-matthis: Good catch and thanks for the pointer! Addressed in latest commit

omry:
awesome

PR #777, opened 2020-07-15
Add resolver for runtime python version

jieru-hu:
## Motivation

(I'm breaking all the hydra core related changes introduced by Ray launcher into smaller PRs)

With this change we can detect runtime python version, this is helpful for configuring hydra when version info is important.

for example when launching on remote machines we will need ```cloudpickle``` which only works when pickle and unpickle with the same python version.

Testing
```
>>> import sys
>>> import omegaconf
>>> vi = sys.version_info
>>> version_dict = {
...         "major": vi.major,
...         "minor": f"{vi.major}.{vi.minor}",
...         "micro": f"{vi.major}.{vi.minor}.{vi.micro}",
...     }
>>> f=lambda level="minor": version_dict.get(level)
>>> from omegaconf import OmegaConf
>>> c = OmegaConf.create({'key': '${python_version:}'}).key
>>> c
'3.8'
>>> c = OmegaConf.create({'key': '${python_version:major}'}).key
>>> c
3
>>> c = OmegaConf.create({'key': '${python_version:minor}'}).key
>>> c
'3.8'
>>> c = OmegaConf.create({'key': '${python_version:micro}'}).key
>>> c
'3.8.2'
>>> 

```

Yes/No

## Test Plan

## Related Issues and PRs

### DIFF
--- a/tests/test_config_loader.py
+++ b/tests/test_config_loader.py
@@ -443,7 +443,9 @@ class TestConfigLoader:
         assert cfg == {"normal_yaml_config": True, "abc": None}
         assert len(recwarn) == 0
 
-    def test_sweep_config_cache(self, path: str, monkeypatch: Any) -> None:
+    def test_sweep_config_cache(
+        self, hydra_restore_singletons: Any, path: str, monkeypatch: Any
+    ) -> None:
         setup_globals()
 
         config_loader = ConfigLoaderImpl(
--- a/tests/test_core_utils.py
+++ b/tests/test_core_utils.py
@@ -28,15 +28,14 @@ def test_accessing_hydra_config(hydra_restore_singletons: Any) -> Any:
     assert cfg.config_name == "accessing_hydra_config"
 
 
-def test_py_version_resolver(hydra_restore_singletons: Any, monkeypatch: Any) -> Any:
+def test_py_version_resolver(
+    hydra_restore_singletons: Any, monkeypatch: Any, restore_resolvers: Any
+) -> Any:
     Version = namedtuple("Version", "major minor micro")
-    with monkeypatch.context() as m:
-        m.setattr(sys, "version_info", Version(3, 7, 2))
-        OmegaConf.clear_resolvers()
-        utils.setup_globals()
-        assert OmegaConf.create({"key": "${python_version:}"}).key == "3.7"
-        assert OmegaConf.create({"key": "${python_version:major}"}).key == "3"
-        assert OmegaConf.create({"key": "${python_version:minor}"}).key == "3.7"
-        assert OmegaConf.create({"key": "${python_version:micro}"}).key == "3.7.2"
-    # reset OmegaConf resolver with no patching
+    monkeypatch.setattr(sys, "version_info", Version(3, 7, 2))
+    OmegaConf.clear_resolvers()
     utils.setup_globals()
+    assert OmegaConf.create({"key": "${python_version:}"}).key == "3.7"
+    assert OmegaConf.create({"key": "${python_version:major}"}).key == "3"
+    assert OmegaConf.create({"key": "${python_version:minor}"}).key == "3.7"
+    assert OmegaConf.create({"key": "${python_version:micro}"}).key == "3.7.2"
--- a/tests/test_core_utils.py
+++ b/tests/test_core_utils.py
@@ -1,7 +1,9 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+import sys
+from collections import namedtuple
 from typing import Any
 
-from omegaconf import open_dict
+from omegaconf import OmegaConf, open_dict
 
 from hydra._internal.config_loader_impl import ConfigLoaderImpl
 from hydra._internal.utils import create_config_search_path
@@ -24,3 +26,17 @@ def test_accessing_hydra_config(hydra_restore_singletons: Any) -> Any:
         del cfg["hydra"]
     assert cfg.job_name == "UNKNOWN_NAME"
     assert cfg.config_name == "accessing_hydra_config"
+
+
+def test_py_version_resolver(hydra_restore_singletons: Any, monkeypatch: Any) -> Any:
+    Version = namedtuple("Version", "major minor micro")
+    with monkeypatch.context() as m:
+        m.setattr(sys, "version_info", Version(3, 7, 2))
+        OmegaConf.clear_resolvers()
+        utils.setup_globals()
+        assert OmegaConf.create({"key": "${python_version:}"}).key == "3.7"
+        assert OmegaConf.create({"key": "${python_version:major}"}).key == "3"
+        assert OmegaConf.create({"key": "${python_version:minor}"}).key == "3.7"
+        assert OmegaConf.create({"key": "${python_version:micro}"}).key == "3.7.2"
+    # reset OmegaConf resolver with no patching
+    utils.setup_globals()
--- a/tests/test_core_utils.py
+++ b/tests/test_core_utils.py
@@ -13,7 +13,7 @@ from hydra.types import RunMode
 
 
 def test_accessing_hydra_config(
-    hydra_restore_singletons: Any, restore_resolvers: Any
+    hydra_restore_singletons: Any, clear_resolvers: Any
 ) -> Any:
     utils.setup_globals()
 
@@ -31,7 +31,7 @@ def test_accessing_hydra_config(
 
 
 def test_py_version_resolver(
-    hydra_restore_singletons: Any, monkeypatch: Any, restore_resolvers: Any
+    hydra_restore_singletons: Any, monkeypatch: Any, clear_resolvers: Any
 ) -> Any:
     Version = namedtuple("Version", "major minor micro")
     monkeypatch.setattr(sys, "version_info", Version(3, 7, 2))
--- a/tests/test_core_utils.py
+++ b/tests/test_core_utils.py
@@ -1,6 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 import sys
-from collections import namedtuple
 from typing import Any
 
 from omegaconf import OmegaConf, open_dict
@@ -29,8 +28,7 @@ def test_accessing_hydra_config(hydra_restore_singletons: Any) -> Any:
 
 
 def test_py_version_resolver(hydra_restore_singletons: Any, monkeypatch: Any) -> Any:
-    Version = namedtuple("Version", "major minor micro")
-    monkeypatch.setattr(sys, "version_info", Version(3, 7, 2))
+    monkeypatch.setattr(sys, "version_info", (3, 7, 2))
     utils.setup_globals()
     assert OmegaConf.create({"key": "${python_version:}"}).key == "3.7"
     assert OmegaConf.create({"key": "${python_version:major}"}).key == "3"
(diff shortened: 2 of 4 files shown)


omry:
You can use python_version with an optional level:

````python
In [20]: def foo(x : Optional[str] = None):
    ...:     print(x)
    ...: 

In [21]: OmegaConf.register_resolver("foo", foo)

In [22]: OmegaConf.create({"x" : "${foo:}"}).x
None

In [23]: OmegaConf.create({"x" : "${foo:bar}"}).x
bar
```

jieru-hu:
> You can use python_version with an optional level:
> 
> ```python
> In [20]: def foo(x : Optional[str] = None):
>     ...:     print(x)
>     ...: 
> 
> In [21]: OmegaConf.register_resolver("foo", foo)
> 
> In [22]: OmegaConf.create({"x" : "${foo:}"}).x
> None
> 
> In [23]: OmegaConf.create({"x" : "${foo:bar}"}).x
> bar
> ```
> ```

thx for the tips. I've updated the PR along with the test plan.

omry:
Testing and documentation.

jieru-hu:
> Please respond to my docs and tests comment

thx for the comments. I've updated the description of the PR with testing done. and the document are in this PR https://github.com/facebookresearch/hydra/pull/780

omry:
The testing in the test plan is not sufficient. Please add real unit tests.

tests/test_core_utils.py
```diff
@@ -24,3 +25,21 @@ def test_accessing_hydra_config(hydra_restore_singletons: Any) -> Any:
         del cfg["hydra"]
     assert cfg.job_name == "UNKNOWN_NAME"
     assert cfg.config_name == "accessing_hydra_config"
+
+
+def test_py_version_resolver(hydra_restore_singletons: Any) -> Any:
+    utils.setup_globals()
+    vi = sys.version_info
```
omry, line 32: it would be cleaner to mock the version.
this is similar to  what I have done at https://github.com/antlr/antlr4/pull/2866/files#diff-5895bc8c44907952f2f9f6e4e212df6cR12.

You can probably use Pytest's monkeypatch to do this more cleanly.

omry: looks like monkeypatch is not actually working here? (see CI failures)

jieru-hu: yea. It passed locally on my pytest somehow. Im fixing this :) (I've turned the PR into a draft)

tests/test_core_utils.py
```diff
@@ -24,3 +26,17 @@ def test_accessing_hydra_config(hydra_restore_singletons: Any) -> Any:
         del cfg["hydra"]
     assert cfg.job_name == "UNKNOWN_NAME"
     assert cfg.config_name == "accessing_hydra_config"
+
+
+def test_py_version_resolver(hydra_restore_singletons: Any, monkeypatch: Any) -> Any:
+    Version = namedtuple("Version", "major minor micro")
+    with monkeypatch.context() as m:
+        m.setattr(sys, "version_info", Version(3, 7, 2))
+        OmegaConf.clear_resolvers()
```
omry, line 35: what happens to the resolvers after this function finishes?

omry: for now I suggest that you copy [this](https://github.com/omry/omegaconf/blob/master/tests/conftest.py#L9) to automatically restore the resolvers.

It would probably make sense to introduce it as a pytest plugin in OmegaConf similar to the pytest plugin in Hydra.

jieru-hu: > what happens to the resolvers after this function finishes?

Since OmegaConf won't let you overwrite the resolver, the test fails if the the resolver has already been set by other tests (the monkeypatch won't take effect), which is why i try to clear and reset the resolvers after the tests are done.

> It would probably make sense to introduce it as a pytest plugin in OmegaConf similar to the pytest plugin in Hydra.

I'm not sure what you mean here: import the fixture from omegaconf into hydra pytest_plugin.py somehow and call it?

Or do you mean copy the fixture from to ```hydra/extra/pytest_plugin.py``` and use it?

omry: Tempoary: copy the code.
Real solution, send a PR against OmegaConf introducing a pytest plugin like the one I added to Hydra.
This would allow any pytest user that has OmegaConf install to use this fixture.

jieru-hu: sounds good, thx for clarifying

tests/test_core_utils.py
```diff
@@ -28,15 +28,14 @@ def test_accessing_hydra_config(hydra_restore_singletons: Any) -> Any:
...
-        OmegaConf.clear_resolvers()
-        utils.setup_globals()
-        assert OmegaConf.create({"key": "${python_version:}"}).key == "3.7"
-        assert OmegaConf.create({"key": "${python_version:major}"}).key == "3"
-        assert OmegaConf.create({"key": "${python_version:minor}"}).key == "3.7"
-        assert OmegaConf.create({"key": "${python_version:micro}"}).key == "3.7.2"
-    # reset OmegaConf resolver with no patching
+    monkeypatch.setattr(sys, "version_info", Version(3, 7, 2))
+    OmegaConf.clear_resolvers()
     utils.setup_globals()
```
omry: I don't think you need those two lines now.

jieru-hu: Still need them :)

```python
$ git diff
diff --git a/tests/test_core_utils.py b/tests/test_core_utils.py
index 5dbf3e85..ae94e01f 100644
--- a/tests/test_core_utils.py
+++ b/tests/test_core_utils.py
@@ -33,8 +33,8 @@ def test_py_version_resolver(
 ) -> Any:
     Version = namedtuple("Version", "major minor micro")
     monkeypatch.setattr(sys, "version_info", Version(3, 7, 2))
-    OmegaConf.clear_resolvers()
-    utils.setup_globals()
+    # OmegaConf.clear_resolvers()
+    # utils.setup_globals()
     assert OmegaConf.create({"key": "${python_version:}"}).key == "3.7"
     assert OmegaConf.create({"key": "${python_version:major}"}).key == "3"


$ pytest tests/test_core_utils.py 
=============================================================================== test session starts ================================================================================
platform darwin -- Python 3.8.2, pytest-5.4.2, py-1.8.1, pluggy-0.13.1
rootdir: /Users/jieru/workspace/hydra-fork/hydra, inifile: pytest.ini
plugins: snail-0.1.0, hydra-core-1.0.0rc3
collected 2 items                                                                                                                                                                  

tests/test_core_utils.py .F                                                                                                                                                  [100%]

===================================================================================== FAILURES =====================================================================================
_____________________________________________________________________________ test_py_version_resolver _____________________________________________________________________________

hydra_restore_singletons = None, monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7fea682d7b50>, restore_resolvers = None

    def test_py_version_resolver(
        hydra_restore_singletons: Any, monkeypatch: Any, restore_resolvers: Any
    ) -> Any:
        Version = namedtuple("Version", "major minor micro")
        monkeypatch.setattr(sys, "version_info", Version(3, 7, 2))
        # OmegaConf.clear_resolvers()
        # utils.setup_globals()
>       assert OmegaConf.create({"key": "${python_version:}"}).key == "3.7"
E       AssertionError: assert '3.8' == '3.7'
E         - 3.7
E         + 3.8

tests/test_core_utils.py:38: AssertionError
============================================================================= short test summary info ==============================================================================
FAILED tests/test_core_utils.py::test_py_version_resolver - AssertionError: assert '3.8' == '3.7'
=========================================================================== 1 failed, 1 passed in 0.25s ============================================================================
(pytest38) jieru-mbp:hydra jieru$ 

```

omry: Why? could this be related to the OmegaConf cache?
try OmegaConf.clear_cache()

jieru-hu: sure i can try that. [`clear_cache`](https://github.com/omry/omegaconf/blob/a6fde5a748be264f7d8344bc086e0b7aa237cb5d/omegaconf/omegaconf.py#L378) expects a  `conf` param, what should that be in this case? 

I think the reason that it fails is there is an another test `test_accessing_hydra_config` that runs first before this one, and as a result it already register the python_version function before this test run. Since OmegaConf won't let `setup_globals` override the resolver, the monkeypatched function never gets called. as a result, i think these 2 lines of code are needed.

omry: That's actually an error in the API. the cache it's clearing is static so it should not require a config.

Generally speaking, tests should clear after themselves and not before they run.
not sure how hard it is to fix in the current state of things.

jieru-hu: Yea you are right, this does feel weird. Instead of cleaning up before the run in this test, I've edited the other tests to clean up after themselves.

jieru-hu: Did some further digging. I think we need `clear_resolver` instead of `restore`. Added a fixture for that instead.

omry: The idea with tests is that they are not supposed to change the state of the system.
whatever was before, should stay.  Your fix is sweeping the problem under the rug.
It can be a bit tricky to find which test is changing the state of the system. one way I have been using is to  run multiple tests at one, with the test that triggers the problem last. by running different test files before the one the blows up you can find which one is causing issues.

omry:
Looks much better.
can you also restore the resolver cache?

[API](https://github.com/omry/omegaconf/blob/5e133e2d84f588f7eb79e9c3a52bedd27b47d2d1/omegaconf/omegaconf.py#L370-L376).

jieru-hu:
> see my comment.

@omry 
The API takes in a ```conf: BaseContainer```, what should I pass in in this case?

jieru-hu:
> > see my comment.
> 
> @omry
> The API takes in a `conf: BaseContainer`, I'm not sure what should be passed in here. Thoughts?
ping @omry, in case this gets lost in your inbox :)

omry:
My bad, it's per config. no need to clear it.

tests/test_config_loader.py
```diff
@@ -443,7 +443,9 @@ def test_assign_null(
         assert cfg == {"normal_yaml_config": True, "abc": None}
         assert len(recwarn) == 0
 
-    def test_sweep_config_cache(self, path: str, monkeypatch: Any) -> None:
+    def test_sweep_config_cache(
```
omry, line 446: See, you have found the real culprit!

tests/test_core_utils.py
```diff
@@ -24,3 +26,13 @@ def test_accessing_hydra_config(hydra_restore_singletons: Any) -> Any:
         del cfg["hydra"]
     assert cfg.job_name == "UNKNOWN_NAME"
     assert cfg.config_name == "accessing_hydra_config"
+
+
+def test_py_version_resolver(hydra_restore_singletons: Any, monkeypatch: Any) -> Any:
+    Version = namedtuple("Version", "major minor micro")
```
omry, line 32: cam you drop the namedtuple here?
I don't think it's worth depending on it for such a tiny thing.

jieru-hu: :( I'm not sure how else to mock this 😰

omry: This:
```
version = (3,7,2)
```
Does not work?

jieru-hu: > This:
> 
> ```
> version = (3,7,2)
> ```
> 
> Does not work?

didn't work before because I was calling ```sys.version_info.major```, changed the code to ```sys.version_info[0]``` now and removed the dependency.

omry: Yeah, it's a tuple although it has bigger aspirations :) :
https://docs.python.org/3.1/library/sys.html#sys.version_info

```python
In [10]: sys.version_info == (3,8,2,"final", 0)
Out[10]: True
```

PR #781, opened 2020-07-16
Sanitize error stack frames by default

### LINKED ISSUE #647: [Bug] config access errors in the code should include stack trace
When the user is making a config access error in the code the stack trace should include the offending file+line.

omry:
Improve printed stack trace to contain only interesting stack frames by default.
stripping hydra frames from the start and OmegaConf frames from the bottom.

closes #647

### DIFF
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -6,8 +6,10 @@ import logging.config
 import os
 import sys
 import warnings
+from dataclasses import dataclass
 from os.path import dirname, join, normpath, realpath
-from traceback import print_exc
+from traceback import print_exc, print_exception
+from types import FrameType
 from typing import (
     Any,
     Callable,
@@ -21,7 +23,6 @@ from typing import (
 )
 
 from omegaconf import DictConfig, OmegaConf, read_write
-from omegaconf.errors import OmegaConfBaseException
 
 from hydra._internal.config_search_path_impl import ConfigSearchPathImpl
 from hydra.core.config_search_path import ConfigSearchPath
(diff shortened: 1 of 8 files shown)


tests/test_apps/app_with_runtime_config_error/my_app.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+from omegaconf import DictConfig
+
+import hydra
+
+
+def foo(cfg: DictConfig) -> None:
```
jieru-hu, line 7: I don't understand why ```def foo(cfg: DictConfig)``` is needed in this example. Doesnt that just make it more complicated? is it needed to create a more sophisticated stack trace?

hydra/_internal/utils.py
```diff
@@ -195,16 +196,74 @@ def create_config_search_path(search_path_dir: Optional[str]) -> ConfigSearchPat
...
         return func()
-    except (HydraException, OmegaConfBaseException) as ex:
+    except Exception as ex:
         if "HYDRA_FULL_ERROR" in os.environ and os.environ["HYDRA_FULL_ERROR"] == "1":
             print_exc()
         else:
-            sys.stderr.write(str(ex) + "\n")
-            if ex.__cause__ is not None:
-                sys.stderr.write(str(ex.__cause__))
+            if isinstance(ex, HydraException):
```
jieru-hu, line 203: 😮👍

tests/test_apps/app_with_runtime_config_error/my_app.py
omry: yes. I want to show that this shows all interesting stack frames for the user, not just the one with the error.

PR #806, opened 2020-07-21
Adding new Hydra logo to README

iamdarshshah:
## Motivation
Thought of making README.md more catchy by adding Logo to it. And it's a best practice to have a logo in an Open-Source project.
- [x] Updating Readme.md file by adding the Hydra logo.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebookresearch/hydra/blob/master/CONTRIBUTING.md)?

Yes

> Screenshot: 
Hydra

PR #814, opened 2020-07-23
Added Official Logo

sawravchy:


omry:
Hi @sawravchy.
Thanks for the PR, there is already a similar one open (#806).

Issue #752, opened 2020-07-10
Design and implement Hydra Extended Override Syntax


omry:
The extended override syntax aims to support optimizing sweepers like Ax and Nevergrad, as well as add some fundamental capabilities like type casting in the command line.
[Document](https://docs.google.com/document/d/1JDZGHKk4PrZHqsTTS6ao-DQOu2eVD4ULR6uAxVUR-WI/edit#)

Follow task will be:
- Update Ax Sweeper Plugin to take advantage of new Overrides Parser
- Update Nevergrad Sweeper Plugin to take advantage of new Overrides Parser

PR #834, opened 2020-07-29
Expanding instructions for list CLI override

rohitgirdhar:
## Motivation

I just spent a bunch of time debugging a command line override that involved nested lists (something like `param='[["a", 1, 2], ["b", 2, 3]]'`, which gave me a `hydra.errors.OverrideParseException: no viable alternative at input` error until I removed the space between the 2 sublists. So this tip might help future users.

Yes

## Test Plan
Not needed

### DIFF
--- a/website/docs/advanced/command_line_syntax.md
+++ b/website/docs/advanced/command_line_syntax.md
@@ -134,7 +134,7 @@ Constants (null, true, false, inf, nan) are case insensitive.
 foo=[1,2,3]
 nested=[a,[b,[c]]]
 ```
-**IMPORTANT** Always single-quote overrides that contains lists in the shell.
+**IMPORTANT** Always single-quote overrides that contains lists in the shell.  Also, note the *no spaces* between list elements, otherwise you might get a `hydra.errors.OverrideParseException: no viable alternative at input` error.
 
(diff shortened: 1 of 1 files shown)


omry:
Thanks for reporting @rohitgirdhar.
Please file issues for discussion before sending pull request.
In this case this is actually a bug, not a documentation issue.

omry:
#836

omry:
This is not a list, it's a quoted string that looks like one.
```
param='[["a", 1, 2], ["b", 2, 3]]'
```
This is a bit confusing because both Hydra and the shell and using quotes.

This is the correct form:
```
'param=[[a, 1, 2], [b, 2, 3]]'
```

The quotes here are for the shell, to make sure it sends this as a single command line argument.
Note that you don't need to quote `a` and `b`, This is not JSON.
I have a fix coming for this issue as a part of a bigger change to the grammar (#752).

omry:
This is actually a list, but the way you present it confused even me :).

You can tell what the shell will send to Hydra using echo:
```
$ echo param='[["a", 1, 2], ["b", 2, 3]]'
param=[["a", 1, 2], ["b", 2, 3]]
```

Issue #858, opened 2020-08-10
[Bug] autocompletion is not working if PATH environment has `.` in the list


thoth291:
# 🐛 Bug
## Description
Basically when you install it for that terminal it is not working.
So I did comparison with my MacOS installation (all the same versions of hydra and stuff) and got that the difference in behavior comes from `command` tool.
Here is how it reacts on working side (on my mac):
```shell
$ command -v python simple.py
/path/to/bin/python
```
And how it reacts in CentOS:
```shell
$ command -v python simple.py
/path/to/bin/python
./simple.py
```
## Expected Behavior
I was able to fix it by replacing this in the autocompletion installation script:
```bash
 if ! [ -x "$(command -v $helper)" ]; then
        false
 fi
```
to this one:
```bash
cmd=($(command -v $helper))
if ! [ -x "${cmd[0]}" ]; then
        false
fi
```

## System information
- **Hydra Version** :  1.0.0rc3
- **Python version** : 3.8.3
- **Virtual environment type and version** : conda, 4.8.3
- **Operating system** : CentOS7
- **Bash Version** : 4.2.46(2)

omry:
Can you address this question at KDE?

thoth291:
@omry , I've tried all possible versions of centos I had access to (from 7.x-8.x) - they all was behaving as your script would expect. I need to dig dipper to see what exactly going on in our systems - it might be some discrepancy between builtin tools. I'll check with systems team and will come back to you.
Thanks!

omry:
Thanks for digging.

thoth291:
@omry , I tracked it down to 2 possible workarounds.
In some systems when you have "." in the PATH - it will interpret your second argument as yet another executable and will report it - that's why I'm getting 2 lines.
But if I use system-wide (not shell-wide) command: return of `which command` - then my experience is regular and not depends on PATH having `.`.
I discussed it with some systems people and they told me that some other clusters in national labs will have the say issue - since their setup is similar to ours.
So I would suggest to change bash autocompletion script using the setup mentioned above or using this method:
```bash
COMMAND=$(which command)
if ! [ -x "$($COMMAND -v $helper)" ]; then
     false
fi
```
It should fix the issue and make it working for everyone regardless of their system settings.

omry:
Nice find.
Feel free put up a pull request. Try to include tests that are testing for this scenario.

thoth291:
```
Try to include tests that are testing for this scenario.
```
[...]

PR #868, opened 2020-08-12
Fix bug with bash autocompletion for the users who has `.` in their PATH environment variable

thoth291:
## Motivation

This should address issue of builtins to consider local files as executables during parsing of the command.
This will fix issues with PATH having "." and script being executable.

## Test Plan

Need to figure out how to make this into a test here - need some help.

```python
import subprocess
#import textwrap

def test_bash_completion():
    process = subprocess.Popen(
#            textwrap.dedent(
"""
bash -l <<- 'HEREDOC'
# Navigate to app with minimal configuration
cd .

# Locate python
PYTHON=$(command -v python)

# Setup hydra bash completion
eval "$(python simple.py -sc install=bash)"

# Setup debugging
export HYDRA_COMP_DEBUG=1

# Do actual test
test=$(COMP_LINE='python simple.py' hydra_bash_completion | grep EXECUTABLE_FIRST | awk '{split($0,a,"=");b=substr(a[2],2,length(a[2])-2);print b}')

if [ $test == $PYTHON ]; then
    echo TRUE
else
    echo FALSE
fi
HEREDOC
"""
#           )
    ,
    shell=True,
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE,
    )
    stdout, stderr = process.communicate()
    assert stderr == b''
    assert stdout == b'TRUE\n'
    return
```

## Related Issues and PRs
Look here: https://github.com/facebookresearch/hydra/issues/858

### DIFF
--- a/hydra/_internal/core_plugins/bash_completion.py
+++ b/hydra/_internal/core_plugins/bash_completion.py
@@ -32,7 +32,7 @@ class BashCompletion(CompletionPlugin):
         helper="${words[0]}"
         true
     fi
-    
+
     EXECUTABLE=($(command -v $helper))
     if [ "$HYDRA_COMP_DEBUG" == "1" ]; then
         printf "\\n"
--- a/hydra/_internal/core_plugins/bash_completion.py
+++ b/hydra/_internal/core_plugins/bash_completion.py
@@ -32,8 +32,17 @@ class BashCompletion(CompletionPlugin):
         helper="${words[0]}"
         true
     fi
-
-    if ! [ -x "$(command -v $helper)" ]; then
+    
+    EXECUTABLE=($(command -v $helper))
+    if [ "$HYDRA_COMP_DEBUG" == "1" ]; then
+        printf "\\n"
+        printf "helper='$helper'\\n"
+        printf "EXECUTABLE='$EXECUTABLE'\\n"
+        printf "EXECUTABLE_FIRST='${EXECUTABLE[0]}'\\n"
+        printf "EXECUTABLE AS ARRAY:\\n"
+        printf "\\t%s\\n" ${EXECUTABLE[@]}
+    fi
+    if ! [ -x "${EXECUTABLE[0]}" ]; then
         false
     fi
 
--- a/hydra/_internal/core_plugins/bash_completion.py
+++ b/hydra/_internal/core_plugins/bash_completion.py
@@ -35,8 +35,7 @@ class BashCompletion(CompletionPlugin):
 
     EXECUTABLE=($(command -v $helper))
     if [ "$HYDRA_COMP_DEBUG" == "1" ]; then
-        printf "\\n"
-        printf "helper='$helper'\\n"
+        printf "\\nhelper='$helper'\\n"
         printf "EXECUTABLE='$EXECUTABLE'\\n"
         printf "EXECUTABLE_FIRST='${EXECUTABLE[0]}'\\n"
         printf "EXECUTABLE AS ARRAY:\\n"
--- a/hydra/_internal/core_plugins/bash_completion.py
+++ b/hydra/_internal/core_plugins/bash_completion.py
@@ -35,11 +35,7 @@ class BashCompletion(CompletionPlugin):
 
     EXECUTABLE=($(command -v $helper))
     if [ "$HYDRA_COMP_DEBUG" == "1" ]; then
-        printf "\\nhelper='$helper'\\n"
-        printf "EXECUTABLE='$EXECUTABLE'\\n"
         printf "EXECUTABLE_FIRST='${EXECUTABLE[0]}'\\n"
-        printf "EXECUTABLE AS ARRAY:\\n"
-        printf "\\t%s\\n" ${EXECUTABLE[@]}
     fi
     if ! [ -x "${EXECUTABLE[0]}" ]; then
         false
--- a/tests/scripts/test_bash_dot_in_path.bash
+++ b/tests/scripts/test_bash_dot_in_path.bash
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# Navigate to app with minimal configuration
+cd tests/test_apps/app_with_cfg/
+
+# Locate python
+PYTHON=$(command -v python)
+
+# Setup hydra bash completion
+eval "$(python my_app.py -sc install=bash)"
+
+# Setup debugging
+export HYDRA_COMP_DEBUG=1
+export PATH=$PATH:.
+
+# Do actual test
+test=$(COMP_LINE='python my_app.py' hydra_bash_completion |\
+    grep EXECUTABLE_FIRST |\
+    awk '{split($0,a,"=");b=substr(a[2],2,length(a[2])-2);print b}')
+
+if [ $test == $PYTHON ]; then
+    echo TRUE
+else
+    echo FALSE
+fi
--- a/tests/test_completion.py
+++ b/tests/test_completion.py
@@ -53,46 +53,48 @@ def create_config_loader() -> ConfigLoaderImpl:
         )
     )
 
+
 @pytest.mark.skipif(sys.platform == "win32", reason="does not run on windows")
-def test_bash_completion_with_dot_in_path() -> None:
-    process = subprocess.Popen(
-        textwrap.dedent(
-            """
-        bash <<- 'HEREDOC'
-        # Navigate to app with minimal configuration
-        cd tests/test_apps/app_with_cfg/
+class TestDotInPathCompletion:
+    def test_bash_completion_with_dot_in_path(self) -> None:
+        process = subprocess.Popen(
+            textwrap.dedent(
+                """
+                bash <<- 'HEREDOC'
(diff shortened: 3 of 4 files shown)


thoth291:
Hi, @omry . It passed all my tests in my setup - so this solution seems to be mature. You asked me to make a test - I shared it above (https://github.com/facebookresearch/hydra/pull/868#issue-466510843) - but I don't know how to put it in your framework for tests - could you please put it together yourself?

hydra/_internal/core_plugins/bash_completion.py
```diff
@@ -33,7 +33,16 @@ def install(self) -> None:
...
 
-    if ! [ -x "$(command -v $helper)" ]; then
+    EXECUTABLE=($(command -v $helper))
+    if [ "$HYDRA_COMP_DEBUG" == "1" ]; then
+        printf "\\n"
+        printf "helper='$helper'\\n"
+        printf "EXECUTABLE='$EXECUTABLE'\\n"
+        printf "EXECUTABLE_FIRST='${EXECUTABLE[0]}'\\n"
+        printf "EXECUTABLE AS ARRAY:\\n"
+        printf "\\t%s\\n" ${EXECUTABLE[@]}
```
omry, line 43: make those print just one line.

thoth291: Just merged 2 lines into 1 - is that fine?
Merging others would break clarity of the code (to my view).

hydra/_internal/core_plugins/bash_completion.py
```diff
@@ -33,7 +33,15 @@ def install(self) -> None:
...
 
-    if ! [ -x "$(command -v $helper)" ]; then
+    EXECUTABLE=($(command -v $helper))
+    if [ "$HYDRA_COMP_DEBUG" == "1" ]; then
+        printf "\\nhelper='$helper'\\n"
+        printf "EXECUTABLE='$EXECUTABLE'\\n"
+        printf "EXECUTABLE_FIRST='${EXECUTABLE[0]}'\\n"
+        printf "EXECUTABLE AS ARRAY:\\n"
+        printf "\\t%s\\n" ${EXECUTABLE[@]}
+    fi
```
omry, line 43: I mean the whole thing. I don't think it's worth destroying the debug output for this.
```suggestion
```

omry:
Can you create a test bash script that validates this behavior of the bash function you changed?
you can execute it from the test suite as a subprocess.

You can add your script under tests/scripts

hydra/_internal/core_plugins/bash_completion.py
thoth291: I use debug output (EXECUTABLE_FIRST line) during my test. And all of these lines were useful for me during debugging. Packaging them in one line will destroy clarity of those lines. So I hesitate to package them all into one line.

omry: Your debugging is now over. This will get in the way of anyone else that wants to debug the completion.

I am not actually seeing a test here, a comment with code does not count :)

thoth291:
I don't know how to make it to see proper python within my script. Is there any environment I need to activate?

omry:
> I don't know how to make it to see proper python within my script. Is there any environment I need to activate?

Is this working locally with pytest and with nox?
see: https://hydra.cc/docs/next/development/testing

thoth291:
Yes, @omry, I tested it with pytest and it passes (locally). I don't really know why it doesn't pass it here in the CI.

omry:
The CI is running nox. you can run it locally. see the doc I send you.
Does it work when you use nox locally (you can tell it to run a specific session).

omry:
Getting close.
You have lint failures, the the [style_guide](https://hydra.cc/docs/next/development/style_guide) for details on how to test locally.

omry:
Also, please add a [news fragment](https://hydra.cc/docs/next/development/documentation#news-entries).

thoth291:
They named `mypy` wrong - it should have been called `mypain`...
The issue of original failing of the test was due to `bash` v.s. `bash -l` - the later do not inherit env variables properly on the test machine.
The rest of the commits were due to tests.

tests/test_completion.py
```diff
@@ -53,6 +54,49 @@ def create_config_loader() -> ConfigLoaderImpl:
...
+
+                if [ $test == $PYTHON ]; then
+                    echo TRUE
+                else
+                    echo FALSE
+                fi
+                HEREDOC
+                """
+            ),
+            shell=True,
```
omry, line 90: `shell=True` is insecure in general.
Can you switch to keeping the shell script in a simple file and calling with subprocess.check_output() ?

thoth291: This is not difficult, but against my habits. In the projects I worked so far we use pytest (and similar things for C/C++/Fortran) and we always make our tests as self contained as possible. So we keep all things in one place - to give ultimate visibility of what is really being tested at particular test.
I would actually suggest you to change some of your expect scripts to my format to make sure that everything in one place.
I have an anecdotal story when push from my colleague changed many files at ones and one of them was part of the test suite - but then at final merge another colleague changed how that test was executed (unintentionally) which caused test to pass for all cases (even where it should have been failed).
If that test would be in the script (like the one above) - this would not happen.

P.S. I have to admit - I never used mypy and nox before - and given what I've seen I will probably not use them any time soon - IMHO - until they will mature a bit more and will make things more automatic.

omry: Okay, I accept your reasoning. but do get rid of the shell=True somehow.

As for mypy and nox:
1. nox is fantastic. What issues did you have with it?
2. mypy is industry standard.
The CI is intentionally testing a lot of things. It does take time to handle all the requirements, as especially given that it's your first PR.

omry: As for the expect script: I am expecting to have a change to it every 10 years.

thoth291: This is your repo - you are the boss here - I'll split it into separate file and will make shell=False.
I said - it's easy - was just sharing some of my personal observations.
P.S. I just take testing easy - nox is like "heavy metal for relaxation" for me (don't take me wrong - I love heavy metal - but consider it to be relaxing only during Armageddons). This is indeed my first PR with mypy - I guess I don't belong to the same industry :-)).

omry: I agree with your reasoning to keep it in one place.
what I am suggesting is that your test can dump a temp file from the script and execute it with bash.

omry: I am being strict on testing because I hate breaking things unknowingly.
If something breaks, it would better be because I intentionally broke it and not because of poor test coverage.

thoth291: Creating files on the fly for the tests is OK if it's data files - but executable scripts - that's something which I try to avoid due to the way different systems handle temporary files and setting permissions - so I decided to split it all together.

omry: 1. sure.
2. This is not an executable, just a text file that is being executed by bash.
3. This shows you the power of good CI. if it your tests pass in CI you know it works on Linux Mac and Windows.

tests/scripts/test_bash_dot_in_path.bash
```diff
@@ -0,0 +1,25 @@
...
+# Setup debugging
+export HYDRA_COMP_DEBUG=1
+export PATH=$PATH:.
+
+# Do actual test
+test=$(COMP_LINE='python my_app.py' hydra_bash_completion |\
+    grep EXECUTABLE_FIRST |\
+    awk '{split($0,a,"=");b=substr(a[2],2,length(a[2])-2);print b}')
+
+if [ $test == $PYTHON ]; then
```
omry, line 21: Not that it matters much here, but I would exit with non zero exit code if there is an error. This way you don't need to compare the output.

tests/test_completion.py
```diff
@@ -53,6 +53,21 @@ def create_config_loader() -> ConfigLoaderImpl:
...
+        process = subprocess.Popen(
+            "tests/scripts/test_bash_dot_in_path.bash",
+            shell=False,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        stdout, stderr = process.communicate()
+        assert stderr == b""
+        assert stdout == b"TRUE\n"
+        return
```
omry, line 68: if exit with an error, this entire thing becomes something like:

```python
subprocess.check_returncode("/bin/bash", "tests/scripts/test_bash_dot_in_path.bash") == 0
```

thoth291: Checking `stdout` could have done in the bash script itself - I agree - but checking for **anything** in `stderr` is something which I like to do in my tests to capture things which I haven't planned for.
I'm also not doing any interactivity here - but may be in the future it will be needed and `popen.communicate` will be there waiting for me.
Both of the points above gives me a feeling that it's better to stay as it is - but if you want - I can recode it - just let me know.

omry: sure, we can go with this for this case.

for Python calls I have run_with_error() that specifically return the stderr.
I have yet  to encounter a case when I wanted to test both stdout and stderr.

Issue #873, opened 2020-08-13
command line is incorrectly removing whitespace in primitives


omry:
```commandline
$ python foo.py "+x=a b c"
```
results in:
```yaml
x: abc
```

because the grammar is ignoring whitespace outside of quoted strings and there is a concatenation phase for primitives.

omry:
cc @odelalleau

odelalleau:
I'll take it

omry:
can you assign to yourself?

odelalleau:
> can you assign to yourself?

No, looks like I don't have permission for that

shagunsodhani:
@odelalleau needs to be added to the repo. I dont recall the procedure for this but had the same problem when I started working on Hydra.

PR #880, opened 2020-08-14
Fix primitive ws

odelalleau:
## Motivation

See #873 + our discussion

Yes

## Test Plan

`pytest`

## Related Issues and PRs

#873


hydra/core/override_parser/overrides_visitor.py
```diff
@@ -83,6 +83,11 @@ def visitPrimitive(
         last_idx = ctx.getChildCount()
         # skip first if whitespace
         if self.is_ws(ctx.getChild(0)):
+            if last_idx == 1:
```
omry, line 86: `key=` is supported as a way to set key to the empty string.
not sure I understand the reason for this exception.

hydra/grammar/OverrideLexer.g4
```diff
@@ -41,12 +63,10 @@ BOOL:
 
 NULL: [Nn][Uu][Ll][Ll];
 
-fragment CHAR: [a-zA-Z];
-ID : (CHAR|'_') (CHAR|DIGIT|'_')*;
-fragment LIST_INDEX: '0' | [1-9][0-9]*;
-DOT_PATH: (ID | LIST_INDEX) ('.' (ID | LIST_INDEX))+;
-
-WS: (' ' | '\t')+ -> channel(HIDDEN);
+OTHER_CHAR: [/\-\\+.$*];  // other characters allowed in unquoted strings
```
omry, line 66:
```suggestion
UNQUOTED_CHAR: [/\-\\+.$*];  // other characters allowed in unquoted strings
```

hydra/core/override_parser/overrides_parser.py
```diff
@@ -28,6 +28,21 @@
     )
     sys.exit(1)
 
+# Mappings between lexer modes and parser rules.
+MODE_TO_RULES = {
+    "KEY": ["key", "override", "package", "packageOrGroup"],
+    "ARGS": [
```
omry, line 34: ARGS does not seem to correctly describe the mode.
Maybe KEY and NORMAL ?

hydra/core/override_parser/overrides_visitor.py
```diff
@@ -144,18 +149,18 @@ def visitDictValue(
     ) -> Dict[str, ParsedElementType]:
         ret = {}
         children = ctx.getChildren()
-        assert self.is_matching_terminal(next(children), "{")
+        assert self.is_matching_terminal(next(children), OverrideLexer.BRACE_OPEN)
```
omry, line 152: this is leaking to error messages:
```
$ python ~/dev/hydra/examples/tutorials/basic/your_first_hydra_app/5_defaults/my_app.py key
Error parsing override 'key'
missing '=' at '<EOF>'
```

By using constants here, I think it would become:
```
$ python ~/dev/hydra/examples/tutorials/basic/your_first_hydra_app/5_defaults/my_app.py key
Error parsing override 'key'
missing EQUAL at '<EOF>'
```
Which is my opinion is more cryptic.
This is not a huge deal, but if possible I would rather keep the old error.

hydra/grammar/OverrideLexer.g4
```diff
@@ -4,28 +4,50 @@
...
-BRACKET_CLOSE: ']';
-BRACE_OPEN: '{';
-BRACE_CLOSE: '}';
+
+fragment CHAR: [a-zA-Z];
+ID: (CHAR|'_') (CHAR|DIGIT|'_')*;
+fragment LIST_INDEX: '0' | [1-9][0-9]*;
+DOT_PATH: (ID | LIST_INDEX) ('.' (ID | LIST_INDEX))+;
+
+WS: WS_ -> skip;
```
omry, line 32: what is the use case of skipping whitespace in KEY mode?

tests/test_overrides_parser.py
```diff
@@ -1493,19 +1494,19 @@ def error(msg: str) -> Any:
             "interval(1.0, 2.0)",
             CastResults(
                 int=CastResults.error(
-                    "ValueError while evaluating 'int(interval(1.0,2.0))':"
+                    "ValueError while evaluating 'int(interval(1.0, 2.0))':"
```
omry, line 1497: 1. these test changes are not about interval.
2. they are not comprehensive.

I think it's better to add a section to function testing that adds whitespace in all sorts of places.
```
foo( )
foo ()
foo(a )
foo( a)
foo(a, b)
foo(a ,b)
```

etc

hydra/grammar/OverrideLexer.g4
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+// Regenerate lexer by running 'python setup.py antlr' at project root.
+// If you make changes here be sure to update the documentation (and update the grammar in command_line_syntax.md)
```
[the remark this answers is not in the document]
odelalleau: Updated in 8023c495

hydra/grammar/OverrideParser.g4
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+// Regenerate parser by running 'python setup.py antlr' at project root.
+// If you make changes here be sure to update the documentation (and update the grammar in command_line_syntax.md)
```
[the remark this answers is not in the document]
odelalleau: Updated in 8023c495

> Speaking of which, did document anything?

Oh, that's a very good point, I assumed no update of the doc was needed since it was a bug fix, but I forgot the grammar was copied into the doc. I added the doc update to 8023c495

I wanted to check with you before documenting the whitespace behavior though: right now there is nothing in the doc about unquoted strings. It is written in such a way that it seems like if it's not quoted, then it needs to be a single primitive (as if the grammar didn't have this `+` in the primitive rule). Do you want me to add a subsection on unquoted strings?

hydra/core/override_parser/overrides_visitor.py
odelalleau: `key=` still works (there is no WS in that case). In normal use cases it should not be possible to have a primitive made only of whitespaces (because all those whitespaces should have been "consumed" into another lexer token). Also there was a test about this that was expecting a crash. It's the following parameterization of `test_parse_errors` (from current master):

```python
        pytest.param(
            "primitive",
            " ",
            pytest.raises(
                HydraException, match=re.escape("mismatched input '<EOF>' expecting"),
            ),
            id="error:value:whitespace",
        ),
```

hydra/grammar/OverrideLexer.g4
odelalleau: Renamed in 83792a8b

hydra/core/override_parser/overrides_parser.py
odelalleau: Right, this came from OmegaConf (where this mode is used for resolver arguments). What about VALUE? (NORMAL sounds a bit vague)

hydra/core/override_parser/overrides_visitor.py
odelalleau: You are right, though this is not related to this piece of code in the example you gave (that one is triggered by antlr, before even calling the visitor). It looks like it should be possible to customize antlr error messages by subclassing the parser. I'd rather do it in a subsequent PR though if that's ok (to avoid inflating this PR too much).

hydra/grammar/OverrideLexer.g4
odelalleau: Backward compatibility with current grammar in master. But it could definitely be removed if you prefer.

tests/test_overrides_parser.py
odelalleau: The extra whitespace in the error message is just a side-effect of not throwing whitespace to the hidden channel anymore, so `getText()` still contains them. Obviously these tests are not about testing whitespace handling.

I haven't looked closely at all existing tests, and it's indeed possible that there weren't enough tests about whitespaces before. I only added a few ones in ab52f476 to highlight the failure case of #873, which is what this PR was targeted at. But I'll see about adding more as you suggested.

odelalleau: > But I'll see about adding more as you suggested.

I added more in 81390659

hydra/core/override_parser/overrides_parser.py
omry: sure, we can go with KEY and VALUE.

hydra/core/override_parser/overrides_visitor.py
omry: roger.

hydra/core/override_parser/overrides_visitor.py
omry: Attempt in a followup PR is fine.

hydra/grammar/OverrideLexer.g4
omry: If there are no tests that ensuring it we can safely ignore it. I can't think of a use case for whitespace in key right now.

tests/test_overrides_parser.py
omry: Oh, I missed the meaning of the change.
Yes, it does make sense.

> (side note: I had to rebase to fix CI tests)

Let me guess, deprecations from cfg.pretty()?

website/docs/advanced/override_grammar/basic.md
```diff
@@ -27,25 +27,25 @@ The rest are manipulating the config object.
 ## Grammar
 Hydra supports a rich [DSL](https://en.wikipedia.org/wiki/Domain-specific_language) in the command line.   
 Below are the parser rules from grammar.
-You can see the full grammar [here](https://github.com/facebookresearch/hydra/tree/master/hydra/grammar/Override.g4).
+You can see the full grammar on GitHub
+([lexer](https://github.com/facebookresearch/hydra/tree/master/hydra/grammar/OverrideLexer.g4) and
+[parser](https://github.com/facebookresearch/hydra/tree/master/hydra/grammar/OverrideParser.g4))
 
 ```antlr4
```
omry, line 34:
```suggestion
```antlr4 title="OverrideParser.g4"
```

website/docs/advanced/override_grammar/basic.md
```diff
@@ -56,33 +56,36 @@ element:
     | function
 ;
 
-argName: ID '=';
-function: ID '('
-    (argName? element (',' argName? element )* )?
-')';
+argName: ID EQUAL;
+function: ID PARENTHESIS_OPEN
```
omry, line 60: POPEN, PCLOSE. this is too long :)

tests/test_overrides_parser.py
odelalleau: > Let me guess, deprecations from cfg.pretty()?

Good guess!

website/docs/advanced/override_grammar/basic.md
odelalleau: Committing your suggestion

hydra/core/override_parser/overrides_parser.py
odelalleau: Done in 19f11b09

hydra/grammar/OverrideLexer.g4
odelalleau: > If there are no tests that ensuring it we can safely ignore it. I can't think of a use case for whitespace in key right now.

There were no tests but I just added some earlier ;) I changed it to forbid whitespaces in KEY mode (and changed these tests accordingly) => this is in 977bca69

website/docs/advanced/override_grammar/basic.md
odelalleau: Done in 047b7c23

hydra/core/override_parser/overrides_parser.py
```diff
@@ -28,6 +28,21 @@
     )
     sys.exit(1)
 
+# Mappings between lexer modes and parser rules.
+MODE_TO_RULES = {
+    "KEY": ["key", "override", "package", "packageOrGroup"],
+    "VALUE": [
```
omry, line 34: would be simpler to go with KEY for the small subset of rules that are in the KEY mode and default to VALUE for the rest.

tests/test_overrides_parser.py
```diff
@@ -1812,3 +1816,95 @@ def test_glob_filter(
...
     strings = ["the", "quick", "brown", "fox", "jumped", "under", "the", "lazy", "dog"]
     assert Glob(include=include, exclude=exclude).filter(strings) == expected
+
+
+@pytest.mark.parametrize(  # type: ignore
+    "prefix,override_type",
+    [
+        pytest.param("", OverrideType.CHANGE, id="change"),
+        pytest.param("+", OverrideType.ADD, id="add"),
+        pytest.param("~", OverrideType.DEL, id="del"),
```
omry, line 1826: we don't really need to rest everything in those 3 prefixes.
I would add explicit testing for + and ~ though, with and without a space between them and the key.

website/docs/advanced/override_grammar/basic.md
```diff
@@ -27,25 +27,25 @@ The rest are manipulating the config object.
 ## Grammar
 Hydra supports a rich [DSL](https://en.wikipedia.org/wiki/Domain-specific_language) in the command line.   
 Below are the parser rules from grammar.
-You can see the full grammar [here](https://github.com/facebookresearch/hydra/tree/master/hydra/grammar/Override.g4).
-
-```antlr4
-grammar Override;
+You can see the full grammar on GitHub
+([lexer](https://github.com/facebookresearch/hydra/tree/master/hydra/grammar/OverrideLexer.g4) and
```
omry, line 31: I think I didn't answer your question:

Yes, please add a paragraph about whitespace being supported in values and a mention that it has to be quoted for the shell.

tests/test_overrides_parser.py
odelalleau: > I would add explicit testing for + and ~ though, with and without a space between them and the key.

Just to be sure, do you mean tests verifying that with a space there is a crash? (whitespaces aren't allowed anymore in KEY mode due to 977bca69471a4e1229479ca412bc65d74bb58429)

website/docs/advanced/override_grammar/basic.md
odelalleau: I added a sub-section on unquoted strings with shell examples in 6434fc6206726c6a51494e0607e0546c387eeeda, let me know if that looks good to you.

hydra/core/override_parser/overrides_parser.py
odelalleau: Simplified in 13269391

website/docs/advanced/override_grammar/basic.md
```diff
@@ -126,7 +126,22 @@ To include a single quote in a single quoted string escape it : `\'`. Same for d
 </div>
 </div>
 
- 
+### Unquoted strings
```
omry, line 129:
```suggestion
### Unquoted values
```

website/docs/advanced/override_grammar/basic.md
```diff
@@ -126,7 +126,22 @@ To include a single quote in a single quoted string escape it : `\'`. Same for d
...
+Finally, interpolations may be embedded within an unquoted string.
+
+Here are some examples showing how these may be used on the command line (note that here all
+quotes are "consumed" by the shell, so that from Hydra's point of view all these strings are unquoted):
+```bash
+python my_app.py url=https://hydra.cc
+python my_app.py url='https://${server}:${env:PORT}'
+python my_app.py msg='hello world'
+python my_app.py msg='  hello world  '  # same as above
+python my_app.py msg='hello^world'      # invalid, use a quoted '"hello^world"' instead
```
omry, line 142: I would prefer to train people to quote the whole override when quoting for the shell.
```suggestion
Override values can contain whitespaces and other special characters.
Some examples of legal values: `hello world`,`   hello world  `, `https://${server}:${env:PORT}`.
Unquoted whitespaces are allowed in the middle of the word, while leading and trailing whitespaces are stripped.
Normally, your shell will interpret values with whitesspaces as being multiple parameters (`key=a b` would be interpreted as `key=a` and `b`). To prevent this you can quote them with a single quote. For example:

```bash
$ python my_app.py 'msg=hello world'
```

website/docs/advanced/override_grammar/basic.md
odelalleau: Taking your suggested commit

website/docs/advanced/override_grammar/basic.md
odelalleau: Taking your suggested commit -- will just follow-up with a small typo fix in it

website/docs/advanced/override_grammar/basic.md
```diff
@@ -125,7 +126,17 @@ To include a single quote in a single quoted string escape it : `\'`. Same for d
...
+Override values can contain whitespaces and other special characters.
+Some examples of legal values: `hello world`,`   hello world  `, `https://${server}:${env:PORT}`.
+Unquoted whitespaces are allowed in the middle of the word, while leading and trailing whitespaces are stripped.
+Normally, your shell will interpret values with whitespaces as being multiple parameters (`key=a b` would be interpreted as `key=a` and `b`).
+To prevent this you can quote them with a single quote. For example:
+
+```bash
+$ python my_app.py 'msg=hello world'
+```
+
```
omry, line 139: What do you think about (not suggestion because GitHub does not render it correctly).

### Whitespaces in unquoted values
Unquoted Override values can contain non leading or trailing whitespaces.
For example, `msg=hello world` is a legal override (key is `msg` and value is the string `hello world`).
Normally, your shell will interpret values with whitespaces as being multiple parameters (`key=a b` would be interpreted as `key=a` and `b`).
To prevent this you can quote them with a single quote. For example:

```bash
$ python my_app.py 'msg=hello world'
```

Note that trailing and leading whitespace are ignored, the above is equivalent to:
```bash
$ python my_app.py 'msg=    hello world    '
```

odelalleau: Looks good to me, I pushed it in b54615cc

omry:
Before I land, can you organize the stack and force push?
I don't want to push 18 commits but I still think it makes sense to break it into multiple commits.

tests/test_overrides_parser.py
odelalleau: This point hasn't been resolved yet -- I wasn't sure exactly what you meant

odelalleau:
> Before I land, can you organize the stack and force push?
> I don't want to push 18 commits but I still think it makes sense to break it into multiple commits.

Done -- there remains just this point to clarify: https://github.com/facebookresearch/hydra/pull/880#discussion_r474214530

tests/test_overrides_parser.py
omry: Missed your question.
I mean that I think we don't need dimension of "prefix,override_type", we can just drop it and hard code for normal override.
This is tripling the number of tests here and I don't think it's relevant for what you are actually testing.

omry:
> > Before I land, can you organize the stack and force push?
> > I don't want to push 18 commits but I still think it makes sense to break it into multiple commits.
> 
> Done -- there remains just this point to clarify: [#880 (comment)](https://github.com/facebookresearch/hydra/pull/880#discussion_r474214530)

responded to that one.
can you squash the two doc changes?

tests/test_overrides_parser.py
odelalleau: Got it, thanks! Done and rebased (now in e35815bc)

odelalleau:
> can you squash the two doc changes into one commit?

Done in 29ce0ccd

PR #894, opened 2020-08-20
[nevergrad plugin] Fix integer/float choice from spec

### LINKED ISSUE #881: [Bug]: Nevergrad sweeper does not work with integers if there are less than 6 choices
Nevergrad sweeper complains if it has less than 6 values to sweep over in a range (e.g. `lower: 1` `upper:3`) and asks to use a list instead (`ValueError: For integers with 6 or fewer values, use a choice instead`). But if you use a list with integers it does not work because it assumes that choices contain only strings:

Screen Shot 2020-08-15 at 7 42 39 PM

Line where the first error is raised: https://github.com/facebookresearch/hydra/blob/0e001afb2a55275b6f7dc33e79035dbf3a797c00/plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py#L178
Hydra Version: 1.0.0rc2

Of course I can give a string and then convert in my code, but it would probably be better to solve it differently? For example sliently treating it as a list without raising the first error ? Or at least to say in the raised error that you have to use a list and convert the str to int in your own code ? Not sure what is the best way..

jrapin:
## Motivation

Fixes the handling of integer/float choice from spec

Yes

## Test Plan

Tests have been added to repro the issue in #881 

## Related Issues and PRs

Closes #881 


omry:
If there isn't one already, please open an issue describing the problem this is solving.
Add a news fragment file to the PR with a one-line description of the fix. (Also link the issues with "Closing #XYZ" in the description of this PR).

jrapin:
> If there isn't one already, please open an issue describing the problem this is solving.
> Add a news fragment file to the PR with a one-line description of the fix. (Also link the issues with "Closing #XYZ" in the description of this PR).

Are you sure you mean "closing" and not "closes"? The former does nothing special in my understanding, while the latter is a special word and closes the issue when the PR is merged

omry:
You are probably right. my bad.

news/881.bugfix
```diff
@@ -0,0 +1 @@
+Fix integer/float choice from spec in nevergrad plugin
```
omry, line 1: I think case we should probably use [plugin](https://hydra.cc/docs/next/development/documentation) file suffix.

Issue #871, opened 2020-08-12
Update the Nevergrad Plugin to use the new OverridesParser 


omry:
Followup for #752.
Hydra 1.0.0rc3 introduces new command line grammar that supports advanced sweeprs.
You can read about it [here](http://localhost:3000/docs/next/advanced/override_grammar/extended).

The Nevergrad Sweeper should use hydra.core.override_parser.overrides_parser.OverridesParser to parse the command line to take advantage of new grammar.

You can see simple usage example in the [Example Sweeper Plugin ](https://github.com/facebookresearch/hydra/blob/master/plugins/examples/example_sweeper_plugin/hydra_plugins/example_sweeper_plugin/example_sweeper.py#L73-L94) and in the [Basic Sweeper Plugin](https://github.com/facebookresearch/hydra/blob/master/hydra/_internal/core_plugins/basic_sweeper.py#L107-L121).

omry:
cc @shagunsodhani , @jrapin , @jieru-hu : Anyone wants this one?

jieru-hu:
> cc @shagunsodhani , @jrapin , @jieru-hu : Anyone wants this one?

i'll take this :)

omry:
Not blocking Hydra 1.0

jieru-hu:
the plugin is released, closing :)

PR #908, opened 2020-08-22
Move nevergrad to the new parser

jieru-hu:
*update 09/01*
Todo list
- [x] address comments
- [x] rename `core.py` to `_core.py`
- [x] update doc
- [ ] new plugin release (bump version to 1.1.0rc1 because this is a breaking change).
- [ ] Update docs in 1.0 snapshot after plugin release (website/versioned_docs/...)

---
https://github.com/facebookresearch/hydra/issues/871

## Motivation

Yes/No

## Test Plan

## Related Issues and PRs


jieru-hu:
@omry 

For the nevergrad sweeper, one of the exisitng assumption is if there's no casting then we by default assume [the input is float](https://github.com/facebookresearch/hydra/blob/fc2cae79f96be17db58f7ad7cbe2a6333c878ed8/plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py#L36)

The casting information doesn't seem to be available with override parse, so in this case if the input is 

`range(1, 12)`, we will return `ng.p.Scalar, int` instead of `ng.p.Scalar, float`, I'm wondering if that's OK or we should figure out from the input to see if it has been cast or not?

jieru-hu:
@omry I thought I'd get your feedback before I move on to updating the doc :)

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
```diff
@@ -23,25 +25,22 @@ def test_discovery() -> None:
...
-        ("0,1,2", ng.p.TransitionChoice, float),
-        ([0, 1, 2], ng.p.TransitionChoice, float),
-        ("int:0,1,2", ng.p.TransitionChoice, int),
-        ("str:0,1,2", ng.p.Choice, str),
-        ("0.0,12.0,2.0", ng.p.Choice, float),
-        ("int:1:12", ng.p.Scalar, int),
-        ("1:12", ng.p.Scalar, float),
-        ("log:0.01:1.0", ng.p.Log, float),
-        ("blublu", str, str),
+        (get_choice_sweep(["blu", "blublu"]), ng.p.Choice, str),
```
omry, line 30: why not this?
```
 (choice(blue,blublu), ng.p.Choice, str),
```

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
```diff
@@ -23,25 +25,22 @@ def test_discovery() -> None:
 
 
 @pytest.mark.parametrize(  # type: ignore
-    "description,param_cls,value_cls",
+    "override,param_cls,value_cls",
```
omry, line 28: this function tested the conversion from a string to a resulting parameter for Nevergrad.
Don't change the meaning of the test. input should still be a string.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -24,43 +25,37 @@ class CommandlineSpec:
 
     Attributes
     ----------
-    bounds: Optional[Tuple[float, float]]
+    bounds: Optional[Tuple[Any, Any]]
```
omry, line 28: CommandlineSpec should be deleted in my opinion. There is no need for it. You should convert from Override directly to the target.
See the the PR converting Ax for inspiration.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -77,42 +72,32 @@ def parse(cls, string: str) -> "CommandlineSpec":
...
+            return cls(bounds=(value.start, value.stop), log=("log" in tags), step=step)
+        elif override.value_type == ValueType.ELEMENT:
+            return override.value()
+        else:
+            raise ValueError(f"Unsupported Override: {override}")
 
 
 # pylint: disable=too-many-branches
 def make_nevergrad_parameter(description: Any) -> Any:
     """Returns a Nevergrad parameter from a definition string or object.
```
omry, line 100: The input for this function should be an Override object.
It should probably be re-written from scratch.

plugins/hydra_nevergrad_sweeper/tests/__init__.py
```diff
@@ -1 +1,37 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+from typing import Any, List
+
+from hydra.core.override_parser.types import (
+    ChoiceSweep,
+    Override,
+    OverrideType,
+    RangeSweep,
```
omry, line 8: Looks like you did not use interval()
This make me wonder how your tests are even working, interval() is critical for the Nevergrad sweeper.

plugins/hydra_nevergrad_sweeper/tests/__init__.py
```diff
@@ -1 +1,37 @@
...
+    )
+
+
+def get_override_element(val: Any) -> Override:
+    return Override(
+        type=OverrideType.CHANGE,
+        key_or_group="test",
+        value_type=ValueType.ELEMENT,
+        _value=val,
+    )
```
omry, line 37: Why do you need to construct Override?!

plugins/hydra_nevergrad_sweeper/tests/__init__.py
jieru-hu: according to the [nevergrad doc](https://hydra.cc/docs/next/plugins/nevergrad_sweeper/), [these params](https://github.com/facebookresearch/hydra/blob/master/plugins/hydra_nevergrad_sweeper/example/config.yaml#L16-L37) are all `ranges`, they are all translated into the same Nevergrad `Scalar` object.  I think we should be consistent with Nevergrad in this case.

omry: These are he docs for this plugin, not Nevergrad's docs. They will need to be changed. This change is a dramatic change in API for this plugin.

Please pause work on this until we talk. I think we have very different ideas of what this task is about.

jieru-hu: Cool, I will dig deeper into nevergrad's doc then.

omry: The plugin is creating an interface to Nevergrad.
That interface is using the Hydra grammar, and should be consistent with that, not with the underlying implementation.
This is a substantial shift, before each plugin was free to define it's own command line semantics. this is not the case anymore. No both optimizer plugins, Nevergrad and Ax need to use the command line parsing provided by Hydra, and they should do it in a consistent way to allow users to easily switch to the optimizer that works best for them.

The plugins are responsible for translating from the language Hydra is using to express the desire of the user to the language used by the underlying framework (Ax, Nevergrad etc).

jieru-hu: I see, yea, that makes sense! Thx

jieru-hu: @omry, I read the nevergrad doc more thoroughly and have created the following mapping, does it look more reasonable now?  
```
# discrete
   choice(1,2,3)  ==>  ng.p.Choice([["Helium", "Nitrogen", "Oxygen"]])
   tag(transition,choice(2,3,4))  ==> ng.p.TransitionChoice(["Solid", "Liquid", "Gas"])
   tag(transition, range(1, 11, 2)) ==> ng.p.TransitionChoice([1, 3, 5 ...])
   range(1,5) ==>  ng.p.Scalar(lower=1, upper=12).set_integer_casting(),
# continuous
   interval(1.0,5.0) ==> np.p.Scalar(lower=1.0, upper=5.0)
   tag(log, interval(0.1, 1)) ==> np.p.Log(lower=0.1, upper=1)
```

omry: Looking better.

Looking at the existing tests:
```
        ("blu,blublu", ng.p.Choice, str),
        (["blu", "blublu"], ng.p.Choice, str),
        ("0,1,2", ng.p.TransitionChoice, float),
        ([0, 1, 2], ng.p.TransitionChoice, float),
        ("int:0,1,2", ng.p.TransitionChoice, int),
        ("str:0,1,2", ng.p.Choice, str),
        ("0.0,12.0,2.0", ng.p.Choice, float),
        ("int:1:12", ng.p.Scalar, int),
        ("1:12", ng.p.Scalar, float),
        ("log:0.01:1.0", ng.p.Log, float),
        ("blublu", str, str),
```

It looks like any numeric choice is considered transitional, and string choices are considered categorical (normal choice).
I think you should not introduce a new tag for this. we can keep the same semantics.

jieru-hu: I think the tests are a bit misleading here:

According to the nevergrad doc here https://facebookresearch.github.io/nevergrad/machinelearning.html?highlight=transitionchoice
`arg1 = ng.p.TransitionChoice(["a", "b"])  # 1st arg. = positional discrete argument`

Strings can also be transitional.

omry: What does transitional actually mean? that there is an initial guess in the center of the range?

jieru-hu: Choice and TransitionChoice behaves differently: 

[Choice](https://facebookresearch.github.io/nevergrad/parametrization_ref.html?highlight=transition#nevergrad.p.Choice) draws values randomely

[TransitionChoice](https://facebookresearch.github.io/nevergrad/parametrization_ref.html?highlight=transition#nevergrad.p.TransitionChoice): Ordered categorical parameter, The chosen parameter is drawn using transitions between current choice and the next/previous ones. Although I'm not sure if it's common to sweep non-numerical  ordered values, but it is something Nevergrad supports.

omry: Ax also support choices. I am not sure if they expose control over the sampling method.
@jrapin, does it make sense to you to use a tag to indicate transitional vs a regular choice?

```python
choice(a,b,c)  =>  ng.p.Choice(["a", "b", "c"])
tag(transition, choice(a,b,c)) ==> ng.p.TransitionChoice(["a","b","c"])
```

jrapin: To give full(er) control on the parametrization yes, but is it really useful for the plugin, not so sure.
If you are at the point where you want absolute control, better use nevergrad directly.
So I'd probably stick to `TransitionChoice` for ordered lists, and `Choice` for anything else.
If that becomes necessary, we can always add tag support later on to override the default behavior?

> I think the tests are a bit misleading here

Indeed, I should have added other tests, `[3, 2, 1]` is considered as unordered in the plugin since it is not strictly increasing.

jrapin:
I was not aware of this PR until today (feel free to ping through workchat for questions!) and I won't have until next Monday :s (PTO, so no need to wait for me) but I like what I am seeing in the comments ;)

omry:
This is what I am thinking after reading the comments:

```python
# regardless of the choice values, choice is always unordered p.Choice:
choice(a,b,c)  =>  ng.p.Choice(["a", "b", "c"]) 

# We can support forcing a choice to be ordered by tagging it:
# (I prefer order over transition(al). at least to me it's clearer.
tag(ordered, choice(a,b,c)) ==> ng.p.TransitionChoice(["a","b","c"])

# ranges are always ordered:
range(1,5)  =>  ng.p.TransitionChoice([1,2,3,4])

# unless they are shuffled (There is a flag shuffle in the RangeSweep object.
shuffle(range(1,5)) ->  =>  ng.p.Choice([4,2,1,3]) 

# intervals are scalars:
# Note: by intervals are always interpreted as floats (even for int start and end values).
interval(0,1)     -> RangeSweep(start=0.0, end=1.0) -> ng.p.Scalar(lower=0.0, upper=1.0)
interval(0.0,1.0) -> RangeSweep(start=0.0, end=1.0) -> ng.p.Scalar(lower=0.0, upper=1.0)

# a user can cast the interval to int to override that:
int(interval(0,1) -> RangeSweep(start=0, end=1) -> ng.p.Scalar(lower=0.0, upper=1.0).set_integer_casting()
```

If everyone think this makes sense we can go ahead with it.

omry:
> I was not aware of this PR until today (feel free to ping through workchat when I don't react or for any question!) and I won't have until next Monday :s (PTO, so no need to wait for me) but I like what I am seeing in the comments ;)

I did not want to bother you with it because you did not engage with the extended grammar design doc or with #871.

Enjoy your PTO!

plugins/hydra_nevergrad_sweeper/tests/__init__.py
jieru-hu: > This is what I am thinking after reading the comments:
> 
> ```python
> # regardless of the choice values, choice is always unordered p.Choice:
> choice(a,b,c)  =>  ng.p.Choice(["a", "b", "c"]) 
> 
> # We can support forcing a choice to be ordered by tagging it:
> # (I prefer order over transition(al). at least to me it's clearer.
> tag(ordered, choice(a,b,c)) ==> ng.p.TransitionChoice(["a","b","c"])
> 
> # ranges are always ordered:
> range(1,5)  =>  ng.p.TransitionChoice([1,2,3,4])
> 
> # unless they are shuffled (There is a flag shuffle in the RangeSweep object.
> shuffle(range(1,5)) ->  =>  ng.p.Choice([4,2,1,3]) 
> 
> # intervals are scalars:
> # Note: by intervals are always interpreted as floats (even for int start and end values).
> interval(0,1)     -> RangeSweep(start=0.0, end=1.0) -> ng.p.Scalar(lower=0.0, upper=1.0)
> interval(0.0,1.0) -> RangeSweep(start=0.0, end=1.0) -> ng.p.Scalar(lower=0.0, upper=1.0)
> 
> # a user can cast the interval to int to override that:
> int(interval(0,1) -> RangeSweep(start=0, end=1) -> ng.p.Scalar(lower=0.0, upper=1.0).set_integer_casting()
> ```
> 
> If everyone think this makes sense we can go ahead with it.

lgtm. I will go ahead with the implementation then.


Thx @jrapin  for taking a look! Hope you enjoy your time off 🙃

jieru-hu:
> This is what I am thinking after reading the comments:
> 
> ```python
> # regardless of the choice values, choice is always unordered p.Choice:
> choice(a,b,c)  =>  ng.p.Choice(["a", "b", "c"]) 
> 
> # We can support forcing a choice to be ordered by tagging it:
> # (I prefer order over transition(al). at least to me it's clearer.
> tag(ordered, choice(a,b,c)) ==> ng.p.TransitionChoice(["a","b","c"])
> 
> # ranges are always ordered:
> range(1,5)  =>  ng.p.TransitionChoice([1,2,3,4])
> 
> # unless they are shuffled (There is a flag shuffle in the RangeSweep object.
> shuffle(range(1,5)) ->  =>  ng.p.Choice([4,2,1,3]) 
> 
> # intervals are scalars:
> # Note: by intervals are always interpreted as floats (even for int start and end values).
> interval(0,1)     -> RangeSweep(start=0.0, end=1.0) -> ng.p.Scalar(lower=0.0, upper=1.0)
> interval(0.0,1.0) -> RangeSweep(start=0.0, end=1.0) -> ng.p.Scalar(lower=0.0, upper=1.0)
> 
> # a user can cast the interval to int to override that:
> int(interval(0,1) -> RangeSweep(start=0, end=1) -> ng.p.Scalar(lower=0.0, upper=1.0).set_integer_casting()
> ```
> 
> If everyone think this makes sense we can go ahead with it.

lgtm, I will go ahead with the implementation then.

jieru-hu:
> I was not aware of this PR until today (feel free to ping through workchat when I don't react or for any question!) and I won't have until next Monday :s (PTO, so no need to wait for me) but I like what I am seeing in the comments ;)



Thx @jrapin  for taking a look! Hope you enjoy your time off 🙃

alshamiri5 (approved):

jieru-hu:
> # intervals are scalars:
> # Note: by intervals are always interpreted as floats (even for int start and end values).
> interval(0,1)     -> RangeSweep(start=0.0, end=1.0) -> ng.p.Scalar(lower=0.0, upper=1.0)
> interval(0.0,1.0) -> RangeSweep(start=0.0, end=1.0) -> ng.p.Scalar(lower=0.0, upper=1.0)


@omry 
This is what I got from my debugger, looks like the start and end are treated as `int` not `float`. 

Is it expected? (from your previous comments it sounds like Hydra would turn these into floats by default, but not sure if I misunderstood you comment.)
```
(Pdb) override
Override(type=<OverrideType.CHANGE: 1>, key_or_group='batch_size', value_type=<ValueType.INTERVAL_SWEEP: 6>, _value=IntervalSweep(tags=set(), start=1, end=5), pkg1=None, pkg2=None, input_line='batch_size=interval(1,5)', config_loader=None)

(Pdb) type(override.value().start)
<class 'int'>


```

omry:
are you rebased?

jieru-hu:
> are you rebased?

nvm, my bad, i did rebase but didn't install hydra from source again

omry:
> > are you rebased?
> 
> nvm, my bad, i did rebase but didn't install hydra from source again

for most changes (this one included I think), installing in editable/dev mode (`pip install -e .`) should not require a re-installation after a rebase.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -1,10 +1,10 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
-import itertools
 import logging
```
jieru-hu, line 2: @omry 


In future PR/commits I want to change this file to `_core.py` to avoid import. but for now I will not rename so it's easier to review the diff.

jieru-hu:
@omry 

I didn't update the doc. I'd like to get another round of feedback from you to make sure we are aligned on what's expected before updating the doc.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -17,149 +18,55 @@
...
+                scalar = ng.p.Scalar(**params)
+                scalar.set_mutation(sigma=val.step)
+                scalar.integer = isinstance(val.start, int)
+            elif override.is_interval_sweep():  # continuous variable
+                if "log" in val.tags:
+                    scalar = ng.p.Log(lower=val.start, upper=val.end)
+                else:
+                    scalar = ng.p.Scalar(lower=val.start, upper=val.end)
+                if isinstance(val.start, int):
+                    scalar.set_integer_casting()
```
jieru-hu, line 63: @omry 

I'm getting a lot of complaints from `mypy` here, I'm wondering if you have any suggestion on how to handle it more cleanly.


```
(pytest38) jieru-mbp:hydra jieru$ mypy --strict plugins/hydra_nevergrad_sweeper/hydra_plugins --config-file .mypy.ini 
plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py:48: error: Name 'val' already defined on line 46
plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py:49: error: Item "str" of "Union[str, int, float, bool, List[Any], Dict[str, Any], ChoiceSweep, RangeSweep, IntervalSweep, None]" has no attribute "tags"
plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py:49: error: Item "int" of "Union[str, int, float, bool, List[Any], Dict[str, Any], ChoiceSweep, RangeSweep, IntervalSweep, None]" has no attribute "tags"
plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py:49: error: Item "float" of "Union[str, int, float, bool, List[Any], Dict[str, Any], ChoiceSweep, RangeSweep, IntervalSweep, None]" has no attribute "tags"
plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py:49: error: Item "bool" of "Union[str, int, float, bool, List[Any], Dict[str, Any], ChoiceSweep, RangeSweep, IntervalSweep, None]" has no attribute "tags"
plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py:49: error: Item "List[Any]" of "Union[str, int, float, bool, List[Any], Dict[str, Any], ChoiceSweep, RangeSweep, IntervalSweep, None]" has no attribute "tags"
plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py:49: error: Item "Dict[str, Any]" of "Union[str, int, float, bool, List[Any], Dict[str, Any], ChoiceSweep, RangeSweep, IntervalSweep, None]" has no attribute "tags"
plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py:49: error: Item "None" of "Union[str, int, float, bool, List[Any], Dict[str, Any], ChoiceSweep, RangeSweep, IntervalSweep, None]" has no attribute "tags"
plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py:50: error: Item "str" of "Union[str, int, float, bool, List[Any], Dict[str, Any], ChoiceSweep, RangeSweep, IntervalSweep, None]" has no attribute "list"

```

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
omry: Ok.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -17,149 +18,55 @@
...
-                except ValueError:
-                    pass
-            return cls(options=options, cast=cast)
-        # bounded argument
-        bounds: Tuple[float, float] = tuple(float(x) for x in remain)  # type: ignore
-        cast = casts[0] if casts else "float"
-        return cls(bounds=bounds, cast=cast, log="log" in modifiers)
-
+    # We can support forcing a choice to be ordered by tagging it:
+    # (I prefer order over transition(al). at least to me it's clearer.
```
omry, line 28: what is this comment doing in the code?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
omry: not sure why.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -17,149 +18,55 @@
...
+    scalar = None
+    if isinstance(description, Override):
+        override = description
+        val = override.value()
+        if override.is_sweep_override():
+            if override.is_choice_sweep():
+                if "ordered" in val.tags:
+                    return ng.p.TransitionChoice(val.list)
+                else:
+                    return ng.p.Choice(val.list)
```
omry, line 51: 1. What happens on shuffle() and sort()?
2. What happens if the values are lists or dicts? (`[1,2],[3,4]`, `{foo:10},{bar:20}`)

Follow #945, I think it will solve these problems.

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
```diff
@@ -5,43 +5,71 @@
...
-        ("0,1,2", ng.p.TransitionChoice, float),
-        ([0, 1, 2], ng.p.TransitionChoice, float),
-        ("int:0,1,2", ng.p.TransitionChoice, int),
-        ("str:0,1,2", ng.p.Choice, str),
-        ("0.0,12.0,2.0", ng.p.Choice, float),
-        ("int:1:12", ng.p.Scalar, int),
-        ("1:12", ng.p.Scalar, float),
-        ("log:0.01:1.0", ng.p.Log, float),
-        ("blublu", str, str),
+        ("key=choice(1,2)", ng.p.Choice, int),
```
omry, line 52: can you change this to:
`"override,expected"`
e.g:
`key=choice(1,2)", ng.p.Choice(items=[1,2])`
Or however you represetnt a choice between 1 and 2 in Nevergrad?
going through the class here only makes the test more obscure, I think it's not adding any value.

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
```diff
@@ -5,43 +5,71 @@
...
+    assert NevergradSweeper.__name__ in [
         x.__name__ for x in Plugins.instance().discover(Sweeper)
     ]
 
 
 @pytest.mark.parametrize(  # type: ignore
-    "description,param_cls,value_cls",
+    "params,param_cls,value_cls",
+    [
+        ([1, 2, 3], ng.p.Choice, int),
```
omry, line 32: Instead of describing the expected value as the class and the type, can you just use an object?
```
([1, 2, 3], ng.p.Choice, int),
=>
 ([1, 2, 3], ng.p.Choice(values=[1,2,3], type=int) # or something like it.
```

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
```diff
@@ -82,10 +110,10 @@ def test_nevergrad_example(with_commandline: bool, tmpdir: Path) -> None:
...
     if with_commandline:
         cmd += [
-            "db=mnist,cifar",
-            "batch_size=4,8,12,16",
-            "lr=log:0.001:1.0",
-            "dropout=0:1",
+            "db=choice(mnist,cifar)",
+            "batch_size=choice(4,8,12,16)",
+            "lr=tag(log, interval(0.001, 1.0))",
+            "dropout=interval(0,1)",
```
omry, line 116: Keep it simple:
```suggestion
            "db=mnist,cifar",
            "batch_size=4,8,12,16",
            "lr=tag(log, interval(0.001, 1.0))",
            "dropout=interval(0,1)",
```

omry (changes requested):

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jieru-hu: ah I copy pasted it for myself for reference. will clean it up

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jieru-hu: For `shuffle` the Sweep class actually have a field indicating the status `RangeSweep(tags=set(), start=1, stop=10, step=1, shuffle=True)`, but for `sort` there's no corresponding field, should we add one for `sort` as well?

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
jieru-hu: sounds good!

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
jieru-hu: sounds good!

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
jieru-hu: sure, will update these

jieru-hu:
My doc changes was not reflected in https://deploy-preview-908--hydra-preview.netlify.app/docs/plugins/nevergrad_sweeper

I'm wondering if I missed something here.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jieru-hu: i needed to cast the vals in the corresponding sweeps :)

omry:
Your doc changes are probably in the next version.
If you want them to appear in 1.0 as well you need to add them in the versioned_doc files for 1.0.

omry:
Yes, it's in next:
image

A few comments:
1. Run your examples to make sure they run. this one will not work.
2. Try to avoid scroll bar in examples.

jieru-hu:
> Yes, it's in next:
> image
> 
> A few comments:
> 
> 1. Run your examples to make sure they run. this one will not work.
> 2. Try to avoid scroll bar in examples.

Curious why you say this one will not work (I ran it locally and it worked.), is it because I didnt inlcude the single quotes in the example?

```
$ python example/dummy_training.py -m db=mnist,cifar batch_size=4,8,16 lr='tag(log, interval(0.001, 1))' dropout='interval(0,1)'
```

jieru-hu:
> Your doc changes are probably in the next version.
> If you want them to appear in 1.0 as well you need to add them in the versioned_doc files for 1.0.

Ah I see, thx!

omry:
> (not urgent, i know you are off today :) )
> 
> Curious why you say this one will not work (I ran it locally and it worked.), is it because I didnt inlcude the single quotes in the example?
> 
> ```
> $ python example/dummy_training.py -m db=mnist,cifar batch_size=4,8,16 lr='tag(log, interval(0.001, 1))' dropout='interval(0,1)'
> ```

This one works. the one in the screen shot is missing quotes.
Also your quotes is inconsistent with the recommended style.
When quoting for the shell, quote the whole thing:
```
'lr=tag(log, interval(0.001, 1))'
$not
lr='tag(log, interval(0.001, 1))'

```

website/docs/plugins/nevergrad_sweeper.md
```diff
@@ -137,7 +135,29 @@ or a combination of both (eg.: `batch_size=log:int:4:1024`)
...
+
+You can tag an override with ```ordered``` to indicate it's a [```TransitionChoice```](https://facebookresearch.github.io/nevergrad/parametrization_ref.html#nevergrad.p.TransitionChoice)
+```commandline
+key=tag(ordered, choice(1,2,3))
+```
+
+#### Scalar
+```commandline
+key=range(1,8) # Scalar bounds cast to be integer
+key=float(range(1,8))  # Scalar bounds to be float
```
omry, line 160:
```suggestion
key=float(range(1,8))  # Scalar bounds cast to a float
```

website/docs/plugins/nevergrad_sweeper.md
```diff
@@ -137,7 +135,29 @@ or a combination of both (eg.: `batch_size=log:int:4:1024`)
...
+key=interval(1,12)
```
omry, line 161:
```suggestion
key=interval(1,12)   # Interval are float by default
```

omry (changes requested):

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -1,10 +1,17 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
-import itertools
 import logging
-from dataclasses import dataclass
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, cast
 
+import nevergrad as ng
```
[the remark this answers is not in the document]
jieru-hu: I plan on rename `core.py` to `_core.py` after the changes are reviewed (see TODOs in  description of the PR.) and my understanding is that that acts as an lazy import.

Didnt do that with this commit cuz i thought it'd  make the review harder. Thoughts?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -235,8 +128,6 @@ def setup(
         )
 
     def sweep(self, arguments: List[str]) -> None:
-        # lazy initialization to avoid overhead when loading hydra
-        import nevergrad as ng
```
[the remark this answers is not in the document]
jieru-hu: see above.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -217,7 +110,7 @@ def __init__(
         if parametrization is not None:
             assert isinstance(parametrization, DictConfig)
             self.parametrization = {
-                x: make_nevergrad_parameter(y) for x, y in parametrization.items()
+                x: get_nevergrad_parameter(y) for x, y in parametrization.items()
```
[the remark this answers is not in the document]
jieru-hu: sure will update this.

website/docs/plugins/nevergrad_sweeper.md
```diff
@@ -137,7 +135,29 @@ or a combination of both (eg.: `batch_size=log:int:4:1024`)
...
+```
+
+You can tag an override with ```ordered``` to indicate it's a [```TransitionChoice```](https://facebookresearch.github.io/nevergrad/parametrization_ref.html#nevergrad.p.TransitionChoice)
+```commandline
+key=tag(ordered, choice(1,2,3))
+```
+
+#### Scalar
+```commandline
+key=range(1,8) # Scalar bounds cast to be integer
```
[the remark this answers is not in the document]
jieru-hu: The cast happens when we transfer the override to ng params, I see why this is confusing, will fix the wording :)

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
```diff
@@ -6,45 +6,92 @@
...
-    if param_cls is not str:
-        assert isinstance(param.value, value_cls)
+    param = core.get_nevergrad_parameter(params)
+    verify_param(param, param_cls, param_val)
+
+
+@pytest.mark.parametrize(  # type: ignore
+    "override,param_cls, param_val",
+    [
+        ("key=choice(1,2)", ng.p.Choice, {1, 2}),
```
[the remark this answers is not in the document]
jieru-hu: As i was working on the tests I realized why the tests was written in that way originally, the tl;dr is I think this is more straightforward 

The object looks something like the following, with a few default values that we are not setting.
```
TransitionChoice(choices=Tuple(hello,world,nevergrad),positions=Array{Cd(0,3)},transitions=[1. 1.]):world
```

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -17,149 +24,52 @@
...
+        else:
+            return val
+    if isinstance(description, (list, ListConfig)):
+        return ng.p.Choice(list(description))
     if isinstance(description, (dict, DictConfig)):
         description = ScalarConfigSpec(**description)
-    if isinstance(description, ScalarConfigSpec):
         init = ["init", "lower", "upper"]
         init_params = {x: getattr(description, x) for x in init}
         if not description.log:
```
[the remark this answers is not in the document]
jieru-hu: yea, it is covered in `def test_create_nevergrad_parameter_from_config`

jieru-hu:
@omry 
I suspected some of the test failed because of in this version we are not lazy importing nevergrad. going to rename `core.py` to `_core.py` now (as mentioned earlier)

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -17,149 +25,52 @@
...
+                    and val.start is not None
+                    and val.stop - val.start <= 6
+                ):
+                    vals = [
+                        x
+                        for x in override.sweep_iterator(transformer=Transformer.encode)
+                    ]
+                    return ng.p.Choice(vals)
+                params = {"lower": val.start, "upper": val.stop}
+                scalar = ng.p.Scalar(**params)  # type: ignore
```
omry: Why not `Scalar(lower=val.start, upper=val.stop)` ?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -17,149 +25,52 @@
...
+                if "log" in val.tags:
+                    scalar = ng.p.Log(lower=val.start, upper=val.end)
+                else:
+                    scalar = ng.p.Scalar(lower=val.start, upper=val.end)
+                if isinstance(val.start, int):
+                    scalar.set_integer_casting()
+        else:
+            return val
+    if isinstance(description, (list, ListConfig)):
+        return ng.p.Choice(list(description))
```
omry: if you need to convert ListConfig to list use OmegaConf.to_container(description, resolve=True)

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/nevergrad_sweeper.py
```diff
@@ -0,0 +1,31 @@
...
+from .config import OptimConf
+
+
+class NevergradSweeper(Sweeper):
+    """Class to interface with the Ax Platform"""
+
+    def __init__(
+        self, optim: OptimConf, version: int, parametrization: Optional[DictConfig]
+    ):
+        from .core import CoreNevergradSweeper
```
omry: Okay, I think once you rename to _core (or _impl) it would not be imported during plugin discovery.

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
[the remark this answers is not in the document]
omry: I understand that this was the original implementation. There should be a good reason to delegate the initialization to the test function like this and I am just not seeing it.
Please move to explicit initialization, it will make things much easier to understand.

website/docs/plugins/nevergrad_sweeper.md
```diff
@@ -64,7 +64,9 @@ python example/dummy_training.py -m
 
 You can also override the search space parametrization:
 ```bash
-python example/dummy_training.py -m db=mnist,cifar batch_size=4,8,16 lr=log:0.001:1 dropout=0:1
+python example/dummy_training.py -m \
```
omry, line 67: can  you also rename dummy_training to my_app.py to be consistent with the rest of the examples?

website/docs/plugins/nevergrad_sweeper.md
```diff
@@ -137,7 +135,29 @@ or a combination of both (eg.: `batch_size=log:int:4:1024`)
     is multiplicative. 
   - `integer`: set to `true` for integers (favor floats over integers whenever possible)
 
-  Providing only `lower` and `upper` bound will set the initial value to the middle of the range, and the step to a sixth of the range.
-
+Providing only `lower` and `upper` bound will set the initial value to the middle of the range, and the step to a sixth of the range.
 **Note**: unbounded scalars (scalars with no upper and/or lower bounds) can only be defined through a config file.
 
+### Defining through commandline override
```
omry, line 141: I think the section about command line should come before the section about config files.
Most users will be defining sweeps through the command line.

omry (changes requested):

omry:
FYI, Hydra's master branch is going to switch to 1.1 soon.
If this does not land before you will need to land it to both the 1.0 branch and master if you want it in both.

jieru-hu:
> FYI, Hydra's master branch is going to switch to 1.1 soon.
> If this does not land before you will need to land it to both the 1.0 branch and master if you want it in both.

Thx @omry for the heads up!

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jieru-hu: got it, thx for the tip!

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -172,37 +83,20 @@ def make_nevergrad_parameter(description: Any) -> Any:
...
 
-    Parameters
-    ----------
-    config: DictConfig
-      the optimization process configuration
-    version: int
-      version of the API
-    """
 
+class CoreNevergradSweeper(Sweeper):
```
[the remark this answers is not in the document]
jieru-hu: sounds good!

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -172,37 +83,20 @@ def make_nevergrad_parameter(description: Any) -> Any:
...
             scalar.set_integer_casting()
+
+    if scalar:
+        if scalar.integer:
             a, b = scalar.bounds
             if a is not None and b is not None and b - a <= 6:
                 raise ValueError(
                     "For integers with 6 or fewer values, use a choice instead"
                 )
         return scalar
```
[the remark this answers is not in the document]
jieru-hu: I think this is a requirment that comes from [nevergrad](https://github.com/facebookresearch/nevergrad/blob/c042665b302a593cffa19288fc12b273eb4b5cc1/nevergrad/parametrization/data.py#L426-L468).

I could not find the requirements in public docs but maybe I just missed it.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -17,149 +25,52 @@
...
+                ):
+                    vals = [
+                        x
+                        for x in override.sweep_iterator(transformer=Transformer.encode)
+                    ]
+                    return ng.p.Choice(vals)
+                params = {"lower": val.start, "upper": val.stop}
+                scalar = ng.p.Scalar(**params)  # type: ignore
+                if isinstance(val.start, int):
+                    scalar.set_integer_casting()
```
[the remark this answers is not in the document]
jieru-hu: Initially I was trying to stay more consistent with the original impl... but i think your suggestion is better, it will also be easier for users to use: more straightforward and no surprieses.

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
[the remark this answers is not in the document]
jieru-hu: Agreed that explicitly init would make it easier to understand.

However as I pasted above there's a lot of extra fields in the object that we are not touching in the sweeper, we only have expectation for the values in the ng params. Also to explicitly init the object we'd need add dependency on `numpy`.

website/docs/plugins/nevergrad_sweeper.md
jieru-hu: sounds good, moved :)

website/docs/plugins/nevergrad_sweeper.md
jieru-hu: done :)

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -0,0 +1,186 @@
...
+)
+from hydra.core.plugins import Plugins
+from hydra.plugins.launcher import Launcher
+from hydra.plugins.sweeper import Sweeper
+from hydra.types import TaskFunction
+from omegaconf import DictConfig, ListConfig, OmegaConf
+
+from .config import OptimConf, ScalarConfigSpec
+
+# pylint: disable=logging-fstring-interpolation,no-self-used
```
omry, line 22: We are not using pylint with Hydra, why do you need to add this?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -0,0 +1,186 @@
...
+log = logging.getLogger(__name__)
+
+
+def verify_scalar_bounds(scalar: ng.p.Scalar) -> None:
+    if scalar.integer:
+        a, b = scalar.bounds
+        if a is not None and b is not None and b - a <= 6:
+            raise ValueError(
+                "For integers with 6 or fewer values, use a choice instead"
+            )
```
omry, line 32: Why is that?
Is this an actual requirement from Nevergrad? it seems arbitrary.

If so, would it make sense to do this automatically in the plugin instead of surfacing this error to the user?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def create_nevergrad_param_from_config(
+    config: Union[ListConfig, DictConfig, List[Any], Dict[str, Any], str]
```
omry, line 36: The function implementation is not doing anything if the type is Dict or List.
is this intentional?
can you just not call it in this case?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -0,0 +1,186 @@
...
+        return val
+    if override.is_choice_sweep():
+        val = cast(ChoiceSweep, val)
+        vals = [x for x in override.sweep_iterator(transformer=Transformer.encode)]
+        if "ordered" in val.tags:
+            return ng.p.TransitionChoice(vals)
+        else:
+            return ng.p.Choice(vals)
+    elif override.is_range_sweep():
+        # if shuffled or integer range and <=6, use Choice
```
omry, line 73: comment seems misplaced. You should either remove it or do what it's saying :)

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -0,0 +1,186 @@
...
+        vals = [x for x in override.sweep_iterator(transformer=Transformer.encode)]
+        return ng.p.Choice(vals)
+    elif override.is_interval_sweep():
+        val = cast(IntervalSweep, val)
```
omry, line 77: You can use:
```
assert isinstance(val, IntervalSweep)
```
instead of the cast.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -0,0 +1,186 @@
...
+        if "log" in val.tags:
+            scalar = ng.p.Log(lower=val.start, upper=val.end)
+        else:
+            scalar = ng.p.Scalar(lower=val.start, upper=val.end)  # type: ignore
+        if isinstance(val.start, int):
+            scalar.set_integer_casting()
+        verify_scalar_bounds(scalar)
```
omry, line 84: do you have tests covering overrides like these?
```
interval(0,10)       # float interval, 0.0 -> 10.0
int(interval(0,10))  # int interval, 0 -> 10
```

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -0,0 +1,186 @@
...
+        return scalar
+
+
+class NevergradSweeperImpl(Sweeper):
+    def __init__(
+        self, optim: OptimConf, version: int, parametrization: Optional[DictConfig]
```
omry, line 90: pro tip:
you can add a trailing comma and black will format the arguments as a list regardless of how long the line is.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -0,0 +1,186 @@
...
+    ):
+        assert (
+            version == 1
+        ), f"Only version 1 of API is currently available (got {version})"
```
omry, line 94: What is this achieving?

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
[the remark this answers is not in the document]
omry: Just initialize the exact same fields you are initializing now and let NG initialize the rest as defaults?
What you are doing is calling the `__init__` of say ng.p.Choice indirectly, passing multiple parameters.
It seems like it should be possible to do do the same explicitly, passing the same parameters.

Can you give an example of where this is not the case?

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
```diff
@@ -84,8 +140,8 @@ def test_nevergrad_example(with_commandline: bool, tmpdir: Path) -> None:
         cmd += [
             "db=mnist,cifar",
             "batch_size=4,8,12,16",
-            "lr=log:0.001:1.0",
-            "dropout=0:1",
+            "lr=tag(log, interval(0.001, 1.0))",
+            "dropout=interval(0,1)",
         ]
     subprocess.check_call(cmd)
     returns = OmegaConf.load(f"{tmpdir}/optimization_results.yaml")
```
omry, line 147: can you switch to using `get_run_output()` for this and other tests here?

website/docs/plugins/nevergrad_sweeper.md
```diff
@@ -59,12 +59,14 @@ batch_size=4
...
-python example/dummy_training.py -m
+python example/my_app.py.py -m
 ```
 
 You can also override the search space parametrization:
 ```bash
-python example/dummy_training.py -m db=mnist,cifar batch_size=4,8,16 lr=log:0.001:1 dropout=0:1
+python example/my_app.py.py -m \
+db=mnist,cifar batch_size=4,8,16 \
+'lr=tag(log, interval(0.001, 1))' 'dropout=interval(0,1)'
```
omry, line 69: formatting (and full --multirun instead of -m to make it clear).
```suggestion
python example/my_app.py.py --multirun db=mnist,cifar batch_size=4,8,16 \
'lr=tag(log, interval(0.001, 1))' 'dropout=interval(0,1)'
```
Also, can you add a word about quoting the parameters?

website/docs/plugins/nevergrad_sweeper.md
```diff
@@ -112,23 +114,42 @@ name: nevergrad
 
 ## Defining the parameters
 
-The plugin can use 2 types of parameters:
+The plugin supports 2 types of parameters: [Choices](https://facebookresearch.github.io/nevergrad/parametrization_ref.html#nevergrad.p.Choice) and [Scalars](https://facebookresearch.github.io/nevergrad/parametrization_ref.html#nevergrad.p.Scalar). They can be defined either through config file or commandline override.
 
-### Choices
+### Defining through commandline override
+Hydra 1.0 provides a override parser that support rich syntax. More documentation can be found in ([OverrideGrammer/Basic](../advanced/override_grammar/basic.md)) and ([OverrideGrammer/Extended](../advanced/override_grammar/extended.md)). We recommend you go through them first before proceeding with this doc.
```
omry, line 120:
```suggestion
Hydra provides a override parser that support rich syntax. More documentation can be found in ([OverrideGrammer/Basic](../advanced/override_grammar/basic.md)) and ([OverrideGrammer/Extended](../advanced/override_grammar/extended.md)). We recommend you go through them first before proceeding with this doc.
```

omry (changes requested):
Good job.
A few questions and some requests.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jieru-hu: nope, this was from previous code. thx for catching this!

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jieru-hu: I'm not sure what you mean by not doing anything, it actually wraps the the values in a ng param right?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jieru-hu: nice! I was doing the cast so mypy would stop complaining but this works and makes more sense!

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jieru-hu: yes, they are covered. :)

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jieru-hu: from reading the file history and the code, it was used to ping the command line API created for parsing the override for ng. I don't think this is needed any more. will remove.

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
jieru-hu: done! thx

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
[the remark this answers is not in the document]
jieru-hu: You are right. I overthought it :) i've updated the tests accordingly.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jieru-hu: per our offline discussion, removed this check

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jieru-hu: removed :)

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jieru-hu: nice! thank you!

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
omry: Ah, I missed it. the function signature is a bit cumbersome.
this would probably work:

```
def create_nevergrad_param_from_config(config: Union[MutableSequence[Any], MutableMapping[str, Any]]
```

Also, you declare that the input can also be a string but I am not seeing any handling.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/config.py
```diff
@@ -78,9 +80,6 @@ class NevergradSweeperConf:
     # - as a full scalar specification
     parametrization: Dict[str, Any] = field(default_factory=dict)
 
-    # version of the commandline API
-    version: int = 1
-
```
omry, line 83: Good.
Was this removed from the code too?

website/docs/plugins/nevergrad_sweeper.md
```diff
@@ -112,23 +113,42 @@ name: nevergrad
 
 ## Defining the parameters
 
-The plugin can use 2 types of parameters:
+The plugin supports 2 types of parameters: [Choices](https://facebookresearch.github.io/nevergrad/parametrization_ref.html#nevergrad.p.Choice) and [Scalars](https://facebookresearch.github.io/nevergrad/parametrization_ref.html#nevergrad.p.Scalar). They can be defined either through config file or commandline override.
```
omry, line 116: It's not the plugin, it's Nevergrad, right?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -257,7 +282,13 @@ def sweep(self, arguments: List[str]) -> None:
         parsed = parser.parse_overrides(arguments)
 
         for override in parsed:
-            params[override.get_key_element()] = make_nevergrad_parameter(override)
+            if override.key_or_group == "batch_size":
+                import pdb;pdb.set_trace()
```
omry: oops?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -96,29 +96,54 @@ def parse(cls, override: Override) -> Any:
...
 
-   choice(1,2,3)  ==>  ng.p.Choice([["Helium", "Nitrogen", "Oxygen"]])
-   tag(transition,choice(2,3,4))  ==> ng.p.TransitionChoice(["Solid", "Liquid", "Gas"])
-   tag(transition, range(1, 11, 2)) ==> ng.p.TransitionChoice([1, 3, 5 ...])
-   range(1,5) ==>  ng.p.Scalar(lower=1, upper=12).set_integer_casting(),
-   interval(1.0,5.0) ==> np.p.Scalar(lower=1.0, upper=5.0)
-   tag(log, interval(0.1, 1)) ==> np.p.Log(lower=0.1, upper=1)
-    """
+     # We can support forcing a choice to be ordered by tagging it:
+     # (I prefer order over transition(al). at least to me it's clearer.
```
omry: No need to mention your preferences in the comments. also order vs ordered.
Once the decision is made, it is what it is. no need to include a full history of every decision in the docs.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -96,29 +96,54 @@ def parse(cls, override: Override) -> Any:
...
+     # ranges are always ordered:
+     range(1,5)  =>  ng.p.TransitionChoice([1,2,3,4])
+
+     # unless they are shuffled (There is a flag shuffle in the RangeSweep object.
+     shuffle(range(1,5)) ->  =>  ng.p.Choice([4,2,1,3])
+
+     # intervals are scalars:
+     # Note: by intervals are always interpreted as floats (even for int start and end values).
+     interval(0,1)     -> RangeSweep(start=0.0, end=1.0) -> ng.p.Scalar(lower=0.0, upper=1.0)
+     interval(0.0,1.0) -> RangeSweep(start=0.0, end=1.0) -> ng.p.Scalar(lower=0.0, upper=1.0)
```
omry: RangeSweep is an implementation detail of Hydra.
The reader should know about `interval(0,1)` and about `ng.p.Scalar(lower=0.0, upper=1.0)`

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
```diff
@@ -96,29 +96,54 @@ def parse(cls, override: Override) -> Any:
...
-            pass
+            if val.shuffle:
+                return ng.p.Choice(val.list)
+            else:
+                return ng.p.TransitionChoice(val.list)
         elif override.is_interval_sweep(): # continuous variable
-            pass
+            return ng.p.Scalar(lower=val.start, upper=val.end)
+
+        # int(interval(0,1) -> RangeSweep(start=0, end=1) -> ng.p.Scalar(lower=0.0, upper=1.0).set_integer_casting()
```
omry: remove commented out code?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
omry: Also, try to use your IDE for debugging (pycharm works well).

website/docs/plugins/nevergrad_sweeper.md
jieru-hu: Here's the list of params nevergrad supports :https://facebookresearch.github.io/nevergrad/parametrization_ref.html#parameters

Our plugins covers half of them. For future improvements we could look into support more :)

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jieru-hu: I think this is outdated code :) 

Sounds good! i will check out pycharm debugging!

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jieru-hu: This is outdated code, something wrong with github UI?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jieru-hu: outdated code :(

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
jieru-hu: outdated code :( this is no longer there

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/config.py
jieru-hu: yes this is removed :)

website/docs/plugins/nevergrad_sweeper.md
omry: roger.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jieru-hu: thx, `str` shouldn't be there, removed.

Thx for the typing tips! much cleaner now!

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -0,0 +1,183 @@
...
+
+from .config import OptimConf, ScalarConfigSpec
+
+log = logging.getLogger(__name__)
+
+
+def create_nevergrad_param_from_config(
+    config: Union[MutableSequence[Any], MutableMapping[str, Any]]
+) -> Any:
+    if isinstance(config, (list, ListConfig)):
```
omry, line 37: Does this work now?
```suggestion
    if isinstance(config, MutableSequence):
```

jieru-hu: yea, it worked (tests passed and mypy didnt complain)

thx for catching them, fixed now.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -0,0 +1,183 @@
...
+        return scalar
+    return config
+
+
+def create_nevergrad_parameter_from_override(override: Override) -> Any:
+    val = override.value()
+    if not override.is_sweep_override():
+        return val
+    if override.is_choice_sweep():
+        # val = cast(ChoiceSweep, val)
```
omry, line 64: remove comment like this one.

website/docs/plugins/nevergrad_sweeper.md
```diff
@@ -75,24 +76,24 @@ The initialization of the sweep and the first 5 evaluations (out of 100) look li
...
 
 
 and the final 2 evaluations look like this:
 ```text
 [HYDRA] 	#8 : db=mnist batch_size=4 lr=0.094 dropout=0.381
-[__main__][INFO] - dummy_training(dropout=0.381, lr=0.094, db=mnist, batch_size=4) = 1.077
+[__main__][INFO] - my_app.py(dropout=0.381, lr=0.094, db=mnist, batch_size=4) = 1.077
 [HYDRA] 	#9 : db=mnist batch_size=4 lr=0.094 dropout=0.381
-[__main__][INFO] - dummy_training(dropout=0.381, lr=0.094, db=mnist, batch_size=4) = 1.077
+[__main__][INFO] - my_app.py(dropout=0.381, lr=0.094, db=mnist, batch_size=4) = 1.077
```
omry, line 96: the output really contains `my_app.py` ?

omry (approved):
Overall looks good.
A few minor comments. feel free to merge after you address these.

next is to release and update the docs in the 1.0 docs version, but I think it can be in a different issue.

website/docs/plugins/nevergrad_sweeper.md
jieru-hu: 😂 that is a copy paste oversight. will fix

jieru-hu:
> Overall looks good.
> A few minor comments. feel free to merge after you address these.
> 
> next is to release and update the docs in the 1.0 docs version, but I think it can be in a different issue.



thx for the review @omry 
will work on the release tomorrow
My understanding is I need to cherry pick the commit to the 1.0 branch and also update the version 1.0 doc, is that correct?

omry:
sounds right.

jieru-hu:
> sounds right.

cool! will get to them tomorrow!

Issue #870, opened 2020-08-12
Update the Ax Plugin to use the new OverridesParser 


omry:
Followup for #752.
Hydra 1.0.0rc3 introduces new command line grammar that supports advanced sweeprs.
You can read about it [here](https://hydra.cc/docs/next/advanced/override_grammar/extended).

The Ax sweeper should use hydra.core.override_parser.overrides_parser.OverridesParser to parse the command line to take advantage of new grammar.

You can see simple usage example in the [Example Sweeper Plugin ](https://github.com/facebookresearch/hydra/blob/master/plugins/examples/example_sweeper_plugin/hydra_plugins/example_sweeper_plugin/example_sweeper.py#L73-L94) and in the [Basic Sweeper Plugin](https://github.com/facebookresearch/hydra/blob/master/hydra/_internal/core_plugins/basic_sweeper.py#L107-L121).

shagunsodhani:
I can take it up :)

shagunsodhani:
Closed in https://github.com/facebookresearch/hydra/commit/108af1aec8ef86d752f46ec4c5542ee64fd80412

PR #912, opened 2020-08-24
[Ax Sweeper]: Integrate override parser

shagunsodhani:
## Motivation

Integrate override parser.

Todo:

- [x] Clarify how to integrate fixed-params
- [x] Update documentation
- [x] Remove type inference logic
- [x] Add news entry

Yes

## Test Plan

CI

## Related Issues and PRs

#870


shagunsodhani:
@omry I have pushed a WIP. No need to review at this point. I have one question:

Ax supports a parameter-type called [FixedParameter](https://ax.dev/api/ax.html#ax.FixedParameter). The extended grammar does not support this type (because we can not sweep over the FixedParameter Type). One hacky workaround could be to force the user to use the Choice type and check (in the implementation of Ax Plugin) if the choice sweep has only one value. If yes, we convert that choice sweep into fixed parameter. The broader problem here is, since we do not have a counterpart for fixed params (in extended grammar), [is_sweep_override()](https://github.com/facebookresearch/hydra/pull/912/files#diff-7a93c2dc6eb26f2ff824930dfc9a60b5R288) can not be used to decide if the command line argument is meant for Ax plugin or not. Hence, [usecases where we have more command line args, than what  Ax plugin needs](https://github.com/facebookresearch/hydra/pull/912/files#diff-7e2f83750c977fec26bb8f3b6738e2edL121), can not be supported.

omry:
FixedParameter is not a regular non-sweep parameter?

```
--multirun choice=a,b,c fixed=1
```

shagunsodhani:
It is a regular non-sweep parameter. But now I do not have a way to filter out command line args like [this](https://github.com/facebookresearch/hydra/blob/23d3e1b553756be779492fb3238c20f3b4d97546/plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py#L80).

shagunsodhani:
Should we create a news entry as well ?

omry:
> Edit: sorry I created some confusion. Describing the issue correctly now.
> 
> It is a regular non-sweep parameter. But it will be useful to have a way to filter out command line args like [this](https://github.com/facebookresearch/hydra/blob/23d3e1b553756be779492fb3238c20f3b4d97546/plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py#L80) that are not needed by the AxClient.

Filter it based on what criterion?

omry:
> Should we create a news entry as well ?

Of course, this is breaking change for the plugin. We should also bump a major version. (1.1.0?)

omry:
For some reason your PR is missing netlify.
@jieru-hu, your recent PR #908 does have it, any idea what the difference is?

shagunsodhani:
> > Edit: sorry I created some confusion. Describing the issue correctly now.
> > It is a regular non-sweep parameter. But it will be useful to have a way to filter out command line args like [this](https://github.com/facebookresearch/hydra/blob/23d3e1b553756be779492fb3238c20f3b4d97546/plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py#L80) that are not needed by the AxClient.
> 
> Filter it based on what criterion?

If we had a FixedParam type (probably a special case of choice sweep), then we could check if `is_sweep_override` flag is set or not. Otherwise, we will have to keep passing all the extra params to Ax as FixedParams

shagunsodhani:
> #908

Are you referring to this : https://deploy-preview-912--hydra-preview.netlify.app/

website/docs/plugins/ax_sweeper.md
```diff
@@ -56,7 +56,8 @@ AxSweeper is launching 5 jobs
...
...]`. Note that in the case of `x`, both the upper and the lower range values are integers, and hence only integers are sampled. In the case of `y`, the lower range value is an int while the upper range value is a float. The lower range value is promoted to float as well, and floating-point numbers are sampled from the interval. Other supported formats are fixed parameters (eg `banana.x=5.0`) and choice parameters (eg `banana.x=choice(1,2,3)`). Note that `interval`, `choice` etc. are functions provided by Hydra, and you can read more about them [here](https://hydra.cc/docs/next/advanced/override_grammar/extended/). An important thing to remember is, use [`interval`](https://hydra.cc/docs/next/advanced/override_grammar/extended/#interval-sweep) when we want Ax to sample values from an interval. [`RangeParameter`(https://ax.dev/api/ax.html#ax.RangeParameter) in Ax is equivalent to `interval` in Hydra. [`range`](https://hydra.cc/docs/next/advanced/override_grammar/extended/#range-sweep) can be used as an alternate way of specifying choice parameters. For example `python example/banana.py -m banana.x=choice(1, 2, 3, 4)` is equivalent to `python example/banana.py -m banana.x=range(1, 5)`.
```
omry, line 59: when providing shell examples with functions, remember to quote properly.

website/docs/plugins/ax_sweeper.md
```diff
@@ -70,15 +71,15 @@ banana_y:
...
+Note that when using the config file, the parameter type (int, float, string, etc.) is set via the [`parameter_type` attribute](https://ax.dev/api/core.html?highlight=range#module-ax.core.parameter). One important thing to note is how mixed types (float + int) are handled. For the `y` parameter, the bounds were set to be `-5` to `10.1`. Both the values were upcasted to a float (irrespective of whether the range was set via the command line or the config file). In case the user provides the `parameter_type` attribute in the config, the attribute is not changed when type casting is done. If the user wants to sample integers in the interval `-5` to `5`, they need to specify the range as `interval(-5, 5)` (in the command line) or `[-5, 5]` (in config). If they want to sample floats in range `-5` to `5`, they need to specify the range as `interval(-5.0, 5.0)` (in the command line) or `[-5.0, 5.0]` (in config). The type casted values, and unchanged attributes are passed to the Ax Client.
```
omry, line 82: Intervals are always floating points, can Ax handler integer intervals (representing non contentious set?)

shagunsodhani: Yeah. For example, the output for 

> Nice.
> 
> Side note:
> I abandoned the idea of compact keys in OmegaConf. This is a good chance to switch back to the more intuitive usage of dot and not _ to separate the path element in keys:
> 
> ```yaml
>       params:
>         banana.x:
>           type: range
>           bounds: [-5, 5]
> ```

Would it make sense as a separate issue ? Or should I fix in this?

shagunsodhani:
> Nice.
> 
> Side note:
> I abandoned the idea of compact keys in OmegaConf. This is a good chance to switch back to the more intuitive usage of dot and not _ to separate the path element in keys:
> 
> ```yaml
>       params:
>         banana.x:
>           type: range
>           bounds: [-5, 5]
> ```

Would it make sense as a separate issue or should we fix in this ?

website/docs/plugins/ax_sweeper.md
shagunsodhani: Good catch!

website/docs/plugins/ax_sweeper.md
shagunsodhani: It does. For example, for the command: `python example/banana.py -m banana.x=interval(-5, 5) banana.y=interval(-5, 10.1)`

The output of a run looks like:

```
[HYDRA] AxSweeper is optimizing the following parameters:
banana.x: range=[-5, 5], type = int
banana.y: range=[-5.0, 10.1], type = float
ax.modelbridge.dispatch_utils: Using Bayesian Optimization generation strategy: GenerationStrategy(name='Sobol+GPEI', steps=[Sobol for 5 arms, GPEI for subsequent arms], generated 0 arm(s) so far). Iterations after 5 will take longer to generate due to model-fitting.
AxSweeper is launching 5 jobs
[HYDRA] Launching 5 jobs locally
[HYDRA]     #0 : banana.x=4 banana.y=-1.484
[__main__][INFO] - Banana_Function(x=4, y=-1.484)=30581.473
[HYDRA]     #1 : banana.x=3 banana.y=-3.653
[__main__][INFO] - Banana_Function(x=3, y=-3.653)=16014.261
[HYDRA]     #2 : banana.x=0 banana.y=9.409
[__main__][INFO] - Banana_Function(x=0, y=9.409)=8855.340
[HYDRA]     #3 : banana.x=-4 banana.y=2.059
[__main__][INFO] - Banana_Function(x=-4, y=2.059)=19459.063
[HYDRA]     #4 : banana.x=-3 banana.y=-1.338
[__main__][INFO] - Banana_Function(x=-3, y=-1.338)=10704.497
[HYDRA] New best value: 8855.340, best parameters: {'banana.x': 0, 'banana.y': 9.409}
```

Note the values of x are all integers.

omry:
> > Nice.
> > Side note:
> > I abandoned the idea of compact keys in OmegaConf. This is a good chance to switch back to the more intuitive usage of dot and not _ to separate the path element in keys:
> > ```yaml
> >       params:
> >         banana.x:
> >           type: range
> >           bounds: [-5, 5]
> > ```
> 
> Would it make sense as a separate issue or should we fix in this ?

Separate is fine.

website/docs/plugins/ax_sweeper.md
omry: Does Nevergrad has similar capability?

Currently intervals are always floating points, but we can change that.
an alternative is to utilize tag here:
```python
tag(int,interval(1,2))
```

omry: What is the difference between interval(1,10) and range (1,10) ?

omry: one thing that could work well is to default interval to floating point (even if input is ints), but to allow casting it to int:
```
interval(1,10) -> float interval, 1 <= x < 10
int(interval(1,10)) -> int interval, 1 <= x < 10
```

shagunsodhani: My interpretation for the Ax-plugin is as follows:

1. Hydra interval maps to Ax range.
2. Hydra interval is always continous while Ax range can be discrete or continous.
3. Hydra range maps to Ax choice. 
4. Ax-plugin would map hydra interval (1, 10) to Ax range (1, 10), with type as int. 
5. Ax-plugin would map hydra range (1, 10) to Ax choice [1, 2, ...9], with type as int.

shagunsodhani: Not sure how Nevergrad handles it but `tag(int,interval(1,2))` seems complex. The current behaviour is faithful to the previous behaviour.

omry: interval(1,10) is IntervalSweep(1.0, 10.0).
There is currently no such thing as an integer interval in Hydra.

omry: To keep Hydra 1.0 on track, I am not going to change that.
Use tag to support this for now. we can revisit once we get some feedback and have better understanding of the usefulness of this.
I believe the correct interpretation of an interval is a floating point interval even if the numbers specified are integers.

We can either add support for casting an interval (which is not supported now because it's always a float), or support it via the tag mechanism.

shagunsodhani: I agree with `My concern is that in 95% of the cases the user actually wants a regular floating point interval, so that case should be easy and intuitive.`. What is the best way to go ahead with that ?

omry: what I said.
```
interval(1,10) -> float interval, 1 <= x < 10
int(interval(1,10)) -> int interval, 1 <= x < 10

or
tag(int,interval(1,10))
```

I would rather that we added int casting support to interval.

omry: Looks like Nevergrad does not have a similar int interval (https://facebookresearch.github.io/nevergrad/parametrization_ref.html#nevergrad.p.Scalar).

omry: file a followup task to add support to casting interval to int.

shagunsodhani: Okay will get to it towards the end of the day

omry:
> > #908
> 
> Are you referring to this : https://deploy-preview-912--hydra-preview.netlify.app/

Yes, it was gone before.
be sure to actually test the docs.
The following will fail because the shell is interpreting the braces:

image

website/docs/plugins/ax_sweeper.md
shagunsodhani: I still think that using tags will be somewhat complicated for the end user and will hurt adoption of the plugin. Would the following be a reasonable compromise:

All intervals will be float intervals and if the user wants to use intervals with integer values, they can use range (where we will typecast the values to be int if start, step and stop, all are int

shagunsodhani: Ahh I think one issue with approach is that we are relying on an internal detail of Ax - that Int Range and choice are the same thing in Ax. I will go with your original suggestion of using tags.

shagunsodhani: One more place I could use some clarification is the following:

When we specify the range via the config file (say [-1, 1]) and param type is not explicitly specified, we interpret the type to be int. Do we want to change these semantics as well?

omry: For now do something reasonable with the config, I think we will need to have a pass at standardizing the how the parameters are configured and how the config and the command line interacts.

> I still think that using tags will be somewhat complicated for the end user and will hurt adoption of the plugin

I strongly disagree.
In this context, tags are used to mark an interval as a very special and unusual interval of int types. In my opinion it's also pretty useless but that's remains to be seen.
If the user actually needs a discrete range they should use range() and not interval().

Also, advanced users will learn how to use tags.

shagunsodhani:
I propose that we should no longer infer the parameter type (when constructing the param object for Ax). Ax is able to do it on its own. So we should let them handle it. This will simplify some of our logic.

omry:
> I propose that we should no longer infer the parameter type (when constructing the param object for Ax). Ax is able to do it on its own. So we should let them handle it. This will simplify some of our logic.

the general answer is: No, we should provide a consistent user experience across all Sweepers. This is more important than simplifying some logic.
Show me a concrete example of what you are talking about.

shagunsodhani:
> > I propose that we should no longer infer the parameter type (when constructing the param object for Ax). Ax is able to do it on its own. So we should let them handle it. This will simplify some of our logic.
> 
> the general answer is: No, we should provide a consistent user experience across all Sweepers. This is more important than simplifying some logic.
> Show me a concrete example of what you are talking about.

Parameter type is inferred for passing to the [`create_experiment` method from Ax Client](https://ax.dev/api/service.html#ax.service.ax_client.AxClient.create_experiment). This implementation detail is specific to Ax plugin only. Simplifying the logic is a side benefit, the motivation is that Ax can infer this parameter type on its own and the plugin/hydra does not need this type information.

shagunsodhani:
Based on offline discussion, we will not infer the parameter type (on behalf of Ax)

shagunsodhani:
Tests related to Nevergrad are failing. 
```
nox > Command black . --check failed with exit code 1:
would reformat /home/circleci/project/plugins/hydra_nevergrad_sweeper/tests/test_commandline_spec.py
would reformat /home/circleci/project/plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/core.py
Oh no! 💥 💔 💥
2 files would be reformatted, 6 files would be left unchanged.
nox > Session lint_plugins-3.6 failed.
```

Rebasing again

news/870.plugin
```diff
@@ -0,0 +1 @@
+Integrate override parser in the Ax Plugin
```
omry, line 1:
```suggestion
Switch Ax parser to use new Hydra command line syntax
```

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -375,3 +281,49 @@ def normalize_key(key: str, str_to_replace: str, str_to_replace_with: str) -> st
...
+    return param
+
+
+def create_choice_param_from_choice_override(override):
+    key = override.get_key_element()
+    value = override.value()
+    param = {
+        "name": key,
+        "type": "choice",
+        "values": value.list,
```
omry, line 303: not good. look at the example plugin. use override.sweep_string_iterator(). This will respect shuffling and glob and in general is the right way to access fields in all in choice() and range().

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -375,3 +281,49 @@ def normalize_key(key: str, str_to_replace: str, str_to_replace_with: str) -> st
...
+    return param
+
+
+def create_choice_param_from_range_override(override):
+    key = override.get_key_element()
+    value = override.value()
+    param = {
+        "name": key,
+        "type": "choice",
+        "values": [val for val in value.range()],
```
omry, line 314: sweep_string_iterator().

If you need the values in the underlying type propose a change to the API to support it.
the string variant (sweep_string_iterator()) is designed to be used for generating coverrides.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -375,3 +281,49 @@ def normalize_key(key: str, str_to_replace: str, str_to_replace_with: str) -> st
...
+    }
+
+    return param
+
+
+def create_fixed_param_from_element_override(override):
+    key = override.get_key_element()
```
omry, line 321: What I don't like about this style is that you separated the check that the value is a value to the usage of the value.

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -131,8 +131,8 @@ def test_jobs_configured_via_cmd(hydra_sweep_runner) -> None:
         assert returns["optimizer"] == "ax"
         assert len(returns) == 2
         best_parameters = returns["ax"]
-        assert math.isclose(best_parameters["quadratic_x"], -2.0, abs_tol=1e-4)
-        assert math.isclose(best_parameters["quadratic_y"], 2.0, abs_tol=1e-4)
+        assert math.isclose(best_parameters["quadratic_x"], -2, abs_tol=1e-4)
```
omry, line 134: 2.0 was wrong?

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -145,7 +145,7 @@ def test_jobs_configured_via_cmd_and_config(hydra_sweep_runner) -> None:
         overrides=[
             "hydra/launcher=basic",
             "hydra.sweeper.ax_config.max_trials=2",
-            "quadratic.x=-5:-2",
```
omry, line 148: how come all your tests are in integers?

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -172,8 +175,8 @@ def test_configuration_set_via_cmd_and_default_config(hydra_sweep_runner) -> Non
             "hydra.sweeper.ax_config.max_trials=2",
             "hydra.sweeper.ax_config.early_stop.max_epochs_without_improvement=2",
             "quadratic=basic",
-            "quadratic.x=-5:-2",
-            "quadratic.y=-1:1",
+            "quadratic.x=interval(-5, -2)",
```
omry, line 178: how come in all the other tests you had to cast the range to int and in this one you don't?

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -189,21 +192,57 @@ def test_configuration_set_via_cmd_and_default_config(hydra_sweep_runner) -> Non
...
         "-m",
         "hydra.run.dir=" + str(tmpdir),
-        "polynomial.x=-5:-2",
-        "polynomial.y=-1,1",
+        "polynomial.x=interval(-5, -2)",
         "polynomial.z=10",
         "hydra.sweeper.ax_config.max_trials=2",
-    ]
+    ] + [cmd_arg]
     result = subprocess.check_output(cmd).decode("utf-8").rstrip()
```
omry, line 242: why is this using a subprocess instead of testing the function directly?

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -212,17 +251,23 @@ def test_example_app(tmpdir: Path) -> None:
...
 
-@pytest.mark.parametrize("overrides", [[], ["quadratic.x_arg=-1:1"]])
-def test_jobs_configured_via_nested_config(hydra_sweep_runner, overrides: list) -> None:
+@pytest.mark.parametrize(
+    "overrides",
+    [[], ["quadratic.x_arg=int(interval(-1, 1))"]],
+)
+def test_jobs_configured_via_nested_config(
+    hydra_sweep_runner,
+    overrides: list,
```
omry, line 269: remove trailing comma like the one after list. black will format in a more compact way.

website/docs/plugins/ax_sweeper.md
```diff
@@ -31,7 +31,7 @@ We include an example of how to use this plugin. The file [`example/banana.py`](
 To compute the best parameters for the Banana function, clone the code and run the following command in the `plugins/hydra_ax_sweeper` directory:
 
 ```
-python example/banana.py -m banana.x=-5:5 banana.y=-5:10.1
+python example/banana.py -m 'banana.x=int(interval(-5, 5))' 'banana.y=interval(-5, 10.1)'
```
omry, line 34: What are you trying to demonstrate by casting here?

website/docs/plugins/ax_sweeper.md
```diff
@@ -70,15 +71,15 @@ banana_y:
  bounds: [-5, 10.1]
 ```
 
-The `x` parameter takes on a "range" of integer values, between `-5` to `5`. The `y` parameter takes on a range of floating-point values between `-5` to `10.1`. In general, the plugin supports all the [Parameters](https://ax.dev/api/core.html?highlight=range#module-ax.core.parameter) that Ax supports. According to the [Ax documentation](https://ax.dev/api/service.html#ax.service.ax_client.AxClient.create_experiment), the required elements in the config are:
+The `x` parameter takes on a "range" of integer values, in the interval `[-5, 5]`. The `y` parameter takes on a range of floating-point values in the interval `[-5, 10.1]`. In general, the plugin supports all the [Parameters](https://ax.dev/api/core.html?highlight=range#module-ax.core.parameter) that Ax supports. According to the [Ax documentation](https://ax.dev/api/service.html#ax.service.ax_client.AxClient.create_experiment), the required elements in the config are:
```
omry, line 74: The `x` parameter is in the integer range -5 to 5. `interval` is a keyword and it always means floating point.

website/docs/plugins/ax_sweeper.md
```diff
@@ -70,15 +71,15 @@ banana_y:
...
+Note that if you want to sample integers in the interval `-5` to `5`, you need to specify the range as `int(interval(-5, 5))` (in the command line) or `[-5, 5]` (in config). If you want to sample floats in range `-5` to `5`, you need to specify the range as `interval(-5, 5)` (in the command line) or `[-5.0, 5.0]` (in config).
```
omry, line 82: again, do not use interval to describe ranges. stay consistent with the grammar doc language.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
shagunsodhani: `override.sweep_string_iterator()` returns a list of strings (does not respect the datatype)

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
shagunsodhani: Maybe something like `override.sweep_value_iterator()` that returns an iterator over the values (while maintaining the correct value type).

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
shagunsodhani: `"value": override.value()`, would be better?

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
shagunsodhani: ohh I think it is looking this because at some point (while working on this PR) I changed the interpretation of interval and the value of output became some floating point number. Then when I updated the interpretation, I switched the value to 2 instead of 2.0. Reverting it.

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
shagunsodhani: Not all. https://github.com/facebookresearch/hydra/pull/912/files#diff-7e2f83750c977fec26bb8f3b6738e2edR178

shagunsodhani: I can add another test or make some of them regular interval

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
shagunsodhani: We are just ensuring that the config is properly set.

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
shagunsodhani: We are reading the output written to the console (not saved)

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
shagunsodhani: ahh nice!

alshamiri5 (approved):

website/docs/plugins/ax_sweeper.md
shagunsodhani: Want to highlight `int(interval(...))` construct.

shagunsodhani:
@omry  https://github.com/facebookresearch/hydra/pull/912/commits/9373fe5155562121343ffabd11c9ea311ec74d77 contains the change to the `Override` API as we discussed. Pushing it here, instead of a new PR, to keep the discussion at one place. Once we agree on the API, I can open a new PR.

The proposed change is to add a function called  [`sweep_iterator`](https://github.com/facebookresearch/hydra/commit/9373fe5155562121343ffabd11c9ea311ec74d77).

Does the [newly added test](https://github.com/facebookresearch/hydra/pull/912/commits/f0efd623b17ee2880bd11ac6119af33aa2240216#diff-7e2f83750c977fec26bb8f3b6738e2edR227) reflect the usecase we want to support?

website/docs/plugins/ax_sweeper.md
omry: There are many things to highlight, why this?
I think it does not make sense for the example. maybe it's better to just point users to the extended grammar page to learn about all the things they can do.

hydra/core/override_parser/types.py
```diff
@@ -256,10 +257,13 @@ def value(
         else:
             return Override._convert_value(self._value)
 
-    def sweep_string_iterator(self) -> Iterator[str]:
+    def sweep_iterator(
+        self, transformer: Union[str, ElementTransformerType] = "identity"
```
omry, line 261: no need for the string constant. just make it lambda only. see my other comment.

omry (changes requested):

website/docs/plugins/ax_sweeper.md
shagunsodhani: I want to highlight because this example explicitly shows the changes the default behavior. Having this example upfront will be useful for people who have been using Ax plugin.

shagunsodhani:
Thanks for the feedback. Does the [newly added test](https://github.com/facebookresearch/hydra/pull/912/commits/f0efd623b17ee2880bd11ac6119af33aa2240216#diff-7e2f83750c977fec26bb8f3b6738e2edR227) reflect the usecase we want to support?

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -224,6 +224,28 @@ def test_ax_logging(tmpdir: Path, cmd_arg: str, expected_str: str) -> None:
...
+    cmd = [
+        sys.executable,
+        "tests/apps/polynomial_with_list_coefficients.py",
+        "-m",
+        "hydra.run.dir=" + str(tmpdir),
+        "hydra.sweeper.ax_config.max_trials=3",
+    ] + [cmd_arg]
+    result = subprocess.check_output(cmd).decode("utf-8").rstrip()
+    assert (
+        "polynomial.coefficients: choice=['[-1,0,1]', '[2,3,4]', '[5,6,7]']" in result
```
omry, line 243: How am supposed to know what this is testing?
what does the fact that the choice appears in the output means about what is actually happening?
maybe it ran one job with ta single list?

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -224,6 +224,28 @@ def test_ax_logging(tmpdir: Path, cmd_arg: str, expected_str: str) -> None:
...
+)
+def test_jobs_using_choice_between_lists(tmpdir: Path, cmd_arg: str) -> None:
+    cmd = [
+        sys.executable,
+        "tests/apps/polynomial_with_list_coefficients.py",
+        "-m",
+        "hydra.run.dir=" + str(tmpdir),
+        "hydra.sweeper.ax_config.max_trials=3",
+    ] + [cmd_arg]
+    result = subprocess.check_output(cmd).decode("utf-8").rstrip()
```
omry, line 241: there is a new utility function (get_run_output or something) you should use now.

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -195,7 +195,7 @@ def test_configuration_set_via_cmd_and_default_config(hydra_sweep_runner) -> Non
 @pytest.mark.parametrize(
     "cmd_arg, expected_str",
     [
-        ("polynomial.y=choice(-1, 0, 1)", "polynomial.y: choice=[-1, 0, 1]"),
+        ("polynomial.y=choice(-1, 0, 1)", "polynomial.y: choice=['-1', '0', '1']"),
```
omry, line 198: why is this becoming strings? does it effect the final results?

website/docs/plugins/ax_sweeper.md
shagunsodhani: It this okay to keep @Omry ?

omry:
Who has the ball here?
FYI, Hydra's master branch is going to switch to 1.1 soon.
If this does not land before you will need to land it to both the 1.0 branch and master if you want it in both.

shagunsodhani:
Will get to it tomorrow.

shagunsodhani:
`Error: No such file or directory - /usr/local/Cellar/git/2.19.1` on CI. Will rerun the tests

shagunsodhani:
Rebase, followed by a force push

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
shagunsodhani: Fixed

shagunsodhani:
Hmm `deploy/netifly` preview doesnt include the latest changes in the documentation

shagunsodhani:
Ahh maybe I was looking at the wrong place.

plugins/hydra_ax_sweeper/tests/apps/polynomial_with_coefficients.yaml
```diff
@@ -0,0 +1,22 @@
+# @package _group_
```
omry, line 1: do not specify `@package` for the primary config file. (and certainly not group).

plugins/hydra_ax_sweeper/tests/apps/polynomial_with_coefficients.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+      experiment:
+        # Defaults to minimize, set to false to maximize
+        minimize: true
+
+      early_stop:
+        # Number of epochs without a significant improvement from
+        # the currently known best parameters
+        # An Epoch is defined as a batch of trials executed in parallel
+        max_epochs_without_improvement: 2
```
omry, line 22: since this is for testing, I don't think we need such documentation here.
keep explanatory comments for user examples.

plugins/hydra_ax_sweeper/tests/apps/polynomial_with_dict_coefficients.py
```diff
@@ -0,0 +1,20 @@
...
+from typing import Any
+
+import hydra
+from omegaconf import DictConfig
+
+
+@hydra.main(config_name="polynomial_with_dict_coefficients")
+def polynomial_with_dict_coefficients(cfg: DictConfig) -> Any:
+    coefficients = cfg.polynomial.coefficients
+    x, y, z = coefficients["x"], coefficients["y"], coefficients["z"]
```
omry, line 11:
```suggestion
    c = cfg.polynomial.coefficients
    x, y, z = c.x, c.y, c.z
```

plugins/hydra_ax_sweeper/tests/apps/polynomial_with_dict_coefficients.py
```diff
@@ -0,0 +1,20 @@
...
+    a = 100
+    b = 10
+    c = 1
+    result = a * (x ** 2) + b * y + c * z
+    return result
```
omry, line 16:
```suggestion
    return = a * (x ** 2) + b * y + c * z
```

plugins/hydra_ax_sweeper/tests/apps/polynomial_with_dict_coefficients.yaml
```diff
@@ -0,0 +1,22 @@
+# @package _group_
```
omry, line 1: same comments.

plugins/hydra_ax_sweeper/tests/apps/polynomial_with_list_coefficients.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@hydra.main(config_name="polynomial_with_coefficients")
+def polynomial_with_list_coefficients(cfg: DictConfig) -> Any:
+    x, y, z = cfg.polynomial.coefficients
+    a = 100
+    b = 10
+    c = 1
+    result = a * (x ** 2) + b * y + c * z
+    return result
```
omry, line 15: same comment

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -159,6 +157,9 @@ def test_jobs_configured_via_cmd_and_config(hydra_sweep_runner: TSweepRunner) ->
         assert math.isclose(best_parameters["quadratic_x"], -2.0, abs_tol=1e-4)
         assert math.isclose(best_parameters["quadratic_y"], 1.0, abs_tol=1e-4)
 
+        assert math.isclose(best_parameters["quadratic_x"], -2, abs_tol=1e-4)
+        assert math.isclose(best_parameters["quadratic_y"], 1, abs_tol=1e-4)
```
omry, line 161: Since this is a Hydra plugin, I am assuming best_parameters is a DictConfig.
use attribute style access.

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -191,39 +192,128 @@ def test_configuration_set_via_cmd_and_default_config(
...
+        "hydra.run.dir=" + str(tmpdir),
+        "hydra.sweeper.ax_config.max_trials=3",
+    ] + [cmd_arg]
+    result, _ = get_run_output(cmd)
+    assert f"polynomial.coefficients: {serialized_encoding}" in result
+    assert f"'polynomial.coefficients': {best_coefficients}" in result
+    assert f"New best value: {best_value}" in result
+
+
+@pytest.mark.xfail  # type: ignore
```
omry, line 262: what is this expected to fail?

website/docs/plugins/ax_sweeper.md
omry: Your call.

website/docs/plugins/ax_sweeper.md
```diff
@@ -69,15 +70,15 @@ banana_y:
  bounds: [-5, 10.1]
 ```
 
-The `x` parameter takes on a "range" of integer values, between `-5` to `5`. The `y` parameter takes on a range of floating-point values between `-5` to `10.1`. In general, the plugin supports all the [Parameters](https://ax.dev/api/core.html?highlight=range#module-ax.core.parameter) that Ax supports. According to the [Ax documentation](https://ax.dev/api/service.html#ax.service.ax_client.AxClient.create_experiment), the required elements in the config are:
+The `x` parameter takes on a range of integer values, between `-5` to `5`. The `y` parameter takes on a range of floating-point values between `-5` to `10.1`. In general, the plugin supports all the [Parameters](https://ax.dev/api/core.html?highlight=range#module-ax.core.parameter) that Ax supports. According to the [Ax documentation](https://ax.dev/api/service.html#ax.service.ax_client.AxClient.create_experiment), the required elements in the config are:
```
omry, line 73: This feels misplaced after the long exposition the earlier paragraph.
Read the whole doc and see that it flows well.

omry (changes requested):

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
shagunsodhani: This is the issue https://github.com/facebookresearch/hydra/issues/946. Seems it has been fixed. Will try without marking the test as failed.

plugins/hydra_ax_sweeper/tests/apps/polynomial_with_dict_coefficients.py
```diff
@@ -7,13 +7,12 @@
 
 @hydra.main(config_name="polynomial_with_dict_coefficients")
 def polynomial_with_dict_coefficients(cfg: DictConfig) -> Any:
-    coefficients = cfg.polynomial.coefficients
-    x, y, z = coefficients["x"], coefficients["y"], coefficients["z"]
+    c = cfg.polynomial.coefficients
+    x, y, z = c["x"], c["y"], c["z"]
```
omry, line 11:
```
x, y, z = c.x, c.y, c.z
```

Or better yet, just inline them in 
```
return a * (c.x ** 2) + b * c.y + c * c.z
```

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -103,9 +103,9 @@ def test_jobs_configured_via_config(hydra_sweep_runner: TSweepRunner) -> None:
         assert isinstance(returns, DictConfig)
         assert returns["optimizer"] == "ax"
         assert len(returns) == 2
-        best_parameters = returns["ax"]
-        assert math.isclose(best_parameters["quadratic_x"], 0.0, abs_tol=1e-4)
-        assert math.isclose(best_parameters["quadratic_y"], -1.0, abs_tol=1e-4)
+        best_parameters = returns.ax
```
omry, line 106: Why are you faking the return?

plugins/hydra_ax_sweeper/tests/apps/polynomial_with_dict_coefficients.py
shagunsodhani: Fixed

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
shagunsodhani: Discussed offline.

omry:
Since this is a major change, can you bump the plugin version to 1.1.0dev1
?

news/870.plugin
```diff
@@ -0,0 +1 @@
+Switch Ax parser to use new Hydra command line syntax
```
omry, line 1: Use api_change here, this is a breaking change in this plugin.

shagunsodhani:
> 1.1.0dev1

iirc we didnt do a 1.0 release yet. So shouldnt it be `1.0.0rc8` ?

omry:
> > 1.1.0dev1
> 
> iirc we didnt do a 1.0 release yet. So shouldnt it be `1.0.0rc8` ?

People are already using it, I think this is 1.1.0dev1

omry:
God damn, I accidentally rebased instead of squshing and added 30 commits to master.
I had to do a force push to master to fix it. let me know if you run into any issues.

PR #921, opened 2020-08-25
Add support for casting interval to ints and floats

### LINKED ISSUE #915: Support casting intervals to int
Edit by omry:

Interval are for continuous values, interval(0,1) represents all the infinite range `0.0 <= < x < 1.0`.
For that reason, intervals are always representing floating point values and cannot currently be cast to int.

Ax supports something unique which an interval of integers.
the optimization is done in the continuous space, but the values outputted by the optimization process are always integers (probably achieved via a rounding of the optimization result to the nearest integer).
To support this use case, we can enable casting an interval to an int:

```
# always floating point
interval(0,1) -> IntervalSweep(0.0, 1.0) 
interval(0.0, 1.0)  -> IntervalSweep(0.0, 1.0) 

# integer values start and end can potentially be interpreted as  "perform a continuous optimization but output integer values"
int(interval(0,1)) -> IntervalSweep(0, 1) 
int(interval(0.0, 1.0))  -> IntervalSweep(0, 1) 

# An alternative approach which is already supported by the grammar is to use tagging:
tag(int, interval(0,1)) -> IntervalSweep(0.0, 1.0, tags=[int]) 
tag(int,interval(0.0, 1.0)) -> IntervalSweep(0.0, 1.0, tags=[int]) 
```

Tagging is actually more consistent because what we do here is passing a special parameter to the sweeper and not really casting the input.
I think I prefer the tagging approach but would like to have a discussion around it.
We can also use a less overloaded tag such as `int_output`.

omry:
closes #915

### DIFF
--- a/hydra/core/override_parser/types.py
+++ b/hydra/core/override_parser/types.py
@@ -124,6 +124,23 @@ class IntervalSweep(Sweep):
     start: Optional[float] = None
     end: Optional[float] = None
 
+    def __eq__(self, other: Any) -> Any:
+        if isinstance(other, IntervalSweep):
+            eq = (
+                self.start == other.start
+                and self.end == other.end
+                and self.tags == other.tags
+            )
+
+            st = type(self.start)
+            ost = type(other.start)
+            et = type(self.end)
+            ose = type(other.end)
+            eq = eq and st == ost and et is ose
+            return eq
+        else:
+            return NotImplemented
+
 
 # Ideally we would use List[ElementType] and Dict[str, ElementType] but Python does not seem
 # to support recursive type definitions.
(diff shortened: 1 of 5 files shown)


hydra/core/override_parser/types.py
```diff
@@ -124,6 +124,23 @@ class IntervalSweep(Sweep):
...
+                and self.end == other.end
+                and self.tags == other.tags
+            )
+
+            st = type(self.start)
+            ost = type(other.start)
+            et = type(self.end)
+            ose = type(other.end)
+            eq = eq and st == ost and et is ose
+            return eq
```
shagunsodhani, line 140: `eq = eq` ?

omry: yes

shagunsodhani: sorry I didnt understand. What does that do?

omry: it computes a new boolean from the boolean eq and the expression following it, and assigns it to the variable eq.

shagunsodhani: Ahh right - thanks!

PR #952, opened 2020-09-08
Dictionaries and lists specified in the command line are now merged

### LINKED ISSUE #939: [Feature Request] Support merging dictionaries from command line
```
@dataclass
class Stage:
    value: int = 1

@dataclass
class Config:
  stages: List[Stage] = field(default_factory=list)
```

The following doesn't work when using structured Config from command line as it complains about mismatched types dict vs Stage:
```
stages=[{value:10}]
```

Error:
```
Error merging override stages=[{value: 10}]
Invalid type assigned : dict is not a subclass of Stage. value: {'value': 10}
        full_key: stages[0]
        reference_type=List[Stage]
        object_type=list
```

### LINKED ISSUE #946: [Bug] Assigning a dictionary to untyped field fails
I am expecting this to work:
```
$ python examples/tutorials/basic/your_first_hydra_app/5_defaults/my_app.py  'db.driver={a:10,b:20}' 
Could not override 'db.driver'. No match in config.
To append to your config use +db.driver={a:10,b:20}
Key 'a' is not in struct
        full_key: db.driver.a
        reference_type=Any
        object_type=dict
```

omry:
This closes two issues around merging dicts/lists specified in the command line.
There is a large surface area here so it's likely that there will be more issues around this.

Closes #939 
Closes #946

### DIFF
--- a/hydra/_internal/config_loader_impl.py
+++ b/hydra/_internal/config_loader_impl.py
@@ -8,6 +8,7 @@ import re
 import warnings
 from collections import defaultdict
 from dataclasses import dataclass
+from textwrap import dedent
 from typing import Any, Dict, List, Optional, Tuple
 
 from omegaconf import Container, DictConfig, ListConfig, OmegaConf, open_dict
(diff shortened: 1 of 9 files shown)


omry:
cc @thoth291

PR #879, opened 2020-08-14
[OmegaConf 2.1 FUTURE COMPATIBILITY, REVIEW ONLY] Add support for nested interpolations -- also fix handling of whitespaces in primitive strings

odelalleau:
## Motivation

Allow support for nested interpolations in OmegaConf while keeping Hydra's grammar close to the one in OmegaConf (see https://github.com/omry/omegaconf/pull/321 -- note that this PR may not be up-to-date compared to the one in Hydra)

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebookresearch/hydra/blob/master/CONTRIBUTING.md)?

Yes

## Test Plan

`pytest tests/test_overrides_parser.py`

In  particular I added a test `test_omegaconf_interpolations` that ensures Hydra's grammar can parse typical interpolations being tested in OmegaConf.

## Related Issues and PRs

#873
https://github.com/omry/omegaconf/pull/321

PR #954, opened 2020-09-09
Minor override grammar updates

### LINKED ISSUE #962: [Bug] Cannot use overrides containing '%' character
# 🐛 Bug
## Description
I am unable to override a value that contains the '%' character without doing special quoting around the override

## To reproduce

```
import hydra

@hydra.main()
def foo(*args, **kargs):
    pass

foo()
```

```
python test.py +bar=/path/to/pattern_%r.bin
```

```
LexerNoViableAltException: +bar=/path/to/pattern_%r.bin
                                                 ^

```

## Expected Behavior
bar's value should be set to "/path/to/pattern_%r.bin".

## System information
- **Hydra Version** :  1.0.0
- **Python version** : 3.7.7
- **Operating system** : 

## Additional context
Similar error to: Issue #798.

odelalleau:
## Motivation

This is the part of #879 that contains small changes I made along the path towards support for nested interpolations (but does not actually implement this support -- this will come in a different PR).

It contains a number of minor fixes / refactoring. Having these in master will make the diff to support nested interpolations smaller (and easier to maintain in a different branch while waiting for Hydra's next major version).

NB: the user-facing changes seem too minor to me to warrant new items, but I'm happy to add some if needed (please just tell me which commits should be included in the news)

Yes

## Test Plan

pytest

## Related Issues and PRs

#879 

Closes #962


hydra/grammar/OverrideLexer.g4
```diff
@@ -43,29 +44,28 @@ BRACE_CLOSE: WS? '}';
...
-fragment EXPONENT: [eE] [+-]? DIGIT+;
-fragment EXPONENT_FLOAT: ( INT_PART | POINT_FLOAT) EXPONENT;
-FLOAT: [-]?(POINT_FLOAT | EXPONENT_FLOAT | [Ii][Nn][Ff] | [Nn][Aa][Nn]);
-INT: [-]? ('0' | (NZ_DIGIT DIGIT*));
+// Numbers.
+
+fragment POINT_FLOAT: INT_UNSIGNED '.' | INT_UNSIGNED? '.' DIGIT (('_')? DIGIT)*;
+fragment EXPONENT_FLOAT: (INT_UNSIGNED | POINT_FLOAT) [eE] [+-]? INT_UNSIGNED;
+FLOAT: [+-]? (POINT_FLOAT | EXPONENT_FLOAT | [Ii][Nn][Ff] | [Nn][Aa][Nn]);
+INT: [+-]? INT_UNSIGNED;
```
odelalleau, line 52: These changes are related to what was discussed over chat with @omry (w.r.t. underscores & the + sign). I also condensed the definitions a bit to have fewer lines, as personally I actually find it easier to see at a glance what each type of float looks like, but I can revert back to more fragments if needed.

hydra/grammar/OverrideParser.g4
```diff
@@ -30,31 +36,34 @@ element:
...
+
+argName: ID EQUAL;
+function: ID POPEN (argName? element (COMMA argName? element )* )? PCLOSE;
+
+// Data structures.
+
+listValue: BRACKET_OPEN sequence? BRACKET_CLOSE;                          // [], [1,2,3], [a,b,[1,2]]
+dictValue: BRACE_OPEN (keyValuePair (COMMA keyValuePair)*)? BRACE_CLOSE;  // {}, {a:10,b:20}
+keyValuePair: ID COLON element;
+sequence: element (COMMA element)*;
```
odelalleau, line 53: The motivation for the small changes to list / dict definitions is related to the future addition of interpolations in the parser:

* Lists can share the sequence rule with interpolationResolver
* Writing a dict as several key-value pairs makes both the grammar and visitor code more readable IMO (especially with the future addition of interpolations into the mix)

hydra/grammar/OverrideParser.g4
```diff
@@ -30,31 +36,34 @@ element:
...
         | NULL                                   // null, NULL
         | INT                                    // 0, 10, -20, 1_000_000
         | FLOAT                                  // 3.14, -20.0, 1e-1, -10e3
         | BOOL                                   // true, TrUe, false, False
-        | DOT_PATH                               // foo.bar
         | INTERPOLATION                          // ${foo.bar}, ${env:USER,me}
-        | UNQUOTED_CHAR                          // /, -, \, +, ., $, *
+        | UNQUOTED_CHAR                          // /, -, \, +, ., $, %, *
         | COLON                                  // :
+        | ESC                                    // \\, \ , \\t, \,
```
odelalleau, line 67: Adding support for escaped characters makes it consistent with OmegaConf (and will be needed for proper parsing of interpolations)

hydra/grammar/OverrideLexer.g4
```diff
@@ -43,29 +44,28 @@ BRACE_CLOSE: WS? '}';
...
 BOOL:
       [Tt][Rr][Uu][Ee]      // TRUE
     | [Ff][Aa][Ll][Ss][Ee]; // FALSE
 
 NULL: [Nn][Uu][Ll][Ll];
 
-UNQUOTED_CHAR: [/\-\\+.$*];  // other characters allowed in unquoted strings
-VALUE_ID: ID -> type(ID);
-VALUE_DOT_PATH: DOT_PATH -> type(DOT_PATH);
+UNQUOTED_CHAR: [/\-\\+.$%*];  // other characters allowed in unquoted strings
```
odelalleau, line 62: Adding `%` which is a valid character in OmegaConf (see https://github.com/omry/omegaconf/blob/9f7d3566c1ce99da1d112e0e4656bf6651b3a62e/omegaconf/_utils.py#L328)

I believe I just forgot it in the initial version.

hydra/grammar/OverrideParser.g4
omry: let's do the sequence change later. it doesn't quiet make sense in this context.

omry (changes requested):
Great, thanks.
I commented inline.

Otherwise, please document user facing changes.
- Escaping should be mentioned in the docs and in the news
- Little changes like better parsing of ints can be mentioned in a news fragment (you can have multiple news fragments per issue).
- Don't forget to update the parser copy in the website.

hydra/grammar/OverrideParser.g4
odelalleau: Currently this is documented only in OmegaConf, in https://omegaconf.readthedocs.io/en/latest/usage.html#custom-interpolations: _You can use literal commas and spaces anywhere by escaping (\, and \ )_

tests/test_overrides_parser.py
```diff
@@ -52,6 +52,11 @@ def eq(item1: Any, item2: Any) -> bool:
         pytest.param("a b c\t-\t1 2 3", "a b c\t-\t1 2 3", id="value:str-ws-in"),
         pytest.param(" abc-123 ", "abc-123", id="value:str-ws-out"),
         pytest.param("123abc", "123abc", id="value:str-int-id"),
+        pytest.param(r"a\,b", "a,b", id="value:str-esc-comma"),
+        pytest.param(r"\ ab", " ab", id="value:str-esc-space"),
+        pytest.param("ab\\\t", "ab\t", id="value:str-esc-tab"),
+        pytest.param("ab\\\\", "ab\\", id="value:str-esc-backslash"),
+        pytest.param(r"\,", ",", id="value:str-esc-comma-alone"),
```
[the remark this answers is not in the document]
odelalleau: They are easier to read IMO (due to not having to double the \) -- it's very minor though so I can definitely get rid of them

tests/test_overrides_parser.py
```diff
@@ -259,9 +267,9 @@ def test_shuffle_sequence(value: str, expected: Any) -> None:
         pytest.param("{a:10}", {"a": 10}, id="dict"),
         pytest.param("{a:[a,10]}", {"a": ["a", 10]}, id="dict"),
         pytest.param("{a:[true,10]}", {"a": [True, 10]}, id="dict"),
-        pytest.param("{a:10,b:20}}", {"a": 10, "b": 20}, id="dict"),
-        pytest.param("{a:10,b:{}}}", {"a": 10, "b": {}}, id="dict"),
-        pytest.param("{a:10,b:{c:[1,2]}}}", {"a": 10, "b": {"c": [1, 2]}}, id="dict"),
+        pytest.param("{a:10,b:20}", {"a": 10, "b": 20}, id="dict"),
```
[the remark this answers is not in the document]
odelalleau: See full messase of 6a4a35b98b4a8a6577a854898702f31b38b8f161

tests/test_overrides_parser.py
```diff
@@ -61,9 +66,12 @@ def eq(item1: Any, item2: Any) -> bool:
         pytest.param("null", None, id="value:null"),
         # int
         pytest.param("1", 1, id="value:int:pos"),
+        pytest.param("+1", 1, id="value:int:explicit_pos"),
+        pytest.param("1___0___", "1___0___", id="value:int:not_an_int"),
```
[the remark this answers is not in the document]
odelalleau: No, you should quote instead (`"'1_000'"`), though Hydra's is handling quotes in a special way I don't really understand (instead of `"1_000"` this yields `QuotedString(text="1_000", quote=Quote.single)`).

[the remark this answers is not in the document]
omry: The quotes have to be re-constructed in some cases so the parsed result contains information about them.

tests/test_overrides_parser.py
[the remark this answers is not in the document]
omry: gotcha.

hydra/grammar/OverrideParser.g4
omry: Why do we need this capability in Hydra?

hydra/grammar/OverrideParser.g4
odelalleau: Ok, I force pushed a new version that doesn't have this change

hydra/core/override_parser/overrides_visitor.py
```diff
@@ -93,7 +93,16 @@ def visitPrimitive(
...
-            ret = ctx.getText().strip()
+            # Concatenate, but un-escaping as needed.
+            tokens = [
+                n.symbol.text[1::2]
+                if n.symbol.type == OverrideLexer.ESC
+                else n.symbol.text
+                for i, n in enumerate(ctx.getChildren())
+                # Skip leading / trailing whitespaces.
+                if n.symbol.type != OverrideLexer.WS or first_idx <= i < last_idx
+            ]
```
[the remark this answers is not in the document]
odelalleau: Done in f213c452

tests/test_overrides_parser.py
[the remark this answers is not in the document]
omry: just for "\," ?

odelalleau:
I updated the parser doc in 0070fcb, added some doc in f7ccd1e and news items in c68a527. I believe it should be ready now.

hydra/grammar/OverrideParser.g4
odelalleau: It's not super useful for Hydra (it gives the ability to have leading / trailing whitespaces and commas in unquoted values, but one could use quoted values as well, so it doesn't really count as a new functionality).

The main goal is to make Hydra consistent with OmegaConf. This has two advantages:
1. Being able to re-use the same grammar once we support nested interpolations (=> not having to maintain two different grammars); this is really the original motivation for this change
2. More consistent user experience

tests/test_overrides_parser.py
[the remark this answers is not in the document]
odelalleau: I'm using r strings in the first and last lines you quoted. The reason why I'm not using them in the two other lines is because it's not possible:
* Since I want a tab character with `\t` I can't use r strings (otherwise it would become the string `\t` instead of a tab)
* The second one is less obvious: you can't have an r string that ends with `\"`, because it still believes you're trying to escape the quote.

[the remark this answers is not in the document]
odelalleau: Btw Hydra doc added in f7ccd1e

hydra/grammar/OverrideParser.g4
odelalleau: Btw Hydra doc added in f7ccd1e

omry: And are escaped actually needed by in OmegaConf or are they just there to maintain backward compatibility with the initial (stoneage) implementation or interpolations?

odelalleau: They could be removed by forcing people to use quoted strings instead. This would be a breaking change though. Note also that quoted strings can get quite ugly with nested interpolations (as we discussed at some point in the past), which may be a reason to try and keep supporting as many characters as possible in unquoted strings.

omry: I think in general quoting and escaping serves the same need.
Depending on the circumstances, one may be cleaner than the other.
supporting just space, tab and comma seems arbitrary from the perspective of Hydra: Why those characters and not 
'[]{})()=' ?
(and anything else I forgot which is a special token?)

If we are adding escaping support, we might as well make it complete.

hydra/core/override_parser/overrides_visitor.py
```diff
@@ -125,6 +138,8 @@ def visitPrimitive(
                     ret = False
                 else:
                     assert False
+            elif node.symbol.type == OverrideLexer.ESC:
+                ret = node.symbol.text[1::2]
```
omry, line 142: do you have a unit test for a single escape token?

hydra/grammar/OverrideParser.g4
```diff
@@ -30,31 +36,36 @@ element:
...
+function: ID POPEN (argName? element (COMMA argName? element )* )? PCLOSE;
+
+// Data structures.
+
+listValue: BRACKET_OPEN                          // [], [1,2,3], [a,b,[1,2]]
+    (element(COMMA element)*)?
+BRACKET_CLOSE; 
+
+dictValue: BRACE_OPEN (keyValuePair (COMMA keyValuePair)*)? BRACE_CLOSE;  // {}, {a:10,b:20}
+keyValuePair: ID COLON element;
```
omry, line 55: keyValuePair is too generic of a name.
```suggestion
dictKeyValuePair: ID COLON element;
```

news/954.bugfix
```diff
@@ -0,0 +1 @@
+Fix for integer parsing: previously all underscore patterns were allowed, now only valid ones are recognized as integers (ex: `1___0` is not recognized as an integer anymore)
```
omry, line 1: A bit too verbose for line item.
```suggestion
Fix for Override Grammar integer parsing: Integers are now compatible with Python integers ('+' prefix, correct handling of '_')
```

website/versioned_docs/version-1.0/advanced/override_grammar/basic.md
```diff
@@ -142,6 +151,20 @@ Note that trailing and leading whitespace are ignored, the above is equivalent t
...
+```shell
+$ python my_app.py 'msg=\ test\ '
+```
+
+Here is the full list of characters that may be escaped in unquoted values:
+* Space (`\ `) (for trailing / leading whitespaces as in the example above)
+* Tab (`\\t`) (for trailing / leading tabs, though note that passing tabs through your shell can be [tricky](https://stackoverflow.com/questions/6392249/how-to-enter-a-tab-char-on-command-line))
+* Comma (`\,`) (to avoid confusion with the sweeps described below)
+* Backslash (`\\`) (to allow for an actual backslash preceding any of the above characters, e.g. a sweep value ending with a backslash)
+
```
omry, line 167: This is too verbose, but I don't want to suggest edits before we decide on supporting escaping of other characters.

omry (changes requested):

hydra/core/override_parser/overrides_visitor.py
odelalleau: Yes for instance this one has a single esc token:
```python
pytest.param(r"a\,b", "a,b", id="value:str-esc-comma")
```

hydra/grammar/OverrideParser.g4
odelalleau: I'm fine with that, it's very easy to add (I think the only one you're missing is `:`)

hydra/grammar/OverrideParser.g4
odelalleau: Done in cbeb1048

news/954.bugfix
odelalleau: Taking your suggestion

website/versioned_docs/version-1.0/advanced/override_grammar/basic.md
odelalleau: I'll give it a shot while adding more escape characters. I just realized though that there are two `basic.md` files and they seem to be slightly out-of-sync (`website/docs/advanced/override_grammar/basic.md` and `website/versioned_docs/version-1.0/advanced/override_grammar/basic.md`), even in master. Do you know which one is currently the "official" version?

tests/test_overrides_parser.py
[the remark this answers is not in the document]
omry: I never used r-strings like that.
ipython seems to thinks these are the exact same strings:
```python
In [3]: r"\ ab" == "\ ab"
Out[3]: True

In [4]: r"\," == "\,"
Out[4]: True
```

hydra/grammar/OverrideParser.g4
odelalleau: I went ahead and added them in 8d65ff1a

website/versioned_docs/version-1.0/advanced/override_grammar/basic.md
odelalleau: Doc updated in 8d65ff1a

tests/test_overrides_parser.py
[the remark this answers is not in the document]
odelalleau: Oh right, it's true that if the backslash is not followed by a character that makes it a special escape, then Python just accepts it as a backslash. Note however that this triggers a deprecation warning under the hood (you can verify it by calling first `warnings.filterwarnings("default", category=DeprecationWarning)`)

odelalleau:
I had to rebase on master to (hopefully) fix CI failures => some of the commit hashes I mentioned above might be outdated

tests/test_overrides_parser.py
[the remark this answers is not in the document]
omry: gotcha.
fucking stealthy deprecation warnings. such an idiotic design.

hydra/core/override_parser/overrides_visitor.py
odelalleau: Ooh right I misunderstood your original question. I actually have the exact same test you suggested which indeed goes through that path (now that I think about it, that's the reason I added it):
```python
 pytest.param(r"\,", ",", id="value:str-esc-comma-alone"),
```

hydra/grammar/OverrideParser.g4
odelalleau: > Okay, please add it and update the docs and the news fragment.

I did (see previous comment). Just realized I had forgotten to update the parser / doc so I made another commit (5513fdd9f) for that. I'll refactor this PR to squash a few commits together once it's ready to be merged.

website/versioned_docs/version-1.0/advanced/override_grammar/basic.md
odelalleau: > Update the one in docs (that would appear in https://www.internalfb.com/intern/staticdocs/hydra/docs/next/).

Got it. I reverted my changes to versioned docs in 04a53d8aa

website/docs/advanced/override_grammar/basic.md
```diff
@@ -56,34 +62,39 @@ element:
...
       QUOTED_VALUE                               // 'hello world', "hello world"
     | (   ID                                     // foo_10
         | NULL                                   // null, NULL
         | INT                                    // 0, 10, -20, 1_000_000
         | FLOAT                                  // 3.14, -20.0, 1e-1, -10e3
         | BOOL                                   // true, TrUe, false, False
-        | DOT_PATH                               // foo.bar
         | INTERPOLATION                          // ${foo.bar}, ${env:USER,me}
-        | UNQUOTED_CHAR                          // /, -, \, +, ., $, *
+        | UNQUOTED_CHAR                          // /, -, \, +, ., $, %, *
```
omry, line 93: See #962

omry (approved):
Great, thanks!

omry:
I think you missed a test for unquoted %, can you followup if so?

odelalleau:
> I think you missed a test for unquoted %, can you followup if so?

Oops, thanks, good catch, added in #965

PR #976, opened 2020-09-16
Add support for nested interpolations in Override grammar

odelalleau:
## Motivation

Nested interpolations are coming in OmegaConf. The simple regexp currently found in Hydra's lexer won't work with them (`INTERPOLATION: '${' ~('}')+ '}';`).

Note: this PR is intended for next major version. I'm still putting it up for visibility -- I will try and keep it updated until it is time to merge it.

Yes

## Test Plan

`pytest`

## Related Issues and PRs

Corresponding changes in OmegaConf: https://github.com/omry/omegaconf/pull/321
Older version of this PR: #879

### DIFF
--- a/hydra/core/override_parser/overrides_visitor.py
+++ b/hydra/core/override_parser/overrides_visitor.py
@@ -1,7 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 import sys
 import warnings
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 
 from antlr4 import TerminalNode, Token
 from antlr4.error.ErrorListener import ErrorListener
@@ -96,20 +96,25 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
             # Concatenate, while un-escaping as needed.
             tokens = []
             for i, n in enumerate(ctx.getChildren()):
-                if n.symbol.type == OverrideLexer.WS and (
+                if isinstance(n, OverrideParser.InterpolationContext):
+                    tokens.append(n.getText())
+                elif n.symbol.type == OverrideLexer.WS and (
                     i < first_idx or i >= last_idx
                 ):
                     # Skip leading / trailing whitespaces.
-                    continue
-                tokens.append(
-                    n.symbol.text[1::2]  # un-escape by skipping every other char
-                    if n.symbol.type == OverrideLexer.ESC
-                    else n.symbol.text
-                )
+                    pass
+                else:
+                    tokens.append(
+                        n.symbol.text[1::2]  # un-escape by skipping every other char
+                        if n.symbol.type == OverrideLexer.ESC
+                        else n.symbol.text
+                    )
             ret = "".join(tokens)
         else:
             node = ctx.getChild(first_idx)
-            if node.symbol.type == OverrideLexer.QUOTED_VALUE:
+            if isinstance(node, OverrideParser.InterpolationContext):
+                ret = node.getText()
+            elif node.symbol.type == OverrideLexer.QUOTED_VALUE:
                 text = node.getText()
                 qc = text[0]
                 text = text[1:-1]
@@ -122,7 +127,7 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
                 else:
                     assert False
                 return QuotedString(text=text, quote=quote)
-            elif node.symbol.type in (OverrideLexer.ID, OverrideLexer.INTERPOLATION):
+            elif node.symbol.type == OverrideLexer.ID:
                 ret = node.symbol.text
             elif node.symbol.type == OverrideLexer.INT:
                 ret = int(node.symbol.text)
@@ -147,8 +152,15 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
     def visitListValue(
         self, ctx: OverrideParser.ListValueContext
     ) -> List[ParsedElementType]:
-        ret: List[ParsedElementType] = []
+        return (
+            []
+            if ctx.getChildCount() == 2
+            else list(self.visitSequence(ctx.getChild(1)))
+        )
 
+    def visitSequence(
+        self, ctx: OverrideParser.SequenceContext
+    ) -> Iterable[ParsedElementType]:
         idx = 0
         while True:
             element = ctx.element(idx)
@@ -156,8 +168,7 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
                 break
             else:
                 idx = idx + 1
-                ret.append(self.visitElement(element))
-        return ret
+                yield self.visitElement(element)
 
     def visitDictValue(
         self, ctx: OverrideParser.DictValueContext
--- a/hydra/grammar/OverrideLexer.g4
+++ b/hydra/grammar/OverrideLexer.g4
@@ -34,13 +34,15 @@ DOT_PATH: (ID | INT_UNSIGNED) ('.' (ID | INT_UNSIGNED))+;
 
 mode VALUE_MODE;
 
+INTER_OPEN: '${' -> pushMode(INTERPOLATION_MODE);
+BRACE_OPEN: '{' WS? -> pushMode(VALUE_MODE);  // must keep track of braces to detect end of interpolation
+BRACE_CLOSE: WS? '}' -> popMode;
+
 POPEN: WS? '(' WS?;  // whitespaces before to allow `func (x)`
 COMMA: WS? ',' WS?;
 PCLOSE: WS? ')';
 BRACKET_OPEN: '[' WS?;
 BRACKET_CLOSE: WS? ']';
-BRACE_OPEN: '{' WS?;
-BRACE_CLOSE: WS? '}';
 VALUE_COLON: WS? ':' WS? -> type(COLON);
 VALUE_EQUAL: WS? '=' WS? -> type(EQUAL);
 
@@ -69,4 +71,17 @@ QUOTED_VALUE:
       '\'' ('\\\''|.)*? '\'' // Single quotes, can contain escaped single quote : /'
     | '"' ('\\"'|.)*? '"' ;  // Double quotes, can contain escaped double quote : /"
 
-INTERPOLATION: '${' ~('}')+ '}';
+////////////////////////
+// INTERPOLATION_MODE //
+////////////////////////
+
+mode INTERPOLATION_MODE;
+
+NESTED_INTER_OPEN: INTER_OPEN -> type(INTER_OPEN), pushMode(INTERPOLATION_MODE);
+INTER_COLON: ':' WS? -> type(COLON), mode(VALUE_MODE);
+INTER_CLOSE: '}' -> popMode;
+
+DOT: '.';
+INTER_ID: ID -> type(ID);
+LIST_INDEX: INT_UNSIGNED;
+INTER_WS: WS -> skip;
--- a/hydra/grammar/OverrideParser.g4
+++ b/hydra/grammar/OverrideParser.g4
@@ -47,12 +47,17 @@ function: ID POPEN (argName? element (COMMA argName? element )* )? PCLOSE;
 
 // Data structures.
 
-listValue: BRACKET_OPEN                          // [], [1,2,3], [a,b,[1,2]]
-    (element(COMMA element)*)?
-BRACKET_CLOSE;
-
+listValue: BRACKET_OPEN sequence? BRACKET_CLOSE;                          // [], [1,2,3], [a,b,[1,2]]
 dictValue: BRACE_OPEN (dictKeyValuePair (COMMA dictKeyValuePair)*)? BRACE_CLOSE;  // {}, {a:10,b:20}
 dictKeyValuePair: ID COLON element;
+sequence: element (COMMA element)*;
+
+// Interpolations.
+
+interpolation: interpolationNode | interpolationResolver;
+interpolationNode: INTER_OPEN DOT* configKey (DOT configKey)* INTER_CLOSE;
+interpolationResolver: INTER_OPEN (interpolation | ID) COLON sequence? BRACE_CLOSE;
+configKey: interpolation | ID | LIST_INDEX;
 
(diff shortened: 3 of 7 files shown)


hydra/core/override_parser/overrides_visitor.py
```diff
@@ -173,6 +184,8 @@ def visitDictKeyValuePair(
     ) -> Tuple[str, ParsedElementType]:
         children = ctx.getChildren()
         item = next(children)
+        # We should never visit dictionaries that have interpolations as keys
+        # (currently these are only allowed within interpolations, which are not visited).
```
omry, line 188: I don't understand this comment.
1. Since when do we have interpolation as keys in dictionaries?
2. is the code doing anything to address the requirements from this comment?

hydra/grammar/OverrideLexer.g4
```diff
@@ -69,4 +71,17 @@ QUOTED_VALUE:
...
     | '"' ('\\"'|.)*? '"' ;  // Double quotes, can contain escaped double quote : /"
 
-INTERPOLATION: '${' ~('}')+ '}';
+////////////////////////
+// INTERPOLATION_MODE //
+////////////////////////
+
+mode INTERPOLATION_MODE;
+
+NESTED_INTER_OPEN: INTER_OPEN -> type(INTER_OPEN), pushMode(INTERPOLATION_MODE);
```
omry, line 80: not sure if this token type is needed.
looks like it's exactly like INTER_OPEN.
Is this because we need to handle all possible tokens in the mode?

hydra/grammar/OverrideParser.g4
```diff
@@ -47,12 +47,17 @@ function: ID POPEN (argName? element (COMMA argName? element )* )? PCLOSE;
...
 // Data structures.
 
-listValue: BRACKET_OPEN                          // [], [1,2,3], [a,b,[1,2]]
-    (element(COMMA element)*)?
-BRACKET_CLOSE;
-
+listValue: BRACKET_OPEN sequence? BRACKET_CLOSE;                          // [], [1,2,3], [a,b,[1,2]]
 dictValue: BRACE_OPEN (dictKeyValuePair (COMMA dictKeyValuePair)*)? BRACE_CLOSE;  // {}, {a:10,b:20}
-dictKeyValuePair: ID COLON element;
+dictKeyValuePair: (ID | interpolation) COLON element;
```
omry, line 52: I don't think interpolation in a key makes sense. can you show a use case?

hydra/grammar/OverrideParser.g4
```diff
@@ -47,12 +47,17 @@ function: ID POPEN (argName? element (COMMA argName? element )* )? PCLOSE;
...
+sequence: element (COMMA element)*;
+
+// Interpolations.
+
+interpolation: interpolationNode | interpolationResolver;
+interpolationNode: INTER_OPEN DOT* configKey (DOT configKey)* INTER_CLOSE;
```
omry, line 58: add tests for relative interpolations, I think it's not covered here.

news/976.feature
```diff
@@ -0,0 +1 @@
+Override grammar now supports nested interpolations (ex: `python my_app.py 'key=${foo.${bar}}'`)
```
omry, line 1: Bump the version of Hydra to 1.1.0-dev1.
This is the first major change.

tests/test_overrides_parser.py
```diff
@@ -1953,3 +1954,309 @@ def test_sweep_iterators(
...
+# OmegaConf, and properly identify parsing error cases.
+# (should be replaced by imports from OmegaConf in the future)
+GrammarSyntaxError = HydraException
+# The OmegaConf exceptions below cannot be identified at parsing time.
+ConfigKeyError = GrammarTypeError = UnsupportedInterpolationType = None
+
+# DO NOT EDIT `TEST_CONFIG_DATA`!
+# It should be copied regularly from OmegaConf to ensure Hydra's grammar is able
+# to parse similar expressions as OmegaConf's.
+TEST_CONFIG_DATA: List[Tuple[str, Any, Any]] = [
```
omry, line 1969: this test file is already massive.
Can you put the OmegaConf tests in a different test file?

omry (changes requested):

hydra/core/override_parser/overrides_visitor.py
odelalleau: I'll try to make this comment clearer if I can convince you to keep interpolations as keys ;) But to answer your questions:

1. Since https://github.com/omry/omegaconf/pull/321
2. That's actually the assert just below this comment (but I'm not entirely sure I understood your question)

For additional context:
- In OmegaConf, when writing the grammar I added support for dictionaries with interpolations as keys. Such dictionaries can only happen when inside an interpolation.
- When "merging" OmegaConf's grammar within Hydra, I wondered what to do if someone wanted to define a dictionary with interpolations as keys *outside* of an interpolation. To be honest I didn't spend much time thinking about it and decided just not to support it for now out of simplicity (thus keeping the assert below the comment).

hydra/grammar/OverrideLexer.g4
odelalleau: All modes are entirely independent, if you want to re-use a rule from another mode it has to be declared again (under a different name)

news/976.feature
odelalleau: Done in 416abb1b1

tests/test_overrides_parser.py
odelalleau: Good idea, done in 044a006f2

omry:
cool , before I merge - can you squash the news file addition to the previous commit?

omry:
in fact, make it two commits:
1. everything
2. the version bump

hydra/grammar/OverrideParser.g4
odelalleau: True, my big list of tests from OmegaConf doesn't have examples of nested interpolations. However I ran into an issue when adding some -- will need more time to look into it (this could have to wait until next week).

hydra/grammar/OverrideParser.g4
odelalleau: Well I don't see a reason why it would make more sense to allow interpolations in values compared to keys, for a start.

First thing I can think of (probably not the most complete/simple/convincing), let's say there are different folders, each containing a known type of files, and each file must be processed by a function that depends on its type. And let's say we want to perform some kind of reduce operation over the processed files, whose output is saved in the config variable `reduced`. It may look like this:

```yaml
images:
    path: /path/to/images
    process_function: process_image
videos:
    path: /path/to/videos
    process_function: process_video
captions:
    path: /path/to/captions
    process_function: process_caption
reduced: ${process:{${images.path}: ${images.process_function}, ${videos.path}: ${videos.process_function}, ${captions.path}: ${captions.process_function}}}
```

The code for the `process` resolver would be something like:

```python
def process(**to_process):
    return some_reduce_operation(f(os.path.join(path, x)) for path, f in to_process.items() for x in os.listdir(path))
```

hydra/core/override_parser/overrides_visitor.py
omry: 1. Since I didn't really get to review it yet, I was not aware of this intention. can you show me a convincing use case for it?
2. The assert is the answer to my second question. the comment like an unrelated piece of information. I did not realize the assert was really about enforcing it.

omry (changes requested):
1. I think we should drop interpolation in keys. this is way too crazy.
The key in a hashtable should be fixed. I am not sure how it can even work when an action outside of the key can change it at runtime (what happens to the hash of the key at that point in the hashtable which is map?)
You are welcome to change my mind.

2. Please squash the commits as requested (or alternatively undo the version bump, I can do it separately).

odelalleau:
> 1. I think we should drop interpolation in keys. this is way too crazy.
>    The key in a hashtable should be fixed. I am not sure how it can even work when an action outside of the key can change it at runtime (what happens to the hash of the key at that point in the hashtable which is map?)
>    You are welcome to change my mind.

I won't try too hard to change your mind, since anyway a dict can always be replaced with a list. But your comment confuses me and makes me wonder if we're talking about the same thing, so I'll try to clarify:

This is only for dictionaries that are passed as inputs to resolvers. These dictionaries are re-created on-the-fly on every access to the variable calling such a resolver. So we're never modifying in-place the keys of a dictionary.

hydra/grammar/OverrideParser.g4
omry: Okay, I will comment on the other thread. (easier to find).

hydra/grammar/OverrideParser.g4
omry: it could be in a followup. relative interpolations are very new in OmeaConf.

omry (commented):
Thanks for the use case.
One thing to remember is that the primary use case for dictionaries is to merge them into the config, where they will be accessible later.

```yaml
a:
  x: 10
  y: 20

# command line
a={a:{z:30}}

# result
a:
  x: 10
  y: 20
  z: 30
```
Now, assuming you have support for interpolation in keys, what does this do?
```yaml
a:
  x: 10
  y: 20
b: ???

# command line 1:
#
# this is evaluated and merged right after config composition, later modifications to
# a.x would not apply to it anymore, which is unexpected because interpolations 
# are normally evaluated lazily.

b: {foo_${a.x}: 30}

# result:
a:
  x: 10
  y: 20
b:
  foo_10: 30
```

Interpolations can also point to nodes (dicts or lists in the config), this is another things to consider - lists and dicts cannot be keys in OmegaConf configs.

```
a:
  x: 10
  y: 20
b: ???
# this is hopefully going to blow up with a nice error
b: {${a}: 10}
```

In short, before enabling interpolation into keys, I want tests covering the functionalities it enables.
In particular, OmegaConf.update() should be tested with dict values that has such keys to ensure it does the right thing, but there could be other areas that are affected.
If you want to try to support it, file an issue against OmegaConf and we can iterate on it independently from the rest of the changes.

odelalleau:
> One thing to remember is that the primary use case for dictionaries is to merge them into the config, where they will be accessible later.

Ah I see! So actually what you describe is not supported, trying to use an interpolation as key outside of an interpolation will fail on the assert I pointed out earlier (https://github.com/odelalleau/hydra/blob/044a006f24268a2991a0c37f1bcaca4a40e6621c/hydra/core/override_parser/overrides_visitor.py#L189)

As I mentioned above, _"this is only for dictionaries that are passed as inputs to resolvers"_.

At its source, the problem lies on the assumption that it's ok to use the same grammar to parse command-line overrides that set the value of a config option and to parse resolver arguments. This is helpful to keep Hydra's and OmegaConf's grammar close to each other, but it's not surprising to hit a few edge cases where things don't match perfectly.

Right now I see 3 options:

1. Keep things as they are in the current version of this PR (as I said, the situation you mentioned is currently explicitly not supported)
2. Get rid of interpolations in keys of dictionaries passed as resolver arguments (side note: this has the added benefit of simplifying OmegaConf's code -- this is related to changes in my OmegaConf's PR that you haven't gone through yet)
3. Better separate Hydra and OmegaConf grammars so that they can both evolve separately

As of now my preference is for (1) since it means less work for me, but I don't feel strongly about it so let me know what you prefer.

omry:
> > One thing to remember is that the primary use case for dictionaries is to merge them into the config, where they will be accessible later.
> 
> Ah I see! So actually what you describe is not supported, trying to use an interpolation as dictionary key outside of an interpolation will fail on the assert I pointed out earlier (https://github.com/odelalleau/hydra/blob/044a006f24268a2991a0c37f1bcaca4a40e6621c/hydra/core/override_parser/overrides_visitor.py#L189)

Generally assertions are not a good way to tell users that they are doing something wrong, but that's besides the point.
Good that you (accidentally?) prevented that :).

> As I mentioned above, _"this is only for dictionaries that are passed as inputs to resolvers"_.
> 
> At its source, the problem lies on the assumption that it's ok to use the same grammar to parse command-line overrides that set the value of a config option and to parse resolver arguments. This is helpful to keep Hydra's and OmegaConf's grammar close to each other, but it's not surprising to hit a few edge cases where things don't match perfectly.
> Right now I see 3 options:
> 
> 1. Keep things as they are in the current version of this PR (as I said, the situation you mentioned is currently explicitly not supported)
> 2. Get rid of interpolations in keys of dictionaries passed as resolver arguments (side note: this has the added benefit of simplifying OmegaConf's code -- this is related to changes in my OmegaConf's PR that you haven't gone through yet)
> 3. Better separate Hydra and OmegaConf grammars so that they can both evolve separately
> 
> As of now my preference is for (1) since it means less work for me, but I don't feel strongly about it so let me know what you prefer.

Do you have a concrete use case for this feature or is this theoretical?
my preference is to introduce it separately (if if there are compelling use cases) and not as part of the initial OmegaConf grammar because I am not convinced that there are not unforeseen implications to adding it yet.

Let's go with 2 for now.

odelalleau:
> Let's go with 2 for now.

Alright, I'll make this PR as draft until I can change it and also look into that other issue related to relative interpolations. ETA sometime next week after ICLR deadline.

omry:
> > Let's go with 2 for now.
> 
> Alright, I'll make this PR as draft until I can change it and also look into that other issue related to relative interpolations. ETA sometime next week after ICLR deadline.

Sounds good. hopefully by that time I will have time to start going over your big PR at the OmegaConf repo.

odelalleau:
Pushed new version, should be ok now:
* Squashed commits
* Added test for relative interpolations (this is in `test_omegaconf_grammar.py`)
* Interpolations aren't allowed anymore as keys in dictionaries passed as resolver arguments

omry:
Now that Hydra is on top of OmegaConf 2.1.0dev20 that contains the nested interpolation support in OmegaConf, this is unblocked.
Looks like there are some conflicts.

When you think the time is right, we can work toward integrating this.

odelalleau:
> Now that Hydra is on top of OmegaConf 2.1.0dev20 that contains the nested interpolation support in OmegaConf, this is unblocked.
> Looks like there are some conflicts.
> 
> When you think the time is right, we can work toward integrating this.

Yes, I was planning to get back to it once my current work on OmegaConf would be finalized. There are definitely some changes to make to this PR, based on how things have landed in OmegaConf.

odelalleau:
Superseded by #1594 => closing

PR #989, opened 2020-09-19
566 recursive instantiate

### LINKED ISSUE #566: Recursive instantiation

### LINKED ISSUE #995: Document recursive instantiation once API stabilizes

omry:
- [x] Implement recursive instantiation
- [x] Clean ObjectConf and params support deprecated in Hydra 1.0.
- [x] Change API to default to recursive
- [x] Update documentation to reflect changes
  - [x] Explain param overrides and how it works with nested instantiations.
  - [x] Add examples
  - [x] ~Add an upgrade page explaining change for default recursive~ (no need for an upgrade page).

Closes #566
Closes #995


omry:
#995 : documentation followup

CatOfTheCannals:
Hi!
Will this new feature be released to a 1.0.4 or to a 1.1.0 version?
Will it be available soon?

omry:
Hi @CatOfTheCannals, 
As you can tell by looking at the diff size, it's pretty big.
It will be released with Hydra 1.1, it will be available in master and later in dev or rc releases before Hydra 1.1 gets released.

hydra/utils.py
```diff
@@ -47,31 +50,31 @@ def call(config: Any, *args: Any, **kwargs: Any) -> Any:
...
 
     cls = "<unknown>"
     try:
-        assert isinstance(config, DictConfig)
+        assert OmegaConf.is_config(config)
         OmegaConf.set_readonly(config, False)
         OmegaConf.set_struct(config, False)
+        config._set_flag("allow_objects", True)
         cls = _get_cls_name(config)
         type_or_callable = _locate(cls)
```
ppwwyyxx, line 65: Thanks for supporting objects! Since now any object can be put in the dict, can you allow `_target_` to be the callable object itself?

omry: This is potentially doable but out of scope for this diff.
Can you open a new feature request and show an example where this would be preferred over passing the class/function name?

alexcoca:
@omry I don't know if I should open a feature request or if this is already possible, but suppose you have a complex hierarchy where most objects have to be recursively instantiated but for some we wish to defer the instantiation so that it can be managed inside the program. For example, if we wish to run large models sequentially on the same GPU, we wish to instantiate the second once we are done with the first. I had a feeling that `_recursive_` special key could be used to achieve this so tried this configuration 


```yaml
bleurt:
    _recursive_: false
    _target_: datasets.load_metric
```

but I'm getting the error

```bash
Error in call to target 'datasets.load.load_metric':
TypeError("load_metric() missing 1 required positional argument: 'path'")
full_key: architecture.paraphrase_decoder.semantic_faithfulness_model

Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.```

which suggests that `hydra` does not care about `_recursive_` . I definitely can't pass `_recursive_` false to the the instantiation call for the entire object because the structure is complex and it would take a huge refactor to achieve this. Is this feature not supported?

**Comment by Jasha10** [comment]
@alexcoca I think you will need to pass `_recursive_: false` to the config that is one level above `bleurt`:
```yaml
foo:
  _target_: bar
  _recursive_: false
  bleurt:
    _target_: datasets.load_metric
```
With this setup, I believe `foo` will be instantiated and `bleurt` will not.

alexcoca:
@Jasha10 Yes, this was the only way to get things working. I removed `bleurt` and created a dummy function that returned `datasets.load_metric` as a work around as, `bleurt` was sadly at the same of the hierarchy with a bunch of objects that did need to be instantiated so I could not have `_recursive_` at the level above. I'm happy enough with this workaround ... Having looked through the source code I understand that the logic is to avoid instantiating children of nodes where `_recursive_`: false is specified. Obviously, this would require that in my case the parent of `bleurt` specified `_recusive_`: false, as you suggest.  Thanks for your help, I learned a lot.

Issue #1001, opened 2020-09-22
[Bug] String interpolation in instantiate is broken in hydra 1.0.2


cmnrd:
# 🐛 Bug
## Description

String interpolation does not work when the config node is used in instantiate().

## Checklist
- [x] I checked on the latest version of Hydra
- [x] I created a minimal repro

## To reproduce

** Minimal Code/Config snippet to reproduce **

demo.py:
```
import hydra

class Test:

    def __init__(self, string):
        print("init: " + string)

@hydra.main(config_path='conf', config_name='test')
def main(cfg):
    print("main: " + cfg['test']['string'])
    hydra.utils.instantiate(cfg['test'])

if __name__ == "__main__":
    main()
```

conf/test.yaml:
```
# @package _global_
foo: 42
test:
  _target_: demo.Test
  string: "21 * 2 = ${foo}"
conf/test.yaml (END)
```

** Stack trace/error message **

No error is produced, but the output printed by the code above is erroneous when using hydra 1.0.2 or the latest master.
```
main: 21 * 2 = 42
init: 21 * 2 = ${foo}
```
When referencing `test.string` directly from the python code, `foo` gets resolved correctly to `42` (first line). However, when passing the interpolated string via `instantiate()` to the class `Test`, `${foo}` does not get resolved (second line).

## Expected Behavior

The expected output is:
```
main: 21 * 2 = 42
init: 21 * 2 = 42
```
This output was produced with hydra 1.0.1.

## System information
- **Hydra Version** :  1.0.2
- **Python version** : 3.8.5
- **Virtual environment type and version** :  virtualenv 20.0.23
- **Operating system** : Linux (Manjaro)

omry:
Thanks for reporting, is this a regression in 1.0.2? (does it work with 1.0.1)?
I saw you created a PR and closed, what's going on with that?

cmnrd:
Yes, it is a regression in 1.0.2. It works perfectly in 1.0.1. The 'Expected Behavior' above is the output from hydra 1.0.1.

I have no idea about this PR. It was opened and closed by someone else (@mipo57).

omry:
> Yes, it is a regression in 1.0.2. It works perfectly in 1.0.1. The 'Expected Behavior' above is the output from hydra 1.0.1.
Thanks.

> I have no idea about this PR. It was opened and closed by someone else (@mipo57).

Interesting :)

omry:
[...]

PR #1002, opened 2020-09-22
Fix interpolation not resolved on instantiation

mipo57:
#  Motivation

Fix issue #1001 

Yes

## Test Plan

(Taken from original Issue)

demo.py:
```python
import hydra

class Test:

    def __init__(self, string):
        print("init: " + string)

@hydra.main(config_path='conf', config_name='test')
def main(cfg):
    print("main: " + cfg['test']['string'])
    hydra.utils.instantiate(cfg['test'])

if __name__ == "__main__":
    main()
```

conf/test.yaml:
```yaml
# @package _global_
foo: 42
test:
  _target_: demo.Test
  string: "21 * 2 = ${foo}"
```

Result without fix:
```bash
main: 21 * 2 = 42
init: 21 * 2 = ${foo}
```

Without the fix, when referencing test.string directly from the python code, foo gets resolved correctly to 42 (first line). However, when passing the interpolated string via instantiate() to the class Test, ${foo} does not get resolved (second line).

## Related Issues and PRs

Issue #1001

### DIFF
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -595,7 +595,7 @@ def _get_kwargs(config: Union[ObjectConf, DictConfig], **kwargs: Any) -> Any:
     with read_write(params):
         params.merge_with(config_overrides)
 
-    for k, v in params.items_ex(resolve=False):
+    for k, v in params.items_ex(resolve=True):
         if k == "_target_":
             continue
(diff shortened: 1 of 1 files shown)


omry:
Thanks for trying, this was a bit dicier than that.
look at #1006 and my next comment on #1001.

PR #1006, opened 2020-09-22
Fixed interaction between interpolation and instantiate

### LINKED ISSUE #1001: [Bug] String interpolation in instantiate is broken in hydra 1.0.2

omry:
This includes a nasty bug in the tests that prevented the detection of this and other regressions.
Closes #1001

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -28,11 +28,10 @@ def module_function(x: int) -> int:
 
 @dataclass
 class AClass:
-    def __init__(self, a: Any, b: Any, c: Any, d: Any = "default_value") -> None:
-        self.a = a
-        self.b = b
-        self.c = c
-        self.d = d
+    a: Any
+    b: Any
+    c: Any
+    d: Any = "default_value"
 
     @staticmethod
     def static_method(z: int) -> int:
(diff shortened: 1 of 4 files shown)


tests/__init__.py
```diff
@@ -28,11 +28,10 @@ def module_function(x: int) -> int:
...
 class AClass:
-    def __init__(self, a: Any, b: Any, c: Any, d: Any = "default_value") -> None:
-        self.a = a
-        self.b = b
-        self.c = c
-        self.d = d
+    a: Any
+    b: Any
+    c: Any
+    d: Any = "default_value"
```
omry, line 34: dataclass eq before just returned True because it did not see that there are any fields to compare.

PR #1020, opened 2020-09-29
bump xcode version to get newer MacOS

omry:


.circleci/config.yml
```diff
@@ -98,7 +98,7 @@ jobs:
       py_version:
         type: string
     macos:
-      xcode: "10.0.0"
+      xcode: "12.0.0"
```
jieru-hu, line 101: probably need to update `cache_key_version:`, since master reuse the caching (the cache key is per branch)

PR #1025, opened 2020-09-30
Run conda on linux as well for circleCI

jieru-hu:
## Motivation
Previously we pin the circleCI linux docker image to be python:3.8|3.7|3.6 . However, the python version it ends up with seems to be fluid, for example  python:3.8 lands on python 3.8.6 which is not yet available to be installed in conda.

To be consistent with how tests are run in MACOS and WIN, I added the miniconda installation for linux machines as well. The installation doesn't take long, so I didn't add cache for it.

Yes/No

## Test Plan

CI should pass

## Related Issues and PRs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,7 +115,6 @@ jobs:
           name: Testing Hydra
           no_output_timeout: 10m
           command: |
-            python -c 'import sys; print(sys.version_info[:])'
             export NOX_PYTHON_VERSIONS=<< parameters.py_version >>
             conda activate hydra
             pip install nox
@@ -134,7 +133,6 @@ jobs:
           name: Testing Hydra
           command: |
             export PATH="$HOME/miniconda3/envs/hydra/bin:$PATH"
-            python -c 'import sys; print(sys.version_info[:])'
             export NOX_PYTHON_VERSIONS=<< parameters.py_version >>
             pip install nox
             pip install dataclasses
@@ -181,7 +179,6 @@ jobs:
           name: << parameters.test_plugin >>
           no_output_timeout: 10m
           command: |
-              python -c 'import sys; print(sys.version_info[:])'
               export NOX_PYTHON_VERSIONS=<< parameters.py_version >>
               export PLUGINS=<< parameters.test_plugin >>
               conda activate hydra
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -176,35 +181,34 @@ jobs:
...
           command: |
-                export NOX_PYTHON_VERSIONS=<< parameters.py_version >>
-                export PLUGINS=<< parameters.test_plugin >>
-                pip install nox
-                pip install dataclasses
-                nox -s lint_plugins test_plugins -ts
-
-
+            export PATH="$HOME/miniconda3/envs/hydra/bin:$PATH"
+            python -c 'import sys; print(sys.version_info[:])'
```
omry, line 206: The printout does not seem super useful.
remove?

jieru-hu: done :)

.circleci/config.yml
```diff
@@ -111,26 +120,23 @@ jobs:
...
             nox -s lint test_tools test_core test_jupyter_notebooks -ts
-
-
   test_linux:
     parameters:
       py_version:
         type: string
     docker:
-      - image: circleci/python:<< parameters.py_version >>
+      - image: circleci/python:3.8
```
omry, line 128: at this point we should probably use a vanilla ubuntu image and not a Python one, right?

jieru-hu: it seems like they don't have 18.04, only either 20.04 (beta) or 16.04 (https://circleci.com/docs/2.0/configuration-reference/#available-machine-images)

we could use  `cimg/base:stable-18.04` docker image as an alternative (https://discuss.circleci.com/t/ubuntu-20-04-next-gen-convenience-images/36604)

omry: Interesting. Let's try ubuntu-16.04 until the 20.04 comes out of beta?

jieru-hu: Sure. I'm curious why you prefer the machine to docker? Any specific concern?

omry: Nothing major. less moving parts.
If you want to try a docker image I am fine with it, but I am not sure what the implications are.
Pay attention to startup and runtime speeds if you do.

omry: My initial wrong perception was that the cimg image is a third party one.

jieru-hu: ah i see, yea, the docker image is from circleCI :)

PR #1035, opened 2020-10-02
[circleCI] run plugin pytest verbose

jieru-hu:
## Motivation

run nox session by default `silent=False`

This helps us see the progress and prevents timeouts on circleCI.

Yes/No

## Test Plan
CI
check to see how much logging we are generating.

## Related Issues and PRs

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -34,8 +34,6 @@ SKIP_CORE_TESTS = os.environ.get("SKIP_CORE_TESTS", SKIP_CORE_TESTS) != "0"
 FIX = os.environ.get("FIX", "0") == "1"
 VERBOSE = os.environ.get("VERBOSE", "0")
 SILENT = VERBOSE == "0"
-PYTEST_VERBOSE = os.environ.get("PYTEST_VERBOSE", "0")
-PYTEST_SILENT = PYTEST_VERBOSE == "0"
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -32,7 +32,7 @@
 SKIP_CORE_TESTS = "0"
 SKIP_CORE_TESTS = os.environ.get("SKIP_CORE_TESTS", SKIP_CORE_TESTS) != "0"
 FIX = os.environ.get("FIX", "0") == "1"
-VERBOSE = os.environ.get("VERBOSE", "0")
+VERBOSE = os.environ.get("VERBOSE", "1")
```
omry, line 35: can you just limit this to the place where you want more output, specifically to when we call pytest instead of changing it for everything?

omry (changes requested):
see my comment

noxfile.py
jieru-hu: sounds good!

omry: Also : even if we decided that we want to make VERBOSE true for CI (and I don't think this is what we are talking about), you should do with with the environment variable in the CircleCI config and not by changing the default like this.

jieru-hu:
SILENT output from running hydra-core tests https://app.circleci.com/pipelines/github/jieru-hu/hydra/645/workflows/8d9a6969-a01a-4b33-abe0-3f726bba5ca7/jobs/6958

VERBOSE plugin test for pytest https://app.circleci.com/pipelines/github/jieru-hu/hydra/643/workflows/2c422955-9823-4c93-9e82-bc8c6fe30337/jobs/6955

noxfile.py
```diff
@@ -96,7 +99,7 @@ def pytest_args(*args):
 
 def run_pytest(session, directory=".", *args):
     pytest_cmd = pytest_args(directory, *args)
-    session.run(*pytest_cmd, silent=SILENT)
+    session.run(*pytest_cmd, silent=PYTEST_SILENT)
```
omry, line 102: I think this is enough for now, no need for the rest.
put a comment next to it explaining why.

```suggestion
    session.run(*pytest_cmd, silent=False)
```

jieru-hu: Sure. I thought in your earlier comment you meant only changing it for circleCI tests, guess i misunderstood that :)

PR #1037, opened 2020-10-06
updateNews.md

gujralsanyam22:
there are some issues which i think that it is fixed and also and if it is fixed it looks good


omry:
Thanks, but most of your corrections are wrong (changing capitalization of function and types).
I also do not thing this is a good investment of my time (or yours), so I am closing this PR.

PR #1044, opened 2020-10-07
Recursive Defaults List support

### LINKED ISSUE #171: Recursive defaults support
Hydra does not currently support a recursive defaults list. the defaults are only specified at the primary config file.
There is some special support for Hydra itself, but this can be ignored for the purpose of this issue.

Recursive defaults list are tricky, it's important to ensure intuitive and consistent order of merging.

### Proposed logic

if the defaults in the primary file are in that order:
main.yaml
```yaml
defaults:
  - aa: x1
  - bb: x2
```

The relative order should be preserved. it is however possible for other files to be injected in between aa and bb, for example if aa/x1 had it's own defaults:

aa/x1.yaml
```yaml
defaults:
  - cc: x3
  - dd: x4
```

the resulting defaults list might looks like:
```yaml
  - aa: x1
  - cc: x3  # from aa/x1 
  - dd: x4  # from aa/x1 
  - bb: x2
```

Let's look at a more complicated case:

main.yaml
```yaml
defaults:
  - aa: x1
  - bb: x2
  - cc: must_win
```
aa/x1.yaml
```yaml
defaults:
  - cc: x3
  - dd: x4
```
In this case, the result should be:
the resulting intermediate list will looks like:
```yaml
  - aa: x1
  - cc: x3
  - dd: x4
  - bb: x2
  - cc: must_win
```
but the final result will be:
```yaml
  - aa: x1
  - cc: must_win
  - dd: x4
  - bb: x2
```

Note how cc retains it's order in the list, but got the last value.

The rule here is:
1. Composition order determined by the first time a configuration group is introduced (the `db` in `db=mysql`).
2. Composition choice (the `mysql` in `db=mysql`) is determined by the last mention of the group).

### LINKED ISSUE #326: Ability to specify composition order of primary config file
There are multiple use cases for having configuration in the users config:

Use it as scaffolding and use defaults to populate it:
```yaml
defalts:
  - foo: bar

foo:
   a: ???
   b: ???
```

In the above case foo/bar.yaml can be used to populate a and b.

An alternative approach is to allow the user to override things populated by the defaults from the config file:
```yaml
defalts:
  - foo: bar

foo:
   a: 10
   b: 20
```

In this scenario, the user wants to force foo.a and foo.b to specific values.

Those two approaches are conflicting, each one requires a different composition order.
Currently, the supported mode is the first one and the second is not possible.
Composition order is roughly:

- hydra related configs
- user config file
- user provided defaults

The idea is to support explicit specification of when to merge in the users config via a special keyword in the defaults, like `_self_`.
for example:

```yaml
defalts:
  - _self_
  - foo: bar
```
or:
```yaml
defalts:
  - foo: bar
  - _self_
```
In the first case, the config file would be composed before foo/bar, in the second one after.

### LINKED ISSUE #1080: Changes in ConfigSource Plugins
#171 requires changes to the implementation of all config sources.
This means that existing config source plugins needs to be changed.

In particular, ConfigResult now contains the default list which should be extracted (and removed) from the config object.
Below is updated code adding the defaults list to the ConfigResult in a ConfigSource:

```python
            defaults_list = self._extract_defaults_list(config_path=config_path, cfg=cfg)
            return ConfigResult(
                config=self._embed_config(cfg, header["package"]),
                path=f"{self.scheme()}://{self.path}",
                provider=self.provider,
                header=header,
                defaults_list=defaults_list,
            )
```

### LINKED ISSUE #1089: Standardize syntax for deletion from defaults list
There are two forms to delete from the defaults list, and both can be used in the defaults list and as config overrides:
### form 1:
command line : group=null
config file:
```yaml
defaults:
  - group: null
```

### form 2:
command line: ~group, ~group=value
(the first form is equivalent to group=null, the second will succeed only if group is currently set to value.

config file:
```yaml
defaults:
  - ~group
  - ~group: value
```

- Hydra 1.0 deprecated the first form when used in the command line, but failed to deprecate it in the defaults list.
- Hydra 1.1 will remove support for this form from the command line and deprecated it in favor of the second form in the config files.
- Hydra 1.2 will remove support of the first from from the config files as well.

omry:
Closes #171
Closes #326
Closes #1080
Closes #1089

- [x] Add `_self_` support (#326)
- [x] Support recursive defaults (with all the existing features, deletion, package rename, overriding)
- [x] Clean new TODOs in the code
- [x] Deprecated `- group: null` deletion form from defaults list in favor of `~group` and `~group=value`.
- [x] Ensure no performance regression (will likely require caching of loaded configs because the new system does two passes.
- [x] Include information about actually used defaults list in the generated Hydra config node.
- [X] Consider changing default for unspecified `_self_` to be last in the list and not first as it is now
  - Decided against. While this is arguably more intuitive for config files it's less intuitive for dataclasses with defaults list.
- [x] Document recursive defaults (probably warrants a dedicated page just for the defaults list).
- [x] Document `_self_`
- [x] Deprecated old way of using interpolation in the defaults list. specifically the pattern in [specializing configs](https://hydra.cc/docs/patterns/specializing_config) will be removed in the future in favor of the simpler but more magical new style. Old style is still supported in 1.1 but will be removed in 1.2.

This:
```yaml
defaults:
  - dataset: imagenet
  - model: alexnet
  - dataset_model: ${defaults.0.dataset}_${defaults.1.model}
    optional: true
```
Will become:
```yaml
defaults:
  - dataset: imagenet
  - model: alexnet
  - dataset_model: ${dataset}_${model}`
    optional: true
```

### DIFF
--- a/examples/advanced/automatic_config_schema/conf/config.yaml
+++ b/examples/advanced/automatic_config_schema/conf/config.yaml
@@ -1,0 +1,2 @@
+defaults:
+  - db: mysql
--- a/examples/advanced/automatic_config_schema/conf/db/mysql.yaml
+++ b/examples/advanced/automatic_config_schema/conf/db/mysql.yaml
@@ -1,0 +1,10 @@
+# @package _group_
+defaults:
+  - schema/db/mysql
+  - _self_
+
+host: localhost
+port: 3306
+driver: mysql
+user: omry
+password: secret
--- a/examples/advanced/automatic_config_schema/conf/db/postgresql.yaml
+++ b/examples/advanced/automatic_config_schema/conf/db/postgresql.yaml
@@ -1,0 +1,11 @@
+# @package _group_
+defaults:
+  - schema/db/postgresql
+  - _self_
+
+host: localhost
+port: 5432
+driver: postgresql
+user: postgre_user
+password: secret
+timeout: 10
--- a/examples/advanced/automatic_config_schema/my_app.py
+++ b/examples/advanced/automatic_config_schema/my_app.py
@@ -1,0 +1,40 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+from dataclasses import dataclass
+
+from omegaconf import DictConfig, OmegaConf
+
+import hydra
+from hydra.core.config_store import ConfigStore
+
+
+@dataclass
+class MySQLConfig:
+    host: str
+    port: int
+    user: str
+    password: str
+    driver: str = "mysql"
+
+
+@dataclass
+class PostGreSQLConfig:
+    host: str
+    port: int
+    timeout: int
+    user: str
+    password: str
+    driver: str = "postgresql"
+
+
+cs = ConfigStore.instance()
+cs.store(group="schema/db", name="mysql", node=MySQLConfig, package="db")
+cs.store(group="schema/db", name="postgresql", node=PostGreSQLConfig, package="db")
+
+
+@hydra.main(config_path="conf", config_name="config")
+def my_app(cfg: DictConfig) -> None:
+    print(OmegaConf.to_yaml(cfg))
+
+
+if __name__ == "__main__":
+    my_app()
--- a/examples/patterns/specializing_config/conf/config.yaml
+++ b/examples/patterns/specializing_config/conf/config.yaml
@@ -1,5 +1,5 @@
 defaults:
   - dataset: imagenet
   - model: alexnet
-  - dataset_model: ${defaults.0.dataset}_${defaults.1.model}
+  - dataset_model: ${dataset}_${model}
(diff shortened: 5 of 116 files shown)


nmichlo:
Does this mean that the new style will also support nested variables for specialization? Or rather would the following work if I were to extend your last example?
```
defaults:
  - dataset: imagenet
  - model: alexnet
  - data_wrapper: ${dataset.data_type}_${model.data_input_mode}
```

omry:
No. the defaults list will never be able to use config variables in interpolation.
The defaults list is used to compose final config - it obviously cannot access things in the config that is not there yet to determine how to compose the same config.

omry:
FYI: #1170 re-implements this and will be merged soon.
There are some major changes relative to this implementation there.

PR #1060, opened 2020-10-14
Set `USER` env variable only if it is not already set.

### LINKED ISSUE #1059: [Bug] os.environ["USER"] is set unconditionally in hydra/hydra/test_utils/test_utils.py
Since os.environ["USER"] is set and hydra-core installs a pytest plugin any other pytest tests that use the $USER environment variable will get overridden to use "test_user" as the user. I think this should probably be a conditional check or something. We had a test start failing that was using $USER and it took some time to figure out why that was being set to test_user.

shagunsodhani:
Fixes #1059.

## Motivation

Set `USER` env variable only if it is not already set.

Yes

## Test Plan

CI

## Related Issues and PRs

#1059

### DIFF
--- a/tests/test_examples/test_tutorials_basic.py
+++ b/tests/test_examples/test_tutorials_basic.py
@@ -295,9 +295,9 @@ def test_advanced_ad_hoc_composition(
         "hydra.run.dir=" + str(tmpdir),
     ]
     result, _err = get_run_output(cmd)
+    assert OmegaConf.create(result) == OmegaConf.create(expected)
     if should_remove_variable:
         del os.environ["USER"]
-    assert OmegaConf.create(result) == OmegaConf.create(expected)
 
 
 def test_examples_using_the_config_object(tmpdir: Path) -> None:
--- a/tests/test_examples/test_tutorials_basic.py
+++ b/tests/test_examples/test_tutorials_basic.py
@@ -286,13 +286,17 @@ def test_advanced_ad_hoc_composition(
 ) -> None:
 
     # CircleCI does not have the environment variable USER.
+    should_remove_variable = False
     if "USER" not in os.environ:
         os.environ["USER"] = "test_user"
+        should_remove_variable = True
     cmd = [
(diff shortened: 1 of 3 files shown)


omry (changes requested):
can you remove this and see which tests are breaking? we can probably just mock the environment variable on the specific tests.

shagunsodhani:
Will get to it in a couple of hours.

shagunsodhani:
ahh I think I should be explicitly unsetting the `USER` env variable if it was not originally set. I was earlier relying on the python context but then I realized that `os.environ` is a global object.

tests/test_examples/test_tutorials_basic.py
```diff
@@ -283,12 +284,20 @@ def test_sweeping_example(
 def test_advanced_ad_hoc_composition(
     tmpdir: Path, args: List[str], expected: Any
 ) -> None:
+
+    # CircleCI does not have the environment variable USER.
+    should_remove_variable = False
```
omry, line 289: Use monkeypatch.
```python
def test_advanced_ad_hoc_composition(
    minkeypatch: Any, tmpdir: Path, args: List[str], expected: Any
) -> None:
  monkeypatch.setenv("USER", "test_user")
  ...
```

omry (changes requested):

omry:
Also:
1. Add a news fragment.
2. Cherry pick to the 1.0 branch.

tests/test_examples/test_tutorials_basic.py
shagunsodhani: Ohh sweet :)

shagunsodhani:
> Also:
> 
> 1. Add a news fragment.
> 2. Cherry pick to the 1.0 branch.

For 1, isnt this an internal change? I mean the end-user doesnt (

shagunsodhani:
oops sorry - accidentally closed it :)

omry:
We do, if anything we do not update news in master.
I am planning to eventually release 1.0.4 with some minor patches.

PR #1084, opened 2020-10-20
[RFC] - Hydra callbacks

jieru-hu:
## Motivation

Hydra 1.1, selecting multiple config groups might help here https://hydra.cc/docs/next/patterns/select_multiple_configs_from_config_group

#### Update 11/16
Synced up with Omry offline on this, updated the PR to address some high-level comments from omry, and also some thoughts I have regarding if Callbacks should be plugins.

------
>Why we need 3 levels of callbacks (job, run, sweep)

In Hydra there's two mode: RUN and MULTIRUN, MULTIRUN is triggered by adding a `-m` from the command line. if no `-m` is present, then Hydra will be in RUN mode.

**RUN**
run -> call `run_job` once with the config, one Hydra RUN is one job.

**MultiRUN**
multirun -> sweeper.sweep() -> launcher.launch() -> launch_jobs(list_of_job_configs) -> call `run_job` multiple times with configs generated from sweeper.

In general,  `Job` is the smallest unit in Hydra, each Job is a single run with one set of config. Each job has its own logging dir created.

Other than number of jobs launched is different, MULTIRUN also uses different configs (sweep_config) from RUN. 

Due the differences, I suggest we have 3 levels of callbacks:
For single run, we provide callbacks before/after the RUN (`pre_run`, `post_run`) and before/after each job (`pre_job`, `post_job`).
For multirun, we provide callbacks before/after the MULTIRUN (`pre_sweep`, `post_sweep`) and before/after each job (`pre_job`, `post_job`)

**Why do we need both `pre_run` and  `pre_sweep`? Why cannot we merge `pre_run` and `pre_sweep`?**
RUN and MULTIRUN assumes different configs: `run.dir` and `sweep.dir`, Hydra does not know which one to use until at run time. As a result, when we implement Hydra callback, it makes sense to separate `run` and `sweep`.

Some example use case:
- for example in MLflow's case, user would only need to override `pre_sweep` and `pre_run` to set the MLFLOW_TRACKING_URL
- It is common for researchers to parse experiment result (or upload experiment results somewhere) after jobs finishes. In that case, users can override `post_run`/`post_sweep` for everything or `post_job` for each job depends on the needs.
---------

> How does callback work with Launcher.

With the current implementation, callbacks are decoupled from Launcher.

------------
> How we support Hydra callbacks and user implemented ones.

Suggest we create Hydra supported implementation in `hydra.callback`
Users can configure the callbacks they want to call in their application config, for example, 

```commandline
hydra:
  callbacks:
    example:
      _target_: "example_callback.ExampleCallback"
    mlflow:
      _target_: "hydra.callback.mlflow.MLflow"
```

-------
> Should we have callback as plugins? 

I'm leaning towards yes. The most important reason is so that users do not need to copy paste Hydra callbacks's config in their application config if they do not want to override anything.
However, I suspect supporting callback as plugins require Hydra supporting list of config groups

If we have Callbacks as plugins,  users could define callbacks as follows, 

```
hydra:
  callbacks:
    example:
      _target_: "example_callback.ExampleCallback"
      foo: bar
  
  callbacks_order:
       - example
       - mlflow
```
In the above example, we have `MLFlow` installed as a plugin, `example` is a user implemented Callback.
Users configure the order and which plugins they want to invoke by overriding `hydra.callbacks_order`. If they do not override this config, then only callbacks defined under `hydra.callbacks` will be called.
To achieve this would require us having access to Mlflow's config `hydra.callbacks.mlflow` after Hydra composes the configs, but I don't think this is supported yet. 

---------------------
#### update 11/04
Took another pass in the implementation & also addressed shagun's comments.

```
$ python /Users/jieru/workspace/hydra-fork/hydra/examples/callbacks/my_app.py +db=1 -m
[2020-11-04 13:34:45,070][HYDRA] ExampleCallbacks pre_sweep called
[2020-11-04 13:34:45,070][HYDRA] mlflow callbacks pre-sweep, setting tracking url
[2020-11-04 13:34:45,230][HYDRA] Launching 1 jobs locally
[2020-11-04 13:34:45,230][HYDRA]        #0 : +db=1
[2020-11-04 13:34:45,421][HYDRA] ExampleCallbacks pre_job called
foo: bar
db: 1

[2020-11-04 13:34:45,559][example_callbacks][INFO] - ExampleCallbacks post_job called
[2020-11-04 13:34:45,562][HYDRA] ExampleCallbacks post_sweep called

```
verified `.mlflow` dir was created with logging populated.

------------------------------------------------------------------------------------------------

RFC  

This is to follow up on the design doc we've been having discussions on, I want to get some early feedback on my thoughts on implementation. 

**MULTIRUN**
```
$ python /Users/jieru/workspace/hydra-fork/hydra/examples/callbacks/my_app.py +db=1,2,3 -m
[2020-10-20 16:40:05,188][HYDRA] ExampleCallbacks pre_sweep called
[2020-10-20 16:40:05,475][HYDRA] Launching 3 jobs locally
[2020-10-20 16:40:05,475][HYDRA]        #0 : +db=1
[2020-10-20 16:40:05,573][HYDRA] ExampleCallbacks pre_job called
hello: world
db: 1

[2020-10-20 16:40:05,609][example_callbacks][INFO] - ExampleCallbacks post_job called
[2020-10-20 16:40:05,610][HYDRA]        #1 : +db=2
[2020-10-20 16:40:05,707][HYDRA] ExampleCallbacks pre_job called
hello: world
db: 2

[2020-10-20 16:40:05,744][example_callbacks][INFO] - ExampleCallbacks post_job called
[2020-10-20 16:40:05,745][HYDRA]        #2 : +db=3
[2020-10-20 16:40:05,835][HYDRA] ExampleCallbacks pre_job called
hello: world
db: 3

[2020-10-20 16:40:05,869][example_callbacks][INFO] - ExampleCallbacks post_job called
[2020-10-20 16:40:05,870][HYDRA] ExampleCallbacks post_sweep called
(pytest38) jieru-mbp:hydra jieru$ 

```
**RUN**
```
$ python /Users/jieru/workspace/hydra-fork/hydra/examples/callbacks/my_app.py
[2020-10-20 16:38:48,746][HYDRA] ExampleCallbacks pre_run called
[2020-10-20 16:38:48,746][HYDRA] ExampleCallbacks pre_job called
hello: world

[2020-10-20 16:38:48,781][example_callbacks][INFO] - ExampleCallbacks post_job called
[2020-10-20 16:38:48,781][example_callbacks][INFO] - ExampleCallbacks post_run called

```

Yes/No

## Test Plan

## Related Issues and PRs


examples/callbacks/example_callbacks.py
```diff
@@ -0,0 +1,33 @@
+import logging
+
+from omegaconf import DictConfig
+
+from hydra.extra.callbacks import Callbacks
```
shagunsodhani, line 5: Is this namespace decided? would `hydra.callabacks` be better/simpler?

hydra/conf/__init__.py
```diff
@@ -120,6 +120,9 @@ class HydraConf:
     # Launcher configuration
     launcher: Any = MISSING
 
+    # Callbacks configuration
+    callbacks: Optional[Any] = None
```
shagunsodhani, line 124: callbacks:Any = None

hydra/extra/callbacks.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+
+    @abstractmethod
+    def setup(
+        self,
+        config: DictConfig,
+    ) -> None:
+        raise NotImplementedError()
+
+    def pre_run(self, cfg: DictConfig) -> None:
```
shagunsodhani, line 16: We should document the diffrence between `run`, `sweep` and `job`

examples/callbacks/config.yaml
```diff
@@ -0,0 +1,9 @@
+hydra:
+  callbacks:
+    example:
+      _target_: "example_callbacks.ExampleCallbacks"
+
+    order:
+      - example
```
shagunsodhani, line 7: How about we just follow the order as specified by `callbacks` node and not add an extra node?

jieru-hu: yea, I think that makes more sense.

hydra/extra/callbacks.py
jieru-hu: Agreed!

omry:
sorry for not commenting, I do not have the cycles to look at this area right now.

jieru-hu:
@omry @shagunsodhani 
I updated the PR and the description at the top. I'd appreciate some feedback from you :)
Also happy to chat in person if that's easier.

Thanks!

hydra/callback/mlflow.py
```diff
@@ -0,0 +1,24 @@
...
+log = logging.getLogger(__name__)
+
+
+class MLflow(Callback):
+    def __init__(self, tracking_uri: str):
+        self.tracking_uri = tracking_uri
+
+    def _set_tracking_uri(self):
+        log.info(f"Setting mlflow tracking URI to be {self.tracking_uri}")
+        mlflow.set_tracking_uri(self.tracking_uri)
```
shagunsodhani, line 18: Is this an idempotent function? I think atleast some ml lifecycle plugins will complain if some metadata is set several times within the same process

jieru-hu: Good point! :) 
I will definitely test it more thoroughly once we finalized on the Callbacks interface and config. 
Also, this function should only be called once (either by `pre_run` or `pre_sweep`) during a Hydra app's lifecycle.

examples/callback/config.yaml
```diff
@@ -0,0 +1,9 @@
+hydra:
+  callbacks:
+    example:
+      _target_: example_callback.ExampleCallback
+    mlflow:
+      _target_: hydra.callback.mlflow.MLflow
+      tracking_uri: file://${hydra:runtime.cwd}/.mlflow
```
omry, line 7: Experiment with the idea that a callback can be provided by an external package.
This means that the package will be contributing to a config group like hydra/callbacks:

```
hydra/callbacks/
  mlflow.yaml
```

mlflow.yaml
```yaml
# @package _group_
mlflow:
  _target_: hydra.callback.mlflow.MLflow
  tracking_uri: file://${hydra:runtime.cwd}/.mlflow
```

User primary config may now add it:
```yaml
defaults:
  # note that this is not overriding a config group (because they are mutually exclusive for now).
  - hydra/callbacks/mlflow 
  - hydra/callbacks/something_else
```

The above is providing composition order, and hopefully that translates in the resulting config dictionary.

hydra/_internal/hydra.py
```diff
@@ -103,15 +104,23 @@ def run(
             run_mode=RunMode.RUN,
         )
         HydraConfig.instance().set_config(cfg)
+        callbacks = get_callbacks(cfg)
 
-        return run_job(
+        for c in callbacks:
+            c.pre_run(cfg)
```
omry, line 110: an exception in a callback would prevent the execution of other callbacks.
While it's okay to allow it to flow to let the user know, we should execute all callbacks first.

hydra/_internal/hydra.py
```diff
@@ -103,15 +104,23 @@ def run(
...
+        ret = run_job(
             config=cfg,
             task_function=task_function,
             job_dir_key="hydra.run.dir",
             job_subdir_key=None,
             configure_logging=with_log_configuration,
         )
 
+        for c in reversed(callbacks):
+            c.post_run(cfg)
```
omry, line 120: same exception handling logic.
this will repeat so you should probably introduce a function to actually loop and call the callbacks.

hydra/_internal/hydra.py
```diff
@@ -128,13 +137,24 @@ def multirun(
...
+            config_loader=self.config_loader,
+            task_function=task_function,
         )
         task_overrides = OmegaConf.to_container(cfg.hydra.overrides.task, resolve=False)
         assert isinstance(task_overrides, list)
-        return sweeper.sweep(arguments=task_overrides)
+        ret = sweeper.sweep(arguments=task_overrides)
+
+        for c in callbacks:
+            c.post_sweep(cfg)
```
omry, line 155: reverse, exceptions.

hydra/_internal/hydra.py
```diff
@@ -128,13 +137,24 @@ def multirun(
             run_mode=RunMode.MULTIRUN,
         )
         HydraConfig.instance().set_config(cfg)
+        callbacks = get_callbacks(cfg)
+
+        for c in callbacks:
+            c.pre_sweep(cfg)
```
omry, line 143: exceptions.

hydra/callback/mlflow.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def pre_sweep(self, cfg: DictConfig) -> None:
+        self._set_tracking_uri()
```
omry, line 21: We would want to provide information about he sweep to the mlflow session here.
In particular, all the jobs from a single sweep should be grouped somehow.

Take a look at mlflow and see what would a good way to do it be.
We will probably need to extract this information from the config.
To that end, we may want to also pass the Hydra config node to the callbacks (although it exists in the Singleton, it will be better to pass it explicitly).

hydra/conf/__init__.py
```diff
@@ -122,6 +122,8 @@ class HydraConf:
     sweeper: Any = MISSING
     # Launcher configuration
     launcher: Any = MISSING
+    # Callbacks configuration
+    callbacks: Any = MISSING
```
omry, line 126: It would be better to make this an empty dictionary which is contributed to instead of MISSING.

Something like:
```python
callbacks: Dict[str, Any] = field(default_factory=dict)
```

hydra/callback/callback.py
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
omry, line 1: Put it in hydra.core, I don't think it justifies a new top level module in Hydra.

hydra/core/utils.py
```diff
@@ -79,13 +80,29 @@ def filter_overrides(overrides: Sequence[str]) -> Sequence[str]:
...
+    from hydra.utils import instantiate
+
+    callbacks = []
+    if OmegaConf.select(cfg, "hydra.callbacks"):
+        cfg = cfg.hydra.callbacks
+        callbacks = []
+        for c in cfg.keys():
+            callbacks.append(instantiate(OmegaConf.select(cfg, c)))
+
+    return callbacks
```
omry, line 93: select should be used when you want to do dynamic inspection.
there is nothing dynamic here:
```suggestion
      callbacks = []
      for callback in cfg.hydra.callbacks.values():
          callbacks.append(instantiate(callback))
    return callbacks
```

omry: The callbacks should also only be instantiated once.

hydra/core/utils.py
```diff
@@ -79,13 +80,29 @@ def filter_overrides(overrides: Sequence[str]) -> Sequence[str]:
...
 def run_job(
     config: DictConfig,
     task_function: TaskFunction,
     job_dir_key: str,
     job_subdir_key: Optional[str],
     configure_logging: bool = True,
 ) -> "JobReturn":
+    callbacks = get_callbacks(config)
+    for c in callbacks:
+        c.pre_job(config)
```
omry, line 105: 1. use already instantiated callback.
2. exceptions

omry (changes requested):

jieru-hu:
Close this for now. Will create a new one for the internal callbacks apis.

Issue #1093, opened 2020-10-27
hydra_ax_sweeper integration tests failing


jieru-hu:
I will follow up on this today 

CircleCI job link https://app.circleci.com/pipelines/github/jieru-hu/hydra/843/workflows/e6ba88e9-abb8-4724-bd98-4ac16fbbae3d/jobs/8622

```
----------------------------- Captured stderr call -----------------------------
Subprocess error:
Traceback (most recent call last):
  File "/home/circleci/project/.nox/test_plugins-3-7/lib/python3.7/site-packages/hydra/utils.py", line 109, in instantiate
    return target(*args, **final_kwargs)
  File "/home/circleci/project/.nox/test_plugins-3-7/lib/python3.7/site-packages/hydra_plugins/hydra_ax_sweeper/ax_sweeper.py", line 16, in __init__
    from ._core import CoreAxSweeper
  File "/home/circleci/project/.nox/test_plugins-3-7/lib/python3.7/site-packages/hydra_plugins/hydra_ax_sweeper/_core.py", line 6, in <module>
    from ax.core import types as ax_types  # type: ignore
  File "/home/circleci/project/.nox/test_plugins-3-7/lib/python3.7/site-packages/ax/__init__.py", line 31, in <module>
    from ax.modelbridge import Models
  File "/home/circleci/project/.nox/test_plugins-3-7/lib/python3.7/site-packages/ax/modelbridge/__init__.py", line 10, in <module>
    from ax.modelbridge.factory import (
  File "/home/circleci/project/.nox/test_plugins-3-7/lib/python3.7/site-packages/ax/modelbridge/factory.py", line 375, in <module>
    torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
  File "/home/circleci/project/.nox/test_plugins-3-7/lib/python3.7/site-packages/torch/cuda/__init__.py", line 52, in is_available
    return torch._C._cuda_getDeviceCount() > 0
UserWarning: CUDA initialization: Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx (Triggered internally at  /pytorch/c10/cuda/CUDAFunctions.cpp:100.)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/circleci/project/.nox/test_plugins-3-7/lib/python3.7/site-packages/hydra/_internal/utils.py", line 196, in run_and_report
    return func()
[...]

PR #1095, opened 2020-10-27
ignore UserWarning introduced by torch 1.7.0

jieru-hu:
## Motivation

This is to address https://github.com/facebookresearch/hydra/issues/1093
Once this PR is approved I will also cherry pick the change to 1.0 branch.

@shagunsodhani pointed that it might be related to a new torch release this mornign and indeed it is the case. To verify this assumption I pined ```torch==1.6.0``` and all the tests pass.

The warning was thrown here: https://github.com/pytorch/pytorch/blob/master/c10/cuda/CUDAFunctions.cpp#L100
introduced by this commit recently 
https://github.com/pytorch/pytorch/commit/06d978a9ad329380912fdc50547363e5f822a460

example failure & stacktrace:
https://app.circleci.com/pipelines/github/facebookresearch/hydra/4259/workflows/52186ca6-be5c-41d5-a544-e350786ab999/jobs/30831

Yes/No

## Test Plan

run on my linux box and verified the tests pass, CI should pass now. 
## Related Issues and PRs

### DIFF
--- a/plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
+++ b/plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
@@ -40,6 +40,7 @@ def quadratic(cfg: DictConfig) -> Any:
         (6, [[1, 2, 3, 4, 5]]),
     ],
 )
+@pytest.mark.filterwarnings("ignore::UserWarning")
 def test_chunk_method_for_valid_inputs(n: int, expected: List[List[int]]) -> None:
     from hydra_plugins.hydra_ax_sweeper._core import CoreAxSweeper  # type: ignore
 
@@ -210,7 +211,7 @@ def test_ax_logging(tmpdir: Path, cmd_arg: str, expected_str: str) -> None:
         "polynomial.z=10",
         "hydra.sweeper.ax_config.max_trials=2",
     ] + [cmd_arg]
-    result, _ = get_run_output(cmd)
+    result, _ = get_run_output(cmd=cmd, allow_warnings=True)
     assert "polynomial.x: range=[-5.0, -2.0]" in result
     assert expected_str in result
     assert "polynomial.z: fixed=10" in result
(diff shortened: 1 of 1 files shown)


shagunsodhani:
I think Ax/BoTorch should be adding this check - but for now your solution seems reasonable.

jieru-hu:
> I think Ax/BoTorch should be adding this check - but for now your solution seems reasonable.

Could you explain what you mean by checks? I'm not sure how ax/botorch would get rid of the warnings raise by torch. 🤔

shagunsodhani:
By `check` I meant checking for the warning raised by PyTorch and not passing it to the end user. Probably here: https://github.com/facebook/Ax/blob/master/ax/modelbridge/factory.py#L375

The [PyTorch PR](https://github.com/pytorch/pytorch/pull/42249) says that the warning should be thrown only when there is some issue with the gpu drivers. This suggests to me that the warning could be avoided if Ax/BoTorch makes sure that drivers (if any) match the PyTorch version.

jieru-hu:
> By `check` I meant checking for the warning raised by PyTorch and not passing it to the end user. Probably here: https://github.com/facebook/Ax/blob/master/ax/modelbridge/factory.py#L375
> 
> The [PyTorch PR](https://github.com/pytorch/pytorch/pull/42249) says that the warning should be thrown only when there is some issue with the gpu drivers. This suggests to me that the warning could be avoided if Ax/BoTorch makes sure that drivers (if any) match the PyTorch version.

I see! I will create an issue in their repo.

shagunsodhani:
For now, we should go ahead with this fix :)

shagunsodhani (approved):
LGTM

omry:
If this is something that should be fixed in Ax, be sure to undo the filtering once they fix it.

jieru-hu:
> If this is something that should be fixed in Ax, be sure to undo the filtering once they fix it.


Yes I've created an issue in their repo https://github.com/facebook/Ax/issues/417#issuecomment-718003107
Will remove this once they get it fixed.

PR #1098, opened 2020-10-29
[joblib_launcher] cast params to int when applicable

### LINKED ISSUE #1094: [Bug] Joblib launcher plugin doesn't properly convert some joblib arguments to integers
# 🐛 Bug
## Description

The following `joblib.Parallel` arguments are currently provided as strings to joblib when they should be integers:
* `pre_dispatch`
* `batch_size`
* `max_nbytes`

## Checklist
- [X] I checked on the latest version of Hydra
- [X] I created a minimal repro

## To reproduce

```bash
wget https://raw.githubusercontent.com/facebookresearch/hydra/master/examples/tutorials/basic/your_first_hydra_app/1_simple_cli/my_app.py
python my_app.py hydra.launcher.batch_size=1 hydra/launcher=joblib -m
```

```
[2020-10-27 18:20:48,363][HYDRA] Joblib.Parallel(n_jobs=-1,backend=loky,prefer=processes,require=None,verbose=0,timeout=None,pre_dispatch=2*n_jobs,batch_size=1,temp_folder=None,max_nbytes=None,mmap_mode=r) is launching 1 jobs
[2020-10-27 18:20:48,363][HYDRA] Launching jobs, sweep output dir : multirun/2020-10-27/18-20-48
[2020-10-27 18:20:48,363][HYDRA]        #0 :
  File "/Users/odelalleau/src/hydra/hydra/_internal/utils.py", line 207, in run_and_report
    return func()
  File "/Users/odelalleau/src/hydra/hydra/_internal/utils.py", line 364, in <lambda>
    lambda: hydra.multirun(
  File "/Users/odelalleau/src/hydra/hydra/_internal/hydra.py", line 136, in multirun
    return sweeper.sweep(arguments=task_overrides)
  File "/Users/odelalleau/src/hydra/hydra/_internal/core_plugins/basic_sweeper.py", line 154, in sweep
    results = self.launcher.launch(batch, initial_job_idx=initial_job_idx)
  File "/Users/odelalleau/src/hydra/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/joblib_launcher.py", line 45, in launch
    return _core.launch(
  File "/Users/odelalleau/src/hydra/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py", line 89, in launch
    runs = Parallel(**joblib_cfg)(
  File "/Users/odelalleau/.local/lib/python3.8/site-packages/joblib/parallel.py", line 715, in __init__
    raise ValueError(
ValueError: batch_size must be 'auto' or a positive integer, got: '1'
```

(with `max_nbytes` it also crashes due to the string issue, `pre_dispatch` actually runs but not sure what happens exactly since it's given as a string)

## Expected Behavior

Running without error, providing appropriate integer arguments to `joblib.Parallel` when they should be integers (obviously there are situations where they should remain strings).

## System information
- **Hydra Version** :  master
- **Python version** : 3.8.5
- **Virtual environment type and version** : conda 4.8.5

jieru-hu:
## Motivation

this is to fix #1094
Some of Joblib.Parallel's params supports int, so add casting for that when applicable. Also fixed the typing for `timeout` from `int` to `float`.

Yes

## Test Plan

CI, also added integration tests. 

## Related Issues and PRs

### DIFF
--- a/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py
+++ b/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py
@@ -50,6 +50,18 @@ def execute_job(
     return ret
 
 
+def process_joblib_cfg(joblib_cfg: Dict[str, Any]) -> Dict[str, Any]:
+    for k, v in joblib_cfg.items():
+        if k in ["pre_dispatch", "batch_size", "max_nbytes"] and v:
+            try:
+                joblib_cfg[k] = int(v)
+            except ValueError:
+                joblib_cfg[k] = v
+        else:
+            joblib_cfg[k] = v
+    return joblib_cfg
+
+
 def launch(
     launcher: JoblibLauncher,
     job_overrides: Sequence[Sequence[str]],
@@ -73,6 +85,7 @@ def launch(
     # backend is incompatible with Hydra
     joblib_cfg = launcher.joblib
     joblib_cfg["backend"] = "loky"
+    joblib_cfg = process_joblib_cfg(joblib_cfg)
 
     log.info(
         "Joblib.Parallel({}) is launching {} jobs".format(
--- a/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py
+++ b/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py
@@ -52,9 +52,11 @@ def execute_job(
 
 def process_joblib_cfg(joblib_cfg: Dict[str, Any]) -> None:
     for k in ["pre_dispatch", "batch_size", "max_nbytes"]:
-        if k in joblib_cfg.keys() and joblib_cfg.get(k):
+        if k in joblib_cfg.keys():
             try:
-                joblib_cfg[k] = int(joblib_cfg.get(k))
+                val = joblib_cfg.get(k)
+                if val:
+                    joblib_cfg[k] = int(val)
             except ValueError:
                 pass
 
--- a/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py
+++ b/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py
@@ -50,16 +50,13 @@ def execute_job(
     return ret
 
 
-def process_joblib_cfg(joblib_cfg: Dict[str, Any]) -> Dict[str, Any]:
-    for k, v in joblib_cfg.items():
-        if k in ["pre_dispatch", "batch_size", "max_nbytes"] and v:
+def process_joblib_cfg(joblib_cfg: Dict[str, Any]) -> None:
+    for k in ["pre_dispatch", "batch_size", "max_nbytes"]:
+        if k in joblib_cfg.keys() and joblib_cfg.get(k):
             try:
-                joblib_cfg[k] = int(v)
+                joblib_cfg[k] = int(joblib_cfg.get(k))
             except ValueError:
-                joblib_cfg[k] = v
-        else:
-            joblib_cfg[k] = v
-    return joblib_cfg
+                pass
 
 
 def launch(
@@ -85,7 +82,7 @@ def launch(
     # backend is incompatible with Hydra
     joblib_cfg = launcher.joblib
     joblib_cfg["backend"] = "loky"
-    joblib_cfg = process_joblib_cfg(joblib_cfg)
+    process_joblib_cfg(joblib_cfg)
 
     log.info(
         "Joblib.Parallel({}) is launching {} jobs".format(
--- a/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/config.py
+++ b/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/config.py
@@ -24,7 +24,7 @@ class JobLibLauncherConf:
     # if greater than zero, prints progress messages
     verbose: int = 0
 
-    # timeout limit for each task
+    # timeout limit for each task. Unit dependent on backend implementation; miliseconds for loky.
     timeout: Optional[float] = None
 
     # number of batches to be pre-dispatched
--- a/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/config.py
+++ b/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/config.py
@@ -25,7 +25,7 @@ class JobLibLauncherConf:
     verbose: int = 0
 
     # timeout limit for each task
-    timeout: Optional[int] = None
+    timeout: Optional[float] = None
 
     # number of batches to be pre-dispatched
     pre_dispatch: str = "2*n_jobs"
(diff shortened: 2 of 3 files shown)


plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py
```diff
@@ -50,6 +50,18 @@ def execute_job(
     return ret
 
 
+def process_joblib_cfg(joblib_cfg: Dict[str, Any]) -> Dict[str, Any]:
+    for k, v in joblib_cfg.items():
+        if k in ["pre_dispatch", "batch_size", "max_nbytes"] and v:
+            try:
+                joblib_cfg[k] = int(v)
```
shagunsodhani, line 57: Wouldnt this convert the floats to int as well?

jieru-hu: these three params only allows int (https://joblib.readthedocs.io/en/latest/parallel.html#parallel-reference-documentation), so I think we should be good here?

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py
```diff
@@ -50,6 +50,18 @@ def execute_job(
...
+            except ValueError:
+                joblib_cfg[k] = v
+        else:
```
shagunsodhani, line 60: This `else` is not required

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py
```diff
@@ -50,6 +50,18 @@ def execute_job(
     return ret
 
 
+def process_joblib_cfg(joblib_cfg: Dict[str, Any]) -> Dict[str, Any]:
+    for k, v in joblib_cfg.items():
+        if k in ["pre_dispatch", "batch_size", "max_nbytes"] and v:
```
shagunsodhani, line 55: maybe just check these keys directly (instead of iterating on the dict)

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py
```diff
@@ -50,6 +50,18 @@ def execute_job(
...
+            try:
+                joblib_cfg[k] = int(v)
+            except ValueError:
+                joblib_cfg[k] = v
```
shagunsodhani, line 59: can be replaced by `pass`

plugins/hydra_joblib_launcher/tests/test_joblib_launcher.py
```diff
@@ -67,3 +70,34 @@ def test_example_app(hydra_sweep_runner: TSweepRunner) -> None:
         assert sweep.returns is not None and len(sweep.returns[0]) == 4
         for ret in sweep.returns[0]:
             assert tuple(ret.overrides) in overrides
+
+
+@pytest.mark.parametrize(  # type: ignore
+    "overrides, expected",
+    [
+        ("hydra.launcher.batch_size=1", ("hydra.launcher.batch_size", "1")),
```
shagunsodhani, line 78: maybe I am overlooking something but shouldnt the value be int 1 and not "1" ?

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py
jieru-hu: thx, yea initially I wanted to create a new dict here which is why I looped all the items.

Took your suggestions and simplied the method now.

plugins/hydra_joblib_launcher/tests/test_joblib_launcher.py
jieru-hu: it is still a string, since we are not altering the Hydra config. the casting happens right before we pass the params to joblib.

The test's main goal is to make sure the example app runs with no issue (without the changes the example app will not run)

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/config.py
```diff
@@ -25,7 +25,7 @@ class JobLibLauncherConf:
     verbose: int = 0
 
     # timeout limit for each task
-    timeout: Optional[int] = None
+    timeout: Optional[float] = None
```
omry, line 28: 1. What is the unit of the timeout? can you add a comment?
2. Did you verify in the joblib docs that it can take a float as the timeout?

plugins/hydra_joblib_launcher/tests/test_joblib_launcher.py
omry, line 78: Why put the two values in a tuple?
```
"overrides, tested_key,expected_value",
```

Also, I am not sure testing the key and value here adds any value. this is testing Hydra's overriding capability, which is very well tested already.

shagunsodhani: got it :)

jieru-hu:
Thx for the comments @omry  @shagunsodhani, response down inline :) 

>What is the unit of the timeout? can you add a comment?
Did you verify in the joblib docs that it can take a float as the timeout?

Yes I verified it the doc [here](https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html)
The unit was not documented in their doc (I think it is because it's dependent on backend implementation)
I did find in their source code that for `loky` it is miliseconds. Is that something we should document if it's buried deep in their source code and backend dependent? 
Also I'm wondering should we even allow users to override the backend if `loky` is the only one we supported?
(right now we have backend in the config, but we hardcode it to `loky` inside the launcher.)


> Why put the two values in a tuple?
"overrides, tested_key,expected_value",
Also, I am not sure testing the key and value here adds any value. this is testing Hydra's overriding capability, which is very well tested already.

Thx. I can see now how this is confusing since you and shagun both had the questions.
I've updated the tests to make it more clear :)

shagunsodhani:
> "I did find in their source code that for loky it is miliseconds. Is that something we should document if it's buried deep in their source code and backend dependent?"
Dont know what is the right thing here but I since "loky" is the only backend we support currently, it may be useful to document it

plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/config.py
omry: So, can you add a comment with the time units?
If this is specific to loki, just mention it in the comment.

omry:
What happens if you override the backend now?
Since we only want to support Loki, we can probably remove the backend from the dataclass (hopefully people are not specifying it because that would break their config once we do).

jieru-hu:
> So, can you add a comment with the time units?
If this is specific to loki, just mention it in the comment.

done.

> What happens if you override the backend now?
Since we only want to support Loki, we can probably remove the backend from the dataclass (hopefully people are not specifying it because that would break their config once we do).

The override won't work since the we [override](https://github.com/facebookresearch/hydra/blob/master/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py#L75) it in the launcher to `loky`. Since we already call this out in the doc, I will leave it be for this PR since this PR to fix the params.

omry:
We can have a followup PR deprecate the backend field and eventually kill it.

jieru-hu:
> We can have a followup PR deprecate the backend field and eventually kill it.

sounds good :)

jieru-hu:
@omry @shagunsodhani  can i get a ship it if no more comments?

omry (approved):
looking good. feel free to land.

Issue #1135, opened 2020-11-11
[Feature Request] [Submitit] [array_parallelism]


emilemathieu:
# 🚀 Feature Request

Hi all!

Would there be a way to limit the number of map tasks that can be executed in parallel?

In `submitit` one can use the parameter `slurm_array_parallelism` as in the[ job array example](https://github.com/facebookincubator/submitit/blob/master/docs/examples.md#job-arrays).
Passing `additional_parameters: {"array_parallelism": 2}` does not work since it's not a `slurm` argument.

Cheers,
Emile

omry:
This sounds like a nice addition, want to try to put out a pull request?

emilemathieu:
Would the idea to add in [SlurmQueueConf](https://github.com/facebookresearch/hydra/blob/master/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py#L31) a [parameter `array_parallelism: int = 256`](https://github.com/facebookincubator/submitit/blob/master/submitit/slurm/slurm.py#L410)?

jieru-hu:
> Would the idea to add in [SlurmQueueConf](https://github.com/facebookresearch/hydra/blob/master/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py#L31) a [parameter `array_parallelism: int = 256`](https://github.com/facebookincubator/submitit/blob/master/submitit/slurm/slurm.py#L410)?

Yes, looks like a good place to add the config :)

shagunsodhani:
Closed via https://github.com/facebookresearch/hydra/pull/1137

PR #1137, opened 2020-11-12
[submitit] Can now specify the maximum number of jobs running in parallel

emilemathieu:
## Motivation

There is currently no way to limit the number of tasks that can be executed in parallel as in `submitit` or directly with a slurm `job array`.

Yes I did

## Test Plan

I did not find any test for `hydra/launcher=submitit_slurm` in the [hydra_submitit_launcher plugging test file](https://github.com/facebookresearch/hydra/blob/master/plugins/hydra_submitit_launcher/tests/test_submitit_launcher.py).
For what it's worth I did try it on a slurm cluster and it did work.

## Related Issues and PRs
https://github.com/facebookresearch/hydra/issues/1135

### DIFF
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
@@ -31,9 +31,7 @@ class BaseQueueConf:
 class SlurmQueueConf(BaseQueueConf):
     """Slurm configuration overrides and specific parameters"""
 
-    _target_: str = (
-        "hydra_plugins.hydra_submitit_launcher.submitit_launcher.SlurmLauncher"
-    )
+    _target_: str = ("hydra_plugins.hydra_submitit_launcher.submitit_launcher.SlurmLauncher")
 
     # Params are used to configure sbatch, for more info check:
     # https://github.com/facebookincubator/submitit/blob/master/submitit/slurm/slurm.py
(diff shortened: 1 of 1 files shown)


jieru-hu:
Thx for the PR.

Could you check out our developer guide here https://hydra.cc/docs/next/development/overview 
and fix the lints :)

>I did not find any test for hydra/launcher=submitit_slurm in the hydra_submitit_launcher plugging test file.

Yes, currently we don't have a way to test on slurm on circleCI and as a result there's no test.
Could you update your test plan with what you tried and the console output?

emilemathieu:
Sorry fixed the lint.

Concerning the test, I used `hydra.launcher.array_parallelism=2` and got only 2 jobs at a time with `squeue -u $USER`.
More rigorously the `.submitit/JOB_ID/JOB_ID _submission.sh` sbatch file does feature the option `#SBATCH --array=0-1%2` (with `array_parallelism=2`).

jieru-hu:
lgtm

cc @omry @shagunsodhani

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -60,6 +60,8 @@ class SlurmQueueConf(BaseQueueConf):
     # Useful to add parameters which are not currently available in the plugin.
     # Eg: {"mail-user": "blublu@fb.com", "mail-type": "BEGIN"}
     additional_parameters: Dict[str, Any] = field(default_factory=dict)
+    # Maximum number of jobs running in parallel
+    array_parallelism: int = 256
```
omry, line 64: short and sweet.
Anything else we should be adding from the submitit API while we are here?

emilemathieu: Good question, none that I needed at least, since `additional_parameters` allows to pass parameters directly to `slurm`.
Yet, some parameters from [submitit API](https://github.com/facebookincubator/submitit/blob/abf1993cef94b84c480fc0d7dc7676700996e700/submitit/slurm/slurm.py#L387) are missing from the [hydra plugin API](https://github.com/facebookresearch/hydra/blob/affb12a85c64bce17408aab890ff6de30293ff3a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py) like `cpus_per_gpu` but I'm only starting to use slurm so no idea whether that is useful.

omry: @jrapin, in your opinion - anything missing that we should add while we are at it?

shagunsodhani: I think only the variables [here](https://github.com/facebookincubator/submitit/blob/abf1993cef94b84c480fc0d7dc7676700996e700/submitit/slurm/slurm.py#L443) should be considered (since everything else can be passed via `additional_parameters`). I think `map_count` could be useful.

omry: what is it useful for?

shagunsodhani: It decides the number of jobs to be run in parallel. Digging more, it does not need to be set explicitly.

omry (approved):

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jrapin: > @jrapin, in your opinion - anything missing that we should add while we are at it?

sorry for the late reply :s
I can't think of anything

omry: No worries, thanks @jrapin !

Issue #951, opened 2020-09-08
Create an example for PyTorch DDP


omry:
- [ ] Launching a single DDP group with multirun
- [ ] Launching multiple DDP groups with multirun

[Chat topic](https://hydra-framework.zulipchat.com/#narrow/stream/213879-Hydra/topic/combination.20hydra.20and.20pytorch's.20distributed.2Elaunch) with ideas about implementation strategy.

omry:
Such an example will live in https://github.com/pytorch/hydra-torch when someone makes it.

PR #1141, opened 2020-11-16
Minimal DDP implementation with multirun

briankosw:
## Motivation

Created a minimal implementation of using distributed computation with multirun for a single DDP group, as mentioned in #951. 

```bash
>>> python main.py -m rank=0,1,2,3

[2020-11-16 19:19:22,512][HYDRA] Joblib.Parallel(n_jobs=-1,backend=loky,prefer=processes,require=None,verbose=0,timeout=None,pre_dispatch=2*n_jobs,batch_size=auto,temp_folder=None,max_nbytes=None,mmap_mode=r) is launching 4 jobs
[2020-11-16 19:19:22,513][HYDRA] Launching jobs, sweep output dir : multirun/2020-11-16/19-19-21
[2020-11-16 19:19:22,513][HYDRA]        #0 : rank=0
[2020-11-16 19:19:22,513][HYDRA]        #1 : rank=1
[2020-11-16 19:19:22,513][HYDRA]        #2 : rank=2
[2020-11-16 19:19:22,513][HYDRA]        #3 : rank=3
Rank 1 - 0.035238564014434814
Rank 0 - 0.7433204650878906
Rank 2 - 0.836401641368866
Rank 3 - 0.3215029239654541
Rank 0 has average: 0.4841158986091614
```

(Yes)/No

## Test Plan

### DIFF
--- a/examples/advanced/pytorch_ddp_example/main.py
+++ b/examples/advanced/pytorch_ddp_example/main.py
@@ -1,14 +1,12 @@
-import multiprocessing as mp
 import os
-import time
 
 import hydra
-from omegaconf import DictConfig, OmegaConf
+from omegaconf import DictConfig
 import torch
 import torch.distributed as dist
 
 
-def setup(master_addr, master_port, rank, world_size, backend):
+def setup(master_addr: str, master_port: str, rank: int, world_size: int, backend: str):
     """Initializes distributed process group.
 
     Arguments:
@@ -22,6 +20,7 @@ def setup(master_addr, master_port, rank, world_size, backend):
 
 
 def cleanup():
+    """Cleans up distributed backend resources."""
     dist.destroy_process_group()
 
 
--- a/examples/advanced/pytorch_ddp_example/main.py
+++ b/examples/advanced/pytorch_ddp_example/main.py
@@ -1,9 +1,10 @@
 import os
 
-import hydra
-from omegaconf import DictConfig
 import torch
 import torch.distributed as dist
+from omegaconf import DictConfig
+
+import hydra
 
 
 def setup(master_addr: str, master_port: str, rank: int, world_size: int, backend: str):
--- a/examples/advanced/pytorch_ddp_example/main.py
+++ b/examples/advanced/pytorch_ddp_example/main.py
@@ -1,0 +1,42 @@
+import multiprocessing as mp
+import os
+import time
+
+import hydra
+from omegaconf import DictConfig, OmegaConf
+import torch
+import torch.distributed as dist
+
+
+def setup(master_addr, master_port, rank, world_size, backend):
+    """Initializes distributed process group.
+
+    Arguments:
+        rank: the rank of the current process.
+        world_size: the total number of processes.
+        backend: the backend used for distributed processing.
+    """
+    os.environ["MASTER_ADDR"] = master_addr
+    os.environ["MASTER_PORT"] = master_port
+    dist.init_process_group(backend=backend, rank=rank, world_size=world_size)
+
+
+def cleanup():
+    dist.destroy_process_group()
+
+
+@hydra.main(config_path="conf", config_name="config")
+def main(cfg: DictConfig):
+    setup(cfg.master_addr, cfg.master_port, cfg.rank, cfg.world_size, cfg.backend)
+    group = dist.new_group(list(range(cfg.world_size)))
+    tensor = torch.rand(1)
+    print("Rank {} - {}".format(cfg.rank, tensor[0]))
+    dist.reduce(tensor, dst=0, op=dist.ReduceOp.SUM, group=group)
+    if cfg.rank == 0:
+        tensor /= 4
+        print("Rank {} has average: {}".format(cfg.rank, tensor[0]))
+    cleanup()
+
+
+if __name__ == "__main__":
+    main()
--- a/examples/advanced/pytorch_ddp_example/conf/config.yaml
+++ b/examples/advanced/pytorch_ddp_example/conf/config.yaml
@@ -1,0 +1,8 @@
+master_addr: "localhost"
+master_port: "12345"
+backend: "gloo"
+rank: 0
+world_size: 4
+
+defaults:
+    - hydra/launcher: joblib


examples/advanced/pytorch_ddp_example/main.py
```diff
@@ -0,0 +1,42 @@
...
+@hydra.main(config_path="conf", config_name="config")
+def main(cfg: DictConfig):
+    setup(cfg.master_addr, cfg.master_port, cfg.rank, cfg.world_size, cfg.backend)
+    group = dist.new_group(list(range(cfg.world_size)))
+    tensor = torch.rand(1)
+    print("Rank {} - {}".format(cfg.rank, tensor[0]))
+    dist.reduce(tensor, dst=0, op=dist.ReduceOp.SUM, group=group)
+    if cfg.rank == 0:
+        tensor /= 4
+        print("Rank {} has average: {}".format(cfg.rank, tensor[0]))
```
omry, line 37: 1. Please use Python 3 fstrings.
```suggestion
        print(f"Rank {cfg.rank} has average: {tensor[0]}")
```

2. Please use logging:

```python
log.info(f"Rank {cfg.rank} has average: {tensor[0]}")
```

examples/advanced/pytorch_ddp_example/main.py
```diff
@@ -0,0 +1,42 @@
...
+def cleanup():
+    """Cleans up distributed backend resources."""
+    dist.destroy_process_group()
+
+
+@hydra.main(config_path="conf", config_name="config")
+def main(cfg: DictConfig):
+    setup(cfg.master_addr, cfg.master_port, cfg.rank, cfg.world_size, cfg.backend)
+    group = dist.new_group(list(range(cfg.world_size)))
+    tensor = torch.rand(1)
```
omry, line 32: You can fill the tensors according to the rank and make sure the average is what you expect (not random).

omry (commented):
Nice, thanks!

Can you try to make it work by using the explicit IP address of the parent process instead of localhost? this would translate better for distributed scenarios (using the submitit launcher, for example).

examples/advanced/pytorch_ddp_example/main.py
```diff
@@ -0,0 +1,42 @@
...
+        backend: the backend used for distributed processing.
+    """
+    os.environ["MASTER_ADDR"] = master_addr
+    os.environ["MASTER_PORT"] = master_port
+    dist.init_process_group(backend=backend, rank=rank, world_size=world_size)
+
+
+def cleanup():
+    """Cleans up distributed backend resources."""
+    dist.destroy_process_group()
```
shagunsodhani, line 25: We generally do not need to do this. Did you see any error without doing this?

shagunsodhani (commented):
Thanks for the PR. 

1. It is not guaranteed that all the jobs in JobLib will start at the same time (or around the same time). This is probably outside the scope of this PR but something to think about in general.
2. It is not guaranteed that the different processes will be scheduled on the same node. As Omry mentioned, it will be useful to pass the parent's ip (I think that means that the parent process participates in the rendezvous). 
3. A nit-picky comment - the example shows how to use `torch.distributed` (and not DDP). Maybe we can change the scope of the PR to focus on an example of `torch.distributed`? Alternatively, the [ImageNet example](https://github.com/pytorch/examples/tree/master/imagenet) will be easy to port here or maybe a simpler example, but with a model and dataloader etc.

examples/advanced/pytorch_ddp_example/main.py
omry: > A nit-picky comment - the example shows how to use torch.distributed (and not DDP). Maybe we can change the scope of the PR to focus on an example of torch.distributed? Alternatively, the ImageNet example will be easy to port here or maybe a simpler example, but with a model and dataloader etc.

I am also considering this this example belongs there.
A more complete example can live in https://github.com/pytorch/hydra-torch (which is still WIP).

briankosw: >  A nit-picky comment - the example shows how to use `torch.distributed` (and not DDP). Maybe we can change the scope of the PR to focus on an example of `torch.distributed`? Alternatively, the [ImageNet example](https://github.com/pytorch/examples/tree/master/imagenet) will be easy to port here or maybe a simpler example, but with a model and dataloader etc.
I completely agree with you. The reason why I implemented a `torch.distributed` example for now is to create a prototypical or minimal example that demonstrates multirun and Joblib being used to launch the distributed processes. I'm in the middle of fixing my GPU cluster, after which I plan to commit a full DDP example.

briankosw: >It is not guaranteed that all the jobs in JobLib will start at the same time (or around the same time). This is probably outside the scope of this PR but something to think about in general.

Is there a way to guarantee that the jobs start together?

> It is not guaranteed that the different processes will be scheduled on the same node. As Omry mentioned, it will be useful to pass the parent's ip (I think that means that the parent process participates in the rendezvous).

I intended this PR to be the single node case. I will definitely try to work on the multi-node case, but I'm less familiar with it compared to the single node case.

briankosw: @omry should I create PR in https://github.com/pytorch/hydra-torch instead for the ImageNet DDP example?

briankosw:
#394 would be super helpful for this, since I need to manually specify multirun every time I want to run the script.

examples/advanced/pytorch_ddp_example/main.py
omry: @briankosw,
If you want to create a DDP example based on ImageNet - yes.
I also think that this example probably should go there as well. (hydra-torch repository did not exist when I first created the issue).

briankosw:
Any consensus on what to do with this particular PR @omry @shagunsodhani? I'll open a separate issue and PR in hydra-torch, but I'll let you guys decide whether this PR (or something similar in spirit) belongs to this repo as well.

omry:
The reason hydra-torch is called `hydra-torch` and not `hydra-torch-configs` is exactly to give space for things like DDP examples :).
This example belongs there. Once the repo there is ready we will be able to link to it from the Hydra website.

briankosw:
Sounds good! Closing this PR then.

omry:
Thanks @briankosw!

PR #1208, opened 2020-12-14
Allow more types of dictionary keys in overrides grammar

odelalleau:
## Motivation

Requested in https://github.com/omry/omegaconf/commit/d959a131d3e21dc0321dcb4299bbb7bfb5b56b77#r45076606

## Remarks & questions

1. I wasn't entirely sure where to put this test: https://github.com/odelalleau/hydra/blob/ccde9814d66042f4c6b9fe7ec3db77e253297265/tests/test_hydra.py#L1190 . It's important to have it in addition to the grammar tests in test_overrides_parser because it also tests the config merge (in particular, it matters for quoted strings). Also ideally this would have been a parameterized test, but since `integration_test()` is slow, I thought it would be more convenient to have a single test to avoid slowing down the test suite even more.

2. Compared to what I did in OmegaConf, I'm not supporting interpolations as dictionary keys in Hydra's overrides grammar. This is because dictionaries in Hydra are meant to be merged into DictConfigs, and allowing interpolations would raise several questions that I believe are better left for later (should we resolve interpolations before or after merging overrides? do we allow keys to change names dynamically?).

3. As in my OmegaConf PR, I renamed `dictValue` and `listValue` to `dictContainer` and `listContainer`, to avoid any confusion that may arise due to the new `dictKey` parser rule. However, this change impacts more things in Hydra than it did in OmegaConf. Let me know if that's still ok (edit: I should add that there remain some `DictValues` and `ListValues` in tests that I didn't change, as I wasn't entirely sure if the "value" there was directly related to the name of the rules)

4. When implementing equality for quoted strings, it made sense to me to ignore the type of quotes in `__eq__`. It also didn't break any existing test. Just wanted to flag it here in case it might be a concern.

5. In this PR the grammar allows non-string dictionary keys like int / float / bool, but in practice trying to use them will not work since OmegaConf does not support them currently. Not sure if I should keep them to be "future proof", or just remove them for now.

6. I assumed this would make it into 1.0.x and thus I updated the 1.0 doc as well. If that's not the case let me know and I'll revert this change.

7. I didn't add a news item yet as it didn't really seem worth one on its own (the ability to use non-string keys would be more meaningful to advertise IMO)

## Test Plan

pytest (see new tests in the diff)

## Related Issues and PRs

Related to https://github.com/omry/omegaconf/pull/445


hydra/core/override_parser/types.py
```diff
@@ -26,6 +26,16 @@ class QuotedString:
...
+    def __hash__(self) -> int:
+        return hash(self.text)
+
+    def __eq__(self, other: Any) -> Any:
+        # We do not care whether quotes match for equality.
+        if isinstance(other, QuotedString):
+            return self.text == other.text
+        else:
+            return NotImplemented
+
```
omry, line 38: There are multiple tests that are comparing to QuotedString. This makes them softer.
What is the motivation for this change?

hydra/core/override_parser/types.py
```diff
@@ -258,7 +268,10 @@ def _convert_value(value: ParsedElementType) -> Optional[ElementType]:
         if isinstance(value, list):
             return [Override._convert_value(x) for x in value]
         elif isinstance(value, dict):
-            return {k: Override._convert_value(v) for k, v in value.items()}
+            return {
```
omry, line 271: Do you have any test covering this change?
I am not sure I understand why it's needed or what the implications are to callers.

hydra/core/override_parser/overrides_visitor.py
```diff
@@ -76,7 +76,7 @@ def is_ws(self, c: Any) -> bool:
         return isinstance(c, TerminalNodeImpl) and c.symbol.type == OverrideLexer.WS
 
     def visitPrimitive(
-        self, ctx: OverrideParser.PrimitiveContext
+        self, ctx: Union[OverrideParser.PrimitiveContext, OverrideParser.DictKeyContext]
```
omry, line 79: This is extending generated code, right?
I suspect you edited it by hand to get rid of the mypy type error.
If so, please refactor to call a common function with union signature.

website/versioned_docs/version-1.0/advanced/override_grammar/basic.md
```diff
@@ -95,6 +95,19 @@ primitive:
...
+      QUOTED_VALUE                               // 'hello world', "hello world"
+    | (   ID                                     // foo_10
+        | NULL                                   // null, NULL
+        | INT                                    // 0, 10, -20, 1_000_000
+        | FLOAT                                  // 3.14, -20.0, 1e-1, -10e3
+        | BOOL                                   // true, TrUe, false, False
+        | UNQUOTED_CHAR                          // /, -, \, +, ., $, %, *, @
+        | ESC                                    // \\, \(, \), \[, \], \{, \}, \:, \=, \ , \\t, \,
+        | WS                                     // whitespaces
+    )+;
```
omry, line 110: This change is not for 1.0, no need to update it's docs.

tests/test_hydra.py
```diff
@@ -1047,6 +1047,29 @@ def test_run_pass_list(self, cmd_base: List[str], tmpdir: Any) -> None:
         ret, _err = get_run_output(cmd)
         assert OmegaConf.create(ret) == OmegaConf.create(expected)
 
+    def test_multirun_dict_keys(self, cmd_base: List[str], tmpdir: Any) -> None:
```
omry, line 1050: This test, and test_overrides_dict_keys - are very high level.
you have some low level tests in test_overrides_parser and you are jumping straight into testing as processes.

If I understand the purpose of those tests correctly (and it's hard because they are so high level), I think should should be replaced by lower level config composition test in test_config_loader.py.

omry (changes requested):

omry:
> 1. I wasn't entirely sure where to put this test: https://github.com/odelalleau/hydra/blob/ccde9814d66042f4c6b9fe7ec3db77e253297265/tests/test_hydra.py#L1190 . It's important to have it in addition to the grammar tests in test_overrides_parser because it also tests the config merge (in particular, it matters for quoted strings). Also ideally this would have been a parameterized test, but since `integration_test()` is slow, I thought it would be more convenient to have a single test to avoid slowing down the test suite even more.

I commented about the high level tests.

> 2. Compared to what I did in OmegaConf, I'm not supporting interpolations as dictionary keys in Hydra's overrides grammar. This is because dictionaries in Hydra are meant to be merged into DictConfigs, and allowing interpolations would raise several questions that I believe are better left for later (should we resolve interpolations before or after merging overrides? do we allow keys to change names dynamically?).

I suspect OmegaConf dict keys should also not support interpolation for the same reasons.

> 3. As in my OmegaConf PR, I renamed `dictValue` and `listValue` to `dictContainer` and `listContainer`, to avoid any confusion that may arise due to the new `dictKey` parser rule. However, this change impacts more things in Hydra than it did in OmegaConf. Let me know if that's still ok (edit: I should add that there remain some `DictValues` and `ListValues` in tests that I didn't change, as I wasn't entirely sure if the "value" there was directly related to the name of the rules)

Sounds good. I saw that you actually did rename dictValue to dictContainer in the tests. anything else you left as is?

> 4. When implementing equality for quoted strings, it made sense to me to ignore the type of quotes in `__eq__`. It also didn't break any existing test. Just wanted to flag it here in case it might be a concern.

Commented inline.

> 5. In this PR the grammar allows non-string dictionary keys like int / float / bool, but in practice trying to use them will not work since OmegaConf does not support them currently. Not sure if I should keep them to be "future proof", or just remove them for now.

Does they work when sweeping over dictionaries?
```
--multirun foo={a:10},{10:a}
```

> 6. I assumed this would make it into 1.0.x and thus I updated the 1.0 doc as well. If that's not the case let me know and I'll revert this change.

Commented line, revert. this is for 1.1.

> 7. I didn't add a news item yet as it didn't really seem worth one on its own (the ability to use non-string keys would be more meaningful to advertise IMO)

I think we need a news fragment, but first we need to pin point exactly what this is enabling.

hydra/core/override_parser/types.py
odelalleau: So, my original motivation was that I had to implement `__hash__()` and this made me wonder what it meant for two quoted strings to be equal, and I thought it would make more sense to ignore quotes (just like `'xyz' == "xyz"` in Python). Since these quoted strings can now be dictionary keys, I also thought that if we ever wanted to merge two dictionaries, we may otherwise accidentally end up with a dictionary with two quoted strings representing the same string, but with different quotes... that being said, I have no idea if it's something that may happen in the current codebase.

I can definitely revert that change if you prefer.

hydra/core/override_parser/types.py
odelalleau: Both tests I added in `test_hydra.py` fail without this change.

The error looks like this:
```
Error merging override foo={'quoted_$(){}[]': 1, id123: 1, 123id: 1, a/-\+.$%*@: 1, \\\(\)\[\]\{\}\:\=\ \       \,: 1, white space: 1}
Incompatible key type 'QuotedString'
        full_key:
        reference_type=Optional[Dict[Union[str, Enum], Any]]
        object_type=dict
```

What happens is that otherwise, quoted strings remain in the dictionary passed to OmegaConf, which doesn't like them.

hydra/core/override_parser/overrides_visitor.py
odelalleau: Just to be sure it's clear, this class itself isn't generated: it is a subclass of a generated class. So nothing bad happens when we regenerate the code for instance.

I had considered creating a common function but thought it was a bit overkill... anyway, I did it in ee5eaa807 :)

website/versioned_docs/version-1.0/advanced/override_grammar/basic.md
odelalleau: Reverted doc changes in a99c690f6

tests/test_hydra.py
odelalleau: So, to clarify the motivation: those two new tests are meant to test the changes in `types.py` (which are not tested by the low level grammar tests)

The second one (`test_overrides_dict_keys`) should probably indeed be a lower level test. I wasn't familiar with the various Hydra tests and couldn't figure out where the basic override functionalities were being tested. I just moved it to `test_config_loader.py` as you suggested: see 6bcaac207, let me know if that makes more sense this way. The main thing that needs to be tested is quoted strings as keys, but I thought it wouldn't hurt to add tests for all other kinds of string formatting that we may expect, just in case (also later on, we should also test other types of keys like int / float / bool if we support them).

I'm not sure I can move `test_multirun_dict_keys` to this same file though. This test specifically tests the change to `_get_value_element_as_str()` (l. 429 of types.py in the current diff of this PR), which is used in sweeps to provide the proper command line overrides. Again, the main thing that needs testing are quoted strings, but I added the other key formats as well to be safe.
Maybe to clarify the purpose of this test, I can show you how it fails if I revert the l.429 change:
```
LexerNoViableAltException: +foo={QuotedString(text='null', quote=<Quote.single: 0>):0}
```
(this is because the sweeper would not replace the `QuotedString` with its actual quoted string representation in the command line override)

Side note: I also pushed 0b532092a which removes some comments that seemed irrelevant to me (probably a copy/paste leftover from another comment you can see below them)

odelalleau:
Thanks for the review & comments!

> > 1. Compared to what I did in OmegaConf, I'm not supporting interpolations as dictionary keys in Hydra's overrides grammar. This is because dictionaries in Hydra are meant to be merged into DictConfigs, and allowing interpolations would raise several questions that I believe are better left for later (should we resolve interpolations before or after merging overrides? do we allow keys to change names dynamically?).
> 
> I suspect OmegaConf dict keys should also not support interpolation for the same reasons.

Currently dictionaries in OmegaConf's grammar are used for a different purpose though: they are only inputs to resolvers. So these questions don't actually arise.
I can drop support for interpolations as keys in OmegaConf though if you prefer (it's always possible to replace a dictionary with a list of pairs if someone really needs them).

> Sounds good. I saw that you actually did rename dictValue to dictContainer in the tests. anything else you left as is?

Yeah, so if you search for `DictValues` and `ListValues` (with an "s"), you'll notice some classes in tests with these names. I wasn't 100% sure I should rename them (mostly because of the "s", which threw me off, and I didn't try to fully understand what this was testing). Happy to rename them as well if you tell me I should!

> > 1. In this PR the grammar allows non-string dictionary keys like int / float / bool, but in practice trying to use them will not work since OmegaConf does not support them currently. Not sure if I should keep them to be "future proof", or just remove them for now.
> 
> Does they work when sweeping over dictionaries?
> 
> ```
> --multirun foo={a:10},{10:a}
> ```

Why would they?
```shell
> python my_app.py +foo="{a:10},{10:a}" -m                                                                                 1 ↵  2.48 L

Error merging override +foo={10:a}
Incompatible key type 'int'
        full_key: foo.10
        reference_type=Any
        object_type=dict
```

Note that I used `+`, because anyway I can't create a config with an existing `foo` that would have `10` (the integer) as key.

> I think we need a news fragment, but first we need to pin point exactly what this is enabling.

If we merge it in current master, it allows command line overrides with dictionaries whose keys are strings in "non ID" format (see `test_dict_key_formats()` for examples). Currently if you have a key in your config whose name is not an ID, you can't use it as a dictionary key in your command line overrides.

In the future, it should allow also command line overrides with dictionaries whose keys are *not* strings (ex: int, float, bool). But that can only happen after OmegaConf supports those.

hydra/core/override_parser/types.py
omry: The quoted string objects are mostly used to reconstruct the quotes in some scenarios related to sweeping.
the objects that are merged to the config should be the string without the quotes.

For testing purposes, I do want to differentiate between the two types of quotes.

I think it's best to revert it.

hydra/core/override_parser/types.py
omry: Can you create a lower level test for this?

tests/test_hydra.py
omry: Great.
There is a test in test_overrides_parser that is dedicated for that logic: `test_override_get_value_element_method`

I think testing there should be sufficient.

omry (changes requested):

hydra/core/override_parser/types.py
odelalleau: > I think it's best to revert it.

Done in e249eb5ff. Note that I made `QuotedString` immutable since the dataclass doc suggests it is safer for hashable types (see `unsafe_hash` in https://docs.python.org/3/library/dataclasses.html).

hydra/core/override_parser/types.py
odelalleau: > Can you create a lower level test for this?

So, currently the lower level test is the one I moved in 6bcaac207 (`test_dict_key_formats()`, is it lower level enough?

tests/test_hydra.py
odelalleau: > I think testing there should be sufficient.

Perfect! Done in 8084317f1

At the same time I uncovered a bug with escaped characters (that affected regular unquoted strings as well, not just dictionary keys), which is now fixed in that same commit.

odelalleau:
Side note: I removed a print statement that seemed out of place in fc874646ea0243a553e4bb804014e73b4e163620

hydra/core/override_parser/types.py
omry: that's fine because I don't actually think we are changing it after construction.

hydra/core/override_parser/overrides_visitor.py
omry: dunno why I keep missing comments in fucking GitHub.
Yes, it's an extension of a generated class.
changing the variable types in a subclass is a code smell that will likely get flagged by automated tool as an error (if not now than once they pick up on my ultimate wisdom :) ).

tests/test_config_loader.py
```diff
@@ -593,6 +593,38 @@ def test_sweep_config_cache(
...
+            pytest.param("a/-\\+.$%*@", "a/-\\+.$%*@", id="unquoted_misc_chars"),
+            pytest.param("white space", "white space", id="whitespace"),
+            pytest.param(
+                "\\\\\\(\\)\\[\\]\\{\\}\\:\\=\\ \\\t\\,",
+                "\\()[]{}:= \t,",
+                id="unquoted_esc",
+            ),
+        ],
+    )
+    def test_dict_key_formats(
```
omry, line 613: I think with the the changes to test_override_get_value_element_method we no longer need this test. let me know if you think it's testing anything not covered by that test.

omry (changes requested):

hydra/_internal/grammar/utils.py
```diff
@@ -1,9 +1,37 @@
...
 from omegaconf._utils import is_dict_annotation, is_list_annotation
 
+# All characters that must be escaped (must match the ESC grammar lexer token).
+_ESC = "\\()[]{}:=, \t"
+
+# Regular expression that matches any sequence of characters in `_ESC`.
+_ESC_REGEX = re.compile(f"[{re.escape(_ESC)}]+")
+
+
+def escape(s: str) -> str:
```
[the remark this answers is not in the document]
odelalleau: Done in 4af259e03

hydra/core/override_parser/types.py
```diff
@@ -413,15 +417,17 @@ def _get_value_element_as_str(
...
-                    f"{k}{colon}{Override._get_value_element_as_str(v, space_after_sep=space_after_sep)}"
+                    f"{Override._get_value_element_as_str(k)}{colon}"
+                    f"{Override._get_value_element_as_str(v, space_after_sep=space_after_sep)}"
                     for k, v in value.items()
                 ]
             )
             return "{" + s + "}"
-        elif isinstance(value, (str, int, bool, float)):
+        elif isinstance(value, str):
+            return escape(value)  # ensure special characters are escaped
```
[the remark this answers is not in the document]
odelalleau: Renamed in 4af259e03 while adding test

hydra/core/override_parser/types.py
```diff
@@ -142,9 +143,9 @@ def __eq__(self, other: Any) -> Any:
             return NotImplemented
 
 
-# Ideally we would use List[ElementType] and Dict[str, ElementType] but Python does not seem
-# to support recursive type definitions.
-ElementType = Union[str, int, float, bool, List[Any], Dict[str, Any]]
+# Ideally we would use List[ElementType] and Dict[ElementType, ElementType] but Python
```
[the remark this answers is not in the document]
odelalleau: Right, actually for now they are still strings so I just reverted this change (cfe3932fc)

tests/test_config_loader.py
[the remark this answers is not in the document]
odelalleau: > for new tests, I prefer: (...)

Noted, hopefully I won't forget :)
Resolving since this test is now gone.

odelalleau: So this test was still covering some scenarios that weren't covered elsewhere, however, after looking at it more closely, I think I finally found where it should live (I think)
See 2e545f9ec for the new test (hopefully now in the correct place)

hydra/core/override_parser/types.py
```diff
@@ -259,8 +259,14 @@ def _convert_value(value: ParsedElementType) -> Optional[ElementType]:
         if isinstance(value, list):
             return [Override._convert_value(x) for x in value]
         elif isinstance(value, dict):
+
+            # Currently only strings are allowed as dictionary keys.
+            def check_str(k: Any) -> str:
+                assert isinstance(k, str)
+                return k
+
```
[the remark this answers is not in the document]
odelalleau: Oh, that's a good point! I know your comment was related to the wording, but it made more sense to me to revert the dict key type to `str` since that's the only type that is actually supported right now (once https://github.com/omry/omegaconf/pull/454 is merged it should be changed to `DictKeyType`, but we don't have that yet). And once I reverted it to `str` I had to fix `mypy` complaints and that was the purpose of `check_str`.

I ended up just removing the type check in 0d745041a, this way we let OmegaConf accept / refuse keys instead of checking it in Hydra.

tests/test_config_loader.py
omry: If you feel that is covering the functionality you want to test I am happy this being covered at that level.

tests/test_internal_grammar.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
omry, line 1: this test can go into tests/test_overrides_parser.py.

tests/test_config_loader.py
odelalleau: Yes, that sounds right to me => resolving

tests/test_internal_grammar.py
odelalleau: Yup, just did that in 0c894b3de as mentioned elsewhere :)

hydra/core/override_parser/types.py
odelalleau: > Still feels misplaced. After adding the lower level tests in the grammar parser, do you still thank that test_dict_key_formats() is needed?

As discussed elsewhere, the new test that covers the change on that line is the one from 2e545f9ec

hydra/core/override_parser/overrides_visitor.py
```diff
@@ -77,6 +77,11 @@ def is_ws(self, c: Any) -> bool:
 
     def visitPrimitive(
         self, ctx: OverrideParser.PrimitiveContext
+    ) -> Optional[Union[QuotedString, int, bool, float, str]]:
+        return self.visitPrimitiveOrDictKey(ctx)
+
+    def visitPrimitiveOrDictKey(
+        self, ctx: Union[OverrideParser.PrimitiveContext, OverrideParser.DictKeyContext]
```
omry, line 84: I propose 
```
def _createPrimitive(self, ctx : antlr.ParserRuleContext) ->  Optional[Union[QuotedString, int, bool, float, str]]
```

hydra/core/override_parser/types.py
```diff
@@ -413,15 +420,17 @@ def _get_value_element_as_str(
         elif isinstance(value, dict):
             s = comma.join(
                 [
-                    f"{k}{colon}{Override._get_value_element_as_str(v, space_after_sep=space_after_sep)}"
+                    f"{Override._get_value_element_as_str(k)}{colon}"
+                    f"{Override._get_value_element_as_str(v, space_after_sep=space_after_sep)}"
                     for k, v in value.items()
                 ]
```
omry, line 426: For my sanity reading this later, can you refactor a bit?
I think at this point a regular loop with temp variables will be much easier to parse.

omry (commented):
Looks good, a couple of small comments.

hydra/core/override_parser/overrides_visitor.py
odelalleau: Done in 8a16f54e4

hydra/core/override_parser/types.py
odelalleau: Done in 4f2665eec

omry:
Looking good @odelalleau, I am ready to merge it.

omry:
@jieru-hu , probably unrelated to this PR:
Any idea about the Ray Plugin failure? 
https://app.circleci.com/pipelines/github/facebookresearch/hydra/5566/workflows/96cbbc29-ace4-49e6-a13d-a26d406697ce/jobs/40966

fuhuifang:
Hi, how to parse something like `key={/-\\+.$%*@: 1}` from command line? I'm getting "error: argument --info/-i: ignored explicit argument 'nputFormat'". 

I tried a simpler dictKey like `key={a:1}` and it works.

odelalleau:
> Hi, how to parse something like `key={/-\\+.$%*@: 1}` from command line? I'm getting "error: argument --info/-i: ignored explicit argument 'nputFormat'".

Using single quotes `'key={/-\\+.$%*@: 1}'` seems to be working for me. That being said, note that the support for quoted strings in keys (e.g., `'key={"a^b": 1}'`) that was introduced in this PR has eventually been reverted in Hydra 1.1, so you are limited by the set of (possibly escaped) characters supported in unquoted strings.

fuhuifang:
> > Hi, how to parse something like `key={/-\\+.$%*@: 1}` from command line? I'm getting "error: argument --info/-i: ignored explicit argument 'nputFormat'".
> 
> Using single quotes `'key={/-\\+.$%*@: 1}'` seems to be working for me. That being said, note that the support for quoted strings in keys (e.g., `'key={"a^b": 1}'`) that was introduced in this PR has eventually been reverted in Hydra 1.1, so you are limited by the set of (possibly escaped) characters supported in unquoted strings.

Thanks so much for the quick response. Single quotes doesn't work for me. I just found a working solution: `"key='{/-\\+.$%*@: 1}'"`.

omry:
Single quote does work, but you need to note that your shell is also intercepting the quotes.
See https://hydra.cc/docs/next/advanced/override_grammar/basic.

odelalleau:
> Single quotes doesn't work for me. I just found a working solution: `"key='{/-\\+.$%*@: 1}'"`

Hmm, I think either:
- Your shell doesn't handle single quotes the same as mine (I'm using zsh), or
- You are not using (a release candidate of) Hydra 1.1

The solution you found gives a different result: it sets `key` to the *string* `"{/-\\+.$%*@: 1}"` instead of it being a dictionary.

To check your shell behavior, you can do:
```bash
> python -c "import sys; print(sys.argv[1])" 'key={/-\\+.$%*@: 1}'
key={/-\\+.$%*@: 1}
```

fuhuifang:
> > Single quotes doesn't work for me. I just found a working solution: `"key='{/-\\+.$%*@: 1}'"`
> 
> Hmm, I think either:
> 
> * Your shell doesn't handle single quotes the same as mine (I'm using zsh), or
> * You are not using (a release candidate of) Hydra 1.1
> 
> The solution you found gives a different result: it sets `key` to the _string_ `"{/-\\+.$%*@: 1}"` instead of it being a dictionary.
> 
> To check your shell behavior, you can do:
> 
> ```shell
> > python -c "import sys; print(sys.argv[1])" 'key={/-\\+.$%*@: 1}'
> key={/-\\+.$%*@: 1}
> ```

Hello, I'm getting the same behavior as you when running
` python -c "import sys; print(sys.argv[1])" 'key={/-\\+.$%*@: 1}'`. 

However, when running my real scenario
`python run.py --config-dir pipelines/config --config-name experiments/data_preprocess " 'compute.hdi_conf={spark.yarn.maxAppAttempts:2}'`, I'm getting
no viable alternative at input '{spark.'

Do you know how to fix it? "compute.hdi_conf" is the property I want to override.

odelalleau:
> This works for me, but Hydra 1.1 is not compatible with our code. Do you know it will be released?

There are indeed a bunch of changes from 1.0 to 1.1 you'll need to take care of (you can start here: https://hydra.cc/docs/next/upgrades/intro)

AFAIK the official release of 1.1 is still a few weeks from now. You can give the RC1 a shot (see release notes at https://github.com/facebookresearch/hydra/releases/tag/v1.1.0.rc1) but there might still be a few rough edges, so don't use it in a critical application just yet.

If you want to stick to 1.0 I'm afraid there aren't many options -- you might be able to hack something around with a custom resolver but the 1.0 resolvers are a bit limited so it probably wouldn't be great. A simpler temporary hack (until you upgrade to 1.1) could be to replace whatever character that is not allowed in 1.0 with `_` and manually replace it at the very beginning of your program... please don't keep such a hack for too long though ;)

omry:
At this point I recommend that you migrate to Hydra 1.1 by installing the first release candidate and fixing all the warnings.
In general it should be pretty smooth. If you run into a wall feel free to file an issue with a question (ideally with a minimal repro).

Issue #1227, opened 2020-12-22
[Bug] Submitit Handling of ListConfigs


jgbos:
# 🐛 Bug
## Description

I'm currently using your master branch (love the direction this is going) and have found a bug that may exist for your released version as well.  If I add the following parameter to my `SlurmLauncher` config:

```yaml
setup:
  - export MASTER_ADDR=$(hostname -s)
  - export MASTER_PORT=$(python -c 'import socket; s=socket.socket(); s.bind(("", 0)); print(s.getsockname()[1]); s.close()')
```

I get the following error:

```
Traceback (most recent call last):
  File "/home/gridsan/JU17693/.conda/envs/raiden-release/lib/python3.7/site-packages/hydra/_internal/utils.py", line 196, in run_and_report
    return func()
  File "/home/gridsan/JU17693/.conda/envs/raiden-release/lib/python3.7/site-packages/hydra/_internal/utils.py", line 356, in <lambda>
    overrides=args.overrides,
  File "/home/gridsan/JU17693/.conda/envs/raiden-release/lib/python3.7/site-packages/hydra/_internal/hydra.py", line 137, in multirun
    return sweeper.sweep(arguments=task_overrides)
  File "/home/gridsan/JU17693/.conda/envs/raiden-release/lib/python3.7/site-packages/hydra/_internal/core_plugins/basic_sweeper.py", line 154, in sweep
    results = self.launcher.launch(batch, initial_job_idx=initial_job_idx)
  File "/home/gridsan/JU17693/.conda/envs/raiden-release/lib/python3.7/site-packages/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py", line 122, in launch
    executor.update_parameters(**params)
  File "/home/gridsan/JU17693/.conda/envs/raiden-release/lib/python3.7/site-packages/submitit/core/core.py", line 638, in update_parameters
    self._internal_update_parameters(**kwargs)
  File "/home/gridsan/JU17693/.conda/envs/raiden-release/lib/python3.7/site-packages/submitit/auto/auto.py", line 197, in _internal_update_parameters
    self._executor._internal_update_parameters(**parameters)
  File "/home/gridsan/JU17693/.conda/envs/raiden-release/lib/python3.7/site-packages/submitit/slurm/slurm.py", line 307, in _internal_update_parameters
    _make_sbatch_string(command="nothing to do", folder=self.folder, **kwargs)
[...]

PR #1230, opened 2020-12-23
Support submitit setup parameter

jgbos:
## Motivation

Add support to Submitit Slurm parameter `setup`.

Yes

## Test Plan

All plugin tests pass.

## Related Issues and PRs

See #1227

### DIFF
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
@@ -115,17 +115,14 @@ class BaseSubmititLauncher(Launcher):
         # specify resources/parameters
         baseparams = set(OmegaConf.structured(BaseQueueConf).keys())
         params = {
-            x if x in baseparams else f"{self._EXECUTOR}_{x}": y
+            x
+            if x in baseparams
+            else f"{self._EXECUTOR}_{x}": y
+            if not OmegaConf.is_config(y)
+            else OmegaConf.to_container(y, resolve=True)
             for x, y in params.items()
             if x not in init_keys
         }
-        # submitit `setup` parameter must be a primitive list container
-        if f"{self._EXECUTOR}_setup" in params and isinstance(
-            params[f"{self._EXECUTOR}_setup"], ListConfig
-        ):
-            params[f"{self._EXECUTOR}_setup"] = OmegaConf.to_container(
-                params[f"{self._EXECUTOR}_setup"], resolve=True
-            )
         executor.update_parameters(**params)
 
         log.info(
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
@@ -11,7 +11,7 @@ from hydra.core.singleton import Singleton
 from hydra.core.utils import JobReturn, filter_overrides, run_job, setup_globals
 from hydra.plugins.launcher import Launcher
 from hydra.plugins.search_path_plugin import SearchPathPlugin
-from omegaconf import DictConfig, OmegaConf, open_dict
+from omegaconf import DictConfig, ListConfig, OmegaConf, open_dict
 
 from .config import BaseQueueConf
 
@@ -119,6 +119,9 @@ class BaseSubmititLauncher(Launcher):
             for x, y in params.items()
             if x not in init_keys
         }
+        # submitit `setup` parameter must be a primitive list container
+        if "setup" in params and isinstance(params["setup"], ListConfig):
+            params["setup"] = OmegaConf.to_container(params["setup"], resolve=True)
         executor.update_parameters(**params)
 
         log.info(
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
@@ -11,7 +11,7 @@ from hydra.core.singleton import Singleton
 from hydra.core.utils import JobReturn, filter_overrides, run_job, setup_globals
 from hydra.plugins.launcher import Launcher
 from hydra.plugins.search_path_plugin import SearchPathPlugin
-from omegaconf import DictConfig, ListConfig, OmegaConf, open_dict
+from omegaconf import DictConfig, OmegaConf, open_dict
 
 from .config import BaseQueueConf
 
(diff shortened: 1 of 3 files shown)


plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -119,6 +119,13 @@ def launch(
             for x, y in params.items()
             if x not in init_keys
         }
+        # submitit `setup` parameter must be a primitive list container
+        if f"{self._EXECUTOR}_setup" in params and isinstance(
+            params[f"{self._EXECUTOR}_setup"], ListConfig
+        ):
+            params[f"{self._EXECUTOR}_setup"] = OmegaConf.to_container(
```
jieru-hu, line 126: what happens when setup = None?

also why do you need to check if its ListConfig? is it to get past mypy?

jgbos: Won't the if statement be false if it's None?  

I've been reading through the code trying to get a better handle of the assumptions for initializing the Launcher.  The code doesn't indicate that the input for initialization comes from configuration objects, can I assume this? If so, in the `__init__` function could I just do:

```python
self.params["setup"] = OmegaConf.to_container(self.params["setup"], resolve=True)
```

omry: A simpler approach is to convert all the params to primitive containers.
This seems to work:
```diff
diff --git a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
index 1b3635c7e..b862ab5b1 100644
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
@@ -114,12 +114,17 @@ class BaseSubmititLauncher(Launcher):
 
         # specify resources/parameters
         baseparams = set(OmegaConf.structured(BaseQueueConf).keys())
-        params = {
-            x if x in baseparams else f"{self._EXECUTOR}_{x}": y
-            for x, y in params.items()
-            if x not in init_keys
-        }
-        executor.update_parameters(**params)
+
+        params = OmegaConf.create(
+            {
+                x if x in baseparams else f"{self._EXECUTOR}_{x}": y
+                for x, y in params.items()
+                if x not in init_keys
+            }
+        )
+
+        plain_params = OmegaConf.to_container(params, resolve=True)
+        executor.update_parameters(**plain_params)
 
         log.info(

```

You can also convert them inside the existing overly fancy dict comprehension loop (slightly more efficient but much more convoluted, at least in the current form of things.

omry: You can also just convert all the OmegaConf containers in params to containers.

omry:
Please explain how you are testing this.

jgbos:
I ran the tests in the folder and I also run a local (no setup) and cluster job (with setup) on our system

jgbos:
So now I see that the plugin never "checks the schema".

jgbos:
What do you think of adding these line to each of the Launcher classes `__init__` method (similar to the RQ Launcher code):

For `BaseSubmititLauncher`

```python
self.params = OmegaConf.structured(BaseQueueConf(**params))
````

For `LocalSubmititLauncher`

```python
self.params = OmegaConf.structured(LocalQueueConf(**params))
````

For `SlurmSubmititLauncher`

```python
self.params = OmegaConf.structured(SlurmQueueConf(**params))
self.params.setup = OmegaConf.to_container(self.params.setup, resolve=True)
````

**Comment by omry** [comment]
> So now I see that the plugin never "checks the schema".
Explain.

did you see my comment about additional_parameters not being checked?

Please show the output of `--cfg all -p hydra.launcher` when you are overriding setup before this fix attempt.

**Comment by omry** [comment]
1. You still have CI failures.
2. I was thinking about simplifying the code by converting everything passed to submitit to a primitive container.

**Comment by jgbos** [comment]
whoops, sorry.  Forgot that pushing to my fork updates this automatically.  Doing some internal tests across different platforms.  Obviously a novice at this.

1. I guess I need to dig into why ray tests are failing
2. My goal was to convert any config object to primitive containers with this line:

https://github.com/jgbos/hydra/blob/submitit-launcher-setup/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py#L122

Not a big fan of the if statements there, not very clear what's going on.

**Comment by omry** [comment]
> I guess I need to dig into why ray tests are failing

Ray failures are unrelated and should be resolved soon. I missed this being only Ray that is failing.

**Comment by omry** [comment]
see #1269. try to make it work with it.

In Hydra 1.1, we should be able to use `_convert_: ALL` as an alternative. looks like here is some unrelated issue with it in plugins at the moment.

**Comment by jgbos** [comment]
Ah yes, thanks! #1269 was what I was thinking awhile ago but I the wasn't sure what I could assume for the input given it's typed with `Any`, I also didn't know about `OmegaConf.is_config` at the time.

**Review by omry — COMMENTED** [review]
Cool.
Can you add a news fragment in the news directory in the plugin?

See https://hydra.cc/docs/development/documentation#news-entries

**Comment by omry** [comment]
1. Please rebase. all tests should pass.
2. Is there a link to setup in the slurm doc we can use in the news fragment?

**Comment by jgbos** [comment]
No mention of `setup` in the docs unfortunately.  Just here in the code:

https://github.com/facebookincubator/submitit/blob/master/submitit/slurm/slurm.py#L429

**Comment by omry** [comment]
> No mention of `setup` in the docs unfortunately. Just here in the code:
> 
> https://github.com/facebookincubator/submitit/blob/master/submitit/slurm/slurm.py#L429

Is this a submitit functionality of SLURM functionality?

**Comment by jgbos** [comment]
It's for SLURM, specifically the SBATCH script.

**Comment by omry** [comment]
Not my question:
I am trying to understand if there is a documentation of a setup concept in SLURM or if this is a submitit concept.

The news file should be something like:
```
Add support for [Submitit's setup field](link to doc or code)
```

Or:
```
Add support for [SLURM sbatch setup](link to doc or code)
```

(No need to mention the conversion to primitive containers, this is not relevant to users).

jgbos:
Sorry, yes, this is a submitit concept to allow the user to provide additional lines of code in the SBATCH script.  This would be something Submitit would need to document but has not.  So I would have to link to the code.

omry (commented):
looks like this fell off the radar.
Are we ready to integrate it?

omry (approved):
lgtm, as far as I am concerned we can merge this.

omry:
@shagunsodhani , would you mind releasing an updated version?
(remember to use towncrier to generate a news file).

jgbos:
Thanks! I had assumed it was done because I've been working off the master branch, but that just means `hydra-core` and not the plugin which was just my local version 🤦.  Glad I was able to contribute and thanks for working with me on getting it right.

Issue #1183, opened 2020-12-04
[Bug] Impossible to instantiate nested dataclasses 


guillaumegenthial:
# 🐛 Bug
## Description

I want to instantiate recursively nested dataclass instances.

Because OmegaConf considers dataclasses instances to be configs, when calling `_get_kwargs` during the `instantiate` call, the dataclass instance is converted back to a config because parameters are added to a OmegaConf list / dict.

See https://github.com/facebookresearch/hydra/blob/c1eb2435b6314412d7029acbf87efd63d84d384c/hydra/_internal/utils.py#L685

## Checklist
- [x] I checked on the latest version of Hydra
- [x] I created a minimal repro

## To reproduce

In a script `my_app.py`

```python
"""Test hydra dataclass instantiation."""

from dataclasses import dataclass
from omegaconf import OmegaConf
from hydra.utils import instantiate

@dataclass
class A:
    x: int

@dataclass
class B:
    a: A

conf = {
    "_target_": "__main__.B",
    "a": {
        "_target_": "__main__.A",
        "x": 1
    }
}
parsed = OmegaConf.create(conf)
instance = instantiate(parsed)
print(instance)
```

run `python my_app.py`.

You get

```bash
B(a={'x': 1})
```

## Expected Behavior

I expect the result to be `B(a=A(x=1))`.

It seems strange to me that the config schema interferes with the instantiation.

I work with a codebase where a bunch of classes are dataclasses and using hydra for configuration would be a plus.

## System information
- **Hydra Version** :  master
- **Python version** :  3.6
- **Virtual environment type and version** : instantiated with `python -m venv`
- **Operating system** :  macOS 11.0.1

jieru-hu:
Thanks for the report!

I tried this locally and it seems like `A` was actually converted to the correct type however when it was added to `B` the typing info got lost. I'm wondering if this is something more related to `omegaconf` side.

cc @omry

guillaumegenthial:
Thanks for the quick response.

I don't think it's an issue with omegaconf (it's the intended behavior from omegaconf when adding a dataclass instance to a `OmegaConf` object, as listed in the doc).

I am rather wondering why the `_get_kwargs` function uses `OmegaConf` objects to store instantiated objects.

Why not simply unpack kwargs from the cong, instantiate them in a regular python dict?

omry:
@guillaumegenthial, thanks for reporting.
This is indeed a limitation of the current implementation.

[...]

Issue #1200, opened 2020-12-13
[Feature Request] Recursive instantiate can be slow due to converting containers to configs


ppwwyyxx:
# 🚀 Feature Request

`instantiate(cfg)` always convert python containers to configs. If a subtree in `cfg` instantiates to a large python container, such conversion can be very slow. Example:

```
!wget http://images.cocodataset.org/annotations/annotations_trainval2017.zip && unzip annotations_trainval2017.zip
import json
import time
from omegaconf import OmegaConf
from hydra.utils import instantiate

class DataLoader():
    def __init__(self, raw_data):
        self.data = raw_data

def get_data():
    with open("annotations/instances_val2017.json") as f:
        data = json.load(f)
    return data

cfg = OmegaConf.create({
        "_target_": "__main__.DataLoader",
        "raw_data": {
            "_target_": "__main__.get_data"
        },
    }, flags={"allow_objects": True})

_ = get_data()

start = time.time()
instantiate(cfg)
print(time.time() - start)   # >100s

start = time.time()
a = DataLoader(get_data())
print(time.time() - start) # 0.3s
```

## Pitch
It's better if objects created by subtrees are passed to parents without unnecessary conversion. This also makes the behavior `instantiate()` matches better with its equivalent python code (`DataLoader(raw_data=get_data())`).

Related to `_convert_`, but container objects are now always converted to configs, and  `_convert_` only controls whether to convert them back to containers.

omry:
There are many features that depends on this conversion.
There is [room for performance improvements](https://github.com/omry/omegaconf/issues/267) in OmegaConf but generally speaking this is not high priority at this time.

If you can send a PR with an alternative implementation of instantiate that is faster and maintains all the existing functionality (meaning - relevant tests are passing) I will consider including it.

Let me know if you would like to work on it. If not I will close this issue.

omry:
Please evaluate performance of OmegaConf 2.1 (`omegaconf>=2.1.0dev20`). There are some very significant perf improvements.

omry:
Running your script:
```
$ python 1.py 
0.3778083324432373
0.39718198776245117
```

omry:
@ppwwyyxx.
This is primarily due to performance improvements in OmegaConf.

What you are describing sounds unexpected. can you file a new bug report with a minimal repro?

omry:
Having dug a bit deeper I understand why this is happening and it is expected.
[...]

PR #1236, opened 2020-12-24
Refactor Instantiate to support instantiating dataclasses

guillaumegenthial:
## Motivation

Address https://github.com/facebookresearch/hydra/issues/1183

Yes

## Related Issues and PRs

https://github.com/facebookresearch/hydra/issues/1183

### DIFF
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -1,5 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
 import argparse
+import collections.abc
 import inspect
 import logging.config
 import os
@@ -10,8 +12,7 @@ from traceback import print_exc, print_exception
 from types import FrameType
 from typing import Any, Callable, List, Optional, Sequence, Tuple, Union
 
-from omegaconf import DictConfig, ListConfig, OmegaConf, open_dict
-from omegaconf._utils import type_str
+from omegaconf import OmegaConf
 from omegaconf.errors import OmegaConfBaseException
 
 from hydra._internal.config_search_path_impl import ConfigSearchPathImpl
@@ -574,30 +575,21 @@ def _locate(path: str) -> Union[type, Callable[..., Any]]:
 
 
 def _is_target(x: Any) -> bool:
+    if isinstance(x, dict):
+        return "_target_" in x
     if OmegaConf.is_dict(x) and not OmegaConf.is_none(x):
         return "_target_" in x
     return False
 
 
-def _is_recursive(config: Any, kwargs: Any) -> bool:
-    def _is_rec(d: Any) -> Optional[bool]:
-        if "_recursive_" in d:
-            rec = d.pop("_recursive_")
-            if not isinstance(rec, bool):
-                raise ValueError(
-                    f"_recursive_ flag must be a bool, got {type_str(rec)}"
-                )
-            return rec
-        return None
-
-    # pop both in any case
-    kwrec = _is_rec(kwargs)
-    configrec = _is_rec(config)
-    if kwrec is not None:
-        return kwrec
-    elif configrec is not None:
-        return configrec
-    return True
+def _call_target(target: Callable, *args, **kwargs) -> Any:  # type: ignore
+    """Call target (type) with args and kwargs."""
+    try:
+        return target(*args, **kwargs)
+    except Exception as e:
+        raise type(e)(
+            f"Error instantiating '{_convert_target_to_string(target)}' : {e}"
+        ).with_traceback(sys.exc_info()[2])
 
 
 def _convert_target_to_string(t: Any) -> Any:
@@ -613,7 +605,7 @@ def _convert_container_targets_to_strings(d: Any) -> None:
     if isinstance(d, dict):
         if "_target_" in d:
             d["_target_"] = _convert_target_to_string(d["_target_"])
-        for k, v in d.items():
+        for _, v in d.items():
             _convert_container_targets_to_strings(v)
     elif isinstance(d, list):
         for e in d:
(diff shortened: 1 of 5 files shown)


omry:
Thanks @guillaumegenthial!
It will take a few weeks before I can really review it (It looks good at a quick glance but I will have a lot of feedback).
we can discuss some of my feedback on the chat to make things more efficient.

Don't forget to sign the CLA (see message from the bot above, I can't accept it otherwise).

guillaumegenthial:
@omry sure thing, don't hesitate to ping me when you have some time.

To summarize the changes
1. remove kwargs, put all the logic in instantiate
2. handle edge cases carefully (depending on convert, config type, etc.)
3. Add a new ParseMode (not necessary for this PR), that allows the user to avoid the conversion to OmegaConf container and operate on simple python containers (dict + list).

I updated the dataclass test, and I think it now behaves as we would expect. Some cases are counter intuitive (convert = NONE and nesting `dict+dataclass` or `list+dataclass`), as dataclasses instances have no choice but to be absorbed by the top-level OmegaConf container.

omry:
Some high level feedback before I do a proper review based on your summary:

> 1. remove `_get_kwargs`, put all the logic in instantiate
Sure, it becomes very long though so it would still be good to break some parts into smaller functions to make it more readable.
(Once the rest of the feedback is incorporated)

> 3. Add mode inheritance (parent modes are passed to the children when recursively instantiating)

I am actually not sure about this one but I am guessing you can still override the mode in the child node?

> 4. Add `ParseMode` (not necessary for this PR), that allows the user to avoid the conversion to OmegaConf container and operate on simple python containers (dict + list).

While I definitely want better support for simple containers, two things:
- This opens up a surface area that will need proper testing (and proper definition of what happens with interpolation and type safety in dict mode).
- It's a user facing change that should be documented.
- I think the parse mode can be determined automatically based on the type of the input config object. (dict or DictConfig).

We should probably split it to a followup PR once the initial changes settles in.

Finally, some of the CI tests are tailing (particularly plugins, which are being instantiated with instantiate(). be sure to take a look and fix the issues.

guillaumegenthial:
> I am actually not sure about this one but I am guessing you can still override the mode in the child node?

Yes, children modes override the parent's modes if specified

> We should probably split it to a followup PR once the initial changes settles in.

Definitely agree, I included it here because it's easy to do with the new implementation (it kind of demonstrates its flexibility)

> Finally, some of the CI tests are tailing (particularly plugins, which are being instantiated with instantiate(). be sure to take a look and fix the issues.

I see some "version mismatch" errors, I will have a look at it to understand how instantiate has an impact there.

guillaumegenthial:
Ok, I'm pretty sure the failing ray launcher plugin tests are not caused by this PR.
A new version of ray was released 16 hours ago.
The ci instance seems to have both versions co-existing somehow (1.1.0, the new one and 1.0.1.post1, the previous one). Not sure why, maybe just some cache that's not been cleared? I see that the wheels paths are set explicitely in the code to `/tmp/wheels/`, maybe that's the reason...

omry:
Yup, unrelated.

guillaumegenthial:
Simple comment : `ParseMode` is one way to address https://github.com/facebookresearch/hydra/issues/1200

omry:
> Simple comment : `ParseMode` is one way to address #1200

Certainly (assuming that use case is really not depending on type validation, interpolation or any other OmegaConf feature).

hydra/utils.py
```diff
@@ -53,67 +104,184 @@ def instantiate(config: Any, *args: Any, **kwargs: Any) -> Any:
...
-        convert = _pop_convert_mode(final_kwargs)
-        if convert == ConvertMode.PARTIAL:
-            final_kwargs = OmegaConf.to_container(
-                final_kwargs, resolve=True, exclude_structured_configs=True
+    # Finalize config (convert targets to strings, merge with kwargs)
+    # TODO: kwargs are used as a config now. If it contains dataclass
+    # objects, they will be interpreted as config nodes in the
+    # instantiation result. We might want to change that behavior.
+    _convert_container_targets_to_strings(kwargs)
+    _convert_container_targets_to_strings(config)
```
omry, line 150: Could be a good opportunity to simplify this.
Now that OmegaConf supports objects, we should be able to just convert forward.
instead of converting types to strings (and then locate them again), we can convert everything to types.

hydra/utils.py
```diff
@@ -53,67 +104,184 @@ def instantiate(config: Any, *args: Any, **kwargs: Any) -> Any:
...
+    config = _merge(config, kwargs) if kwargs else config
+
+    # Override parent modes from config if specified
+    # Cannot use pop with default because of OmegaConf typed config
```
omry, line 154: Not sure I understand the statement.
if DictConfig.pop() behaves differently than native dict pop(), please file an issue with a minimal repro against OmegaConf.

hydra/utils.py
```diff
@@ -53,67 +104,184 @@ def instantiate(config: Any, *args: Any, **kwargs: Any) -> Any:
...
+            # If ALL or PARTIAL non structured, instantiate in dict
+            if convert == ConvertMode.ALL or (
+                convert == ConvertMode.PARTIAL and config._metadata.object_type is None
+            ):
+                return {
+                    key: instantiate(
+                        value, parse=parse, convert=convert, recursive=recursive
+                    )
+                    for key, value in config.items_ex(resolve=True)
+                }
```
omry, line 253: I think dict/list comprehension is no longer elegant once it starts to get this complex.
can you switch to a simple loop populating a `ret = {}` object?

hydra/utils.py
```diff
@@ -53,67 +104,184 @@ def instantiate(config: Any, *args: Any, **kwargs: Any) -> Any:
...
+            return _call_target(target, *args, **config)
+
+        if recursive:
+            # If ALL or PARTIAL non structured, instantiate in dict
+            if convert == ConvertMode.ALL or (
+                convert == ConvertMode.PARTIAL and config._metadata.object_type is None
+            ):
```
omry, line 247: I don't like how PARTIAL is leaking into this area. somehow this feels off.

tests/test_utils.py
```diff
@@ -979,33 +967,109 @@ def test_convert_params(input_: Any, expected: Any, convert_mode: Any) -> None:
                     "b": None,
                 },
             },
-            SimpleDataClass(a={"foo": 99}, b=[1, 99]),
-            id="simple",
+            (
+                SimpleDataClass(
```
omry, line 971: You seems to have dropped the simple case that was there initially.

omry (changes requested):
Initial review pass.
see my comments.

meta:
At this point, switch to piling additional changes as new commits. it will make this easier to review.
We can squash everything before we merge.

hydra/_internal/utils.py
```diff
@@ -613,140 +605,62 @@ def _convert_container_targets_to_strings(d: Any) -> None:
...
+    This allows us not to do copies of input objects but requires the
+    caller to be aware of the input config mutation during the call.
+    """
+    # Use OmegaConf merge if config is OmegaConf container
+    if OmegaConf.is_config(config):
+        config.merge_with(OmegaConf.create(overrides, flags={"allow_objects": True}))
+        return config
+
+    # Manual Override if native mapping type
+    if isinstance(config, collections.abc.MutableMapping):
```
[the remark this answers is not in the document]
guillaumegenthial: done

hydra/_internal/utils.py
```diff
@@ -613,140 +605,62 @@ def _convert_container_targets_to_strings(d: Any) -> None:
...
-        kwargs_target = kwargs.pop("_target_")
+def _merge(config: Any, overrides: Any) -> Any:
+    """Mutate and return config with overrides.
 
-    if "_target_" in config:
-        config_target = config.pop("_target_")
+    NOTE: This is an anti-pattern since we mutate the config and return.
+    This allows us not to do copies of input objects but requires the
+    caller to be aware of the input config mutation during the call.
+    """
```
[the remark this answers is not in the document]
guillaumegenthial: done

hydra/_internal/utils.py
```diff
@@ -613,140 +605,62 @@ def _convert_container_targets_to_strings(d: Any) -> None:
...
+    # Use OmegaConf merge if config is OmegaConf container
+    if OmegaConf.is_config(config):
+        config.merge_with(OmegaConf.create(overrides, flags={"allow_objects": True}))
+        return config
+
+    # Manual Override if native mapping type
+    if isinstance(config, collections.abc.MutableMapping):
+        if not isinstance(overrides, collections.abc.Mapping):
+            raise TypeError(f"Expected type dict but got {type(overrides)}")
+        config = dict(config.items())  # Shallow copy
```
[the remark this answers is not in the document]
guillaumegenthial: I removed the copy since this function should be ok mutating stuff. 
As a result, I added a recursive copy in `instantiate`.

hydra/_internal/utils.py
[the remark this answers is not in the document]
guillaumegenthial: I changed the function behavior.
It mutates and does not return, so we're all good.

hydra/_internal/utils.py
```diff
@@ -613,140 +605,62 @@ def _convert_container_targets_to_strings(d: Any) -> None:
...
+    # Manual Override if native sequence type
+    if isinstance(config, collections.abc.MutableSequence):
+        if not isinstance(overrides, collections.abc.Sequence):
+            raise TypeError(f"Expected type list but got {type(overrides)}")
+        for idx, item in enumerate(overrides):
+            if idx < len(config):
+                config[idx] = _merge(config[idx], item)
+            else:
+                config.append(item)
+        return config
```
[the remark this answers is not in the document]
guillaumegenthial: I checked OmegaConf behavior.
It overrides lists completely so I removed this to be consistent.
A use case I see is

```python
config = [{"x": 1, "y": 2}]
overrides = [{"x": 2}]
_merge(config, overrides)
assert config == [{"x": 2, "y": 2}]  # Would  be [{"x": 2}] with OmegaConf, the y disappeared.

```

hydra/_internal/utils.py
```diff
@@ -613,140 +605,62 @@ def _convert_container_targets_to_strings(d: Any) -> None:
...
+    # to unexpected behaviors (_merge({1: 2}, None) returns None).
+    return overrides
 
-    if target is None:
-        raise InstantiationException("Unable to determine target")
 
+def _resolve_target(
+    target: Union[str, type, Callable[..., Any]]
+) -> Union[type, Callable[..., Any]]:
+    """Pop target from config and import python type or callable."""
```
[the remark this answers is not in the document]
guillaumegenthial: done

hydra/utils.py
```diff
@@ -1,39 +1,90 @@
...
 
 
-class ConvertMode(Enum):
+class Keys(str, Enum):
+    """Special keys."""
+
+    TARGET = "_target_"
+    CONVERT = "_convert_"
+    PARSE = "_parse_"
+    RECURSIVE = "_recursive_"
```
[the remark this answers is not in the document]
guillaumegenthial: done

hydra/utils.py
guillaumegenthial: switched all multi-line comprehensions to standard assignments

tests/test_utils.py
guillaumegenthial: Now the `expected` test argument is the full instantiated config. It appears like it disappeared but it's still here.

hydra/utils.py
guillaumegenthial: I left a comment on an opened issue https://github.com/omry/omegaconf/issues/425.
I'm not sure this is really a bug, more like some type checking functionality of OmegaConf that forces us to "look before trying"

hydra/utils.py
```diff
@@ -1,39 +1,90 @@
...
+    during instantiate. This is the expected behavior, since the
+    interpolation syntax might be used in the native python dictionary.
+
+    This behavior can be disable with ParseMode.NONE. Please make sure
+    the config given to instantiate is fully defined (no interpolation).
+    """
+
+    NONE = "none"  # Keep original config type
+    NATIVE = "native"  # Convert to dict / list if OmegaConf config
+    OMEGACONF = "omegaconf"  # Convert config to OmegaConf container
```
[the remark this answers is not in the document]
guillaumegenthial: Yes, it should be really easy to remove. I'd just like to remove it once we settle some of the other discussions to be sure the implementation is flexible enough (see below for context)

hydra/utils.py
```diff
@@ -53,67 +104,184 @@ def instantiate(config: Any, *args: Any, **kwargs: Any) -> Any:
...
-        config._set_flag("allow_objects", True)
-        final_kwargs = _get_kwargs(config, **kwargs)
-        convert = _pop_convert_mode(final_kwargs)
-        if convert == ConvertMode.PARTIAL:
-            final_kwargs = OmegaConf.to_container(
-                final_kwargs, resolve=True, exclude_structured_configs=True
+    # Finalize config (convert targets to strings, merge with kwargs)
+    # TODO: kwargs are used as a config now. If it contains dataclass
+    # objects, they will be interpreted as config nodes in the
+    # instantiation result. We might want to change that behavior.
```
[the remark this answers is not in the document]
guillaumegenthial: This is the main friction point.

I'd argue that this would be confusing to treat dataclass instances in overrides as objects and not configs.

1. Instantiate should operate on config-like objects, why would the overrides already have some instantiated objects?
2. DataClass become sources of confusion. If config is `{"obj": {"x": 1, "y": 2}}` and kwargs is `obj=SomeDataClass(x=2)`, what do we expect? `{"obj": {"x": 2, "y": 2}}` (current behavior), or `{"obj": SomeDataClass(x=2)`

Also, (not necessarily a good argument), it would make the implementation more complex. 
The current implementation does
1. merge kwargs fully into config
2. forget about kwargs, only look at config

If we want to treat kwargs as objects, we can't do that because dataclasses instances merged into OmegaConf containers would be converted back to OmegaConf containers. This means that we should keep the kwargs alonside the config during the whole instantiation path.

hydra/utils.py
guillaumegenthial: I agree.

However it seems we can't really store types in OmegaConf containers.

```python

class MyClass:
    pass

OmegaConf.create({"_target_": MyClass})
```

raises

```bash
UnsupportedValueType: Value 'MyClass' is not a supported primitive type
	full_key: _target_
	reference_type=Optional[Dict[Union[str, Enum], Any]]
	object_type=dict
```

guillaumegenthial: actually my bad, I forgot the flag {"allow_objects"} so it works as expected

hydra/utils.py
guillaumegenthial: I opened a separate issue at your request here https://github.com/omry/omegaconf/issues/471

guillaumegenthial:
Thank you for reviewing @omry .
I pushed some changes addressing some of your comments.
Before addressing the others, we need to decide how we should treat kwargs, since it has a potentially big impact on the implementation (see my comment above).

hydra/utils.py
guillaumegenthial: There is some ambiguity with dataclasses though

```python
@dataclass
class MyClass:
    x: int


cfg = OmegaConf.create({"_target_": MyClass}, flags={"allow_objects": True})
print(cfg._target_)  # {'x': '???'}
```

hydra/_internal/utils.py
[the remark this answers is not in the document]
omry: I agree.

hydra/utils.py
omry: That again, sounds like another case fox boxing the object.

omry:
> Thank you for reviewing @omry .
> I pushed some changes addressing some of your comments.
> Before addressing the others, we need to decide how we should treat kwargs, since it has a potentially big impact on the implementation (see my comment above).

Please repeat the comment.
I don't know which comment you are thinking about.
Also, at this point we can have another chat session.

omry:
looks like there are some conflicts, can you rebase and fix?

guillaumegenthial:
I was referring to your comment

> I think now that we support instantiating dictconfigs, we should treat passthrough dictconfigs as passthrough objects and not as configs.
I guess we can make an exception based on the presence of target but I am not sure it's worth the extra complexity.

My answer was 

> This is the main friction point.
I'd argue that this would be confusing to treat dataclass instances in overrides as objects and not configs.
> 1. Instantiate should operate on config-like objects, why would the overrides already have some instantiated objects?
> 2. DataClass become sources of confusion. If config is `{"obj": {"x": 1, "y": 2}}` and kwargs is `obj=SomeDataClass(x=2)`, what do we expect? `{"obj": {"x": 2, "y": 2}}` (current behavior), or `{"obj": SomeDataClass(x=2)`
> Also, (not necessarily a good argument), it would make the implementation more complex. 

> The current implementation does
> 1. merge kwargs fully into config
> 2. forget about kwargs, only look at config

> If we want to treat kwargs as objects, we can't do that because dataclasses instances merged into OmegaConf containers would be converted back to OmegaConf containers. This means that we should keep the kwargs alonside the config during the whole instantiation path.

And I agree, discussing live on the chat would be more efficient

omry:
@guillaumegenthial,
I would like to get some progress on this, would you mind jumping to the chat at some point so we can discuss the remaining issues?

booski360 (approved):
commit

omry:
Taking over. Thanks for working on this.

PR #1244, opened 2020-12-25
[configen] Adds cwd to PYTHONPATH

omry:
1. Add cwd to PYTHONPATH, making it easier to generate code for Python files in the current working directory even if they are not installed.
2. Fixed the generated sample config.


omry:
cc @romesco, @jieru-hu, @snisarg

Issue #1191, opened 2020-12-10
[Bug] Ray AWS plugin uses deprecated flags


samuelstanton:
# 🐛 Bug
## Description
The Ray AWS launcher plugin says it requires ray>=1.0, but the plugin seems to have been written for ray<=0.8.6. In particular, the plugin uses the `--docker` flag [here](https://github.com/facebookresearch/hydra/blob/1.0_branch/plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py#L132) when calling `ray exec`. The `--docker` flag seems to have been deprecated in favor of `--run-env` in [v0.8.7](https://docs.ray.io/en/releases-0.8.7/package-ref.html).

## Checklist
- [x] I checked on the latest version of Hydra
- [x] I created a minimal repro

### The config
```
hydra:
  launcher:
    _target_: hydra_plugins.hydra_ray_launcher.ray_aws_launcher.RayAWSLauncher
    mandatory_install:
      hydra_version: 1.0.3
      ray_version: 1.0.1
      cloudpickle_version: 1.6.0
      omegaconf_version: 2.0.3
      pickle5_version: 0.0.11
      hydra_ray_launcher_version: 0.1.1
      install_commands:
        - pip3 install omegaconf==${hydra.launcher.mandatory_install.omegaconf_version}
        - pip3 install hydra-core==${hydra.launcher.mandatory_install.hydra_version}
        - pip3 install hydra-ray-launcher==${hydra.launcher.mandatory_install.hydra_ray_launcher_version}
        - pip3 install ray==${hydra.launcher.mandatory_install.ray_version}
        - pip3 install cloudpickle==${hydra.launcher.mandatory_install.cloudpickle_version}
        - pip3 install pickle5==${hydra.launcher.mandatory_install.pickle5_version}
    ray_init_cfg: {}
    ray_remote_cfg: {}
    ray_cluster_cfg:
      cluster_name: hydra-ray-cluster
      min_workers: 0
      max_workers: 1
      initial_workers: 0
      autoscaling_mode: default
      target_utilization_fraction: 0.8
      idle_timeout_minutes: 5
      docker:
        image: 'rayproject/ray-ml:latest-gpu'
        container_name: 'hydra-ray-container'
        pull_before_run: true
        run_options: []
      provider:
        type: aws
        region: us-east-2
        availability_zone: us-east-2a,us-east-2b
        cache_stopped_nodes: false
        key_pair:
[...]

PR #1249, opened 2020-12-30
Fix for docker support in Ray AWS Launcher

samuelstanton:
## Motivation

This PR fixes [this issue](https://github.com/facebookresearch/hydra/issues/1191) , and patches a small bug that crashes Docker containers launched with the plugin. Some redundant hidden behavior has been removed from `EnvSetupConf`.

Yes

## Test Plan

All the core tests pass, and the same number of plugin tests fail as on the main branch (22 passed, 20 errors).
Testing containerized jobs managed with Ray on EC2 is pretty tricky, I'm not sure how to design a production-level test for something like that. You'd need an AWS account and something like DockerHub serving an image Ray could download.

## Related Issues and PRs

https://github.com/facebookresearch/hydra/issues/1191


news/1191.plugin
```diff
@@ -0,0 +1,3 @@
+patch the Ray Plugin. Changes `--docker` to `--run-env=docker` in `ray exec` calls.
+Fixes `_remote_invoke.py` file path bug to support Docker.
+Removes redundant hidden behavior in `EnvSetupConf`.
```
omry, line 3: 1. news fragments should almost always be a single line. From the user's pov, just say that you fixed docker support in the RayLauncher
2. move the news fragment to the news section in the plugin dir.

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py
```diff
@@ -84,7 +84,7 @@ def ray_rsync_down(yaml_path: str, remote_dir: str, local_dir: str) -> None:
 def ray_tmp_dir(yaml_path: str, docker: bool) -> Generator[Any, None, None]:
     args = ["ray", "exec"]
     if docker:
-        args.append("--docker")
+        args.append("--run-env=docker")
```
omry, line 87: what is the difference?
Is this the deprecation thing you were reporting?

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/conf/__init__.py
```diff
@@ -85,25 +80,8 @@ class RsyncConf:
...
-            ).version,
-        }
-    )
-
-    commands: List[str] = field(
-        default_factory=lambda: [
-            "conda create -n hydra_${python_version:micro} python=${python_version:micro} -y",
-            "echo 'export PATH=\"$HOME/anaconda3/envs/hydra_${python_version:micro}/bin:$PATH\"' >> ~/.bashrc",
-        ]
-    )
```
omry, line 106: Can you explain this change?

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py
samuelstanton: correct. As of Ray v0.8.7 `--docker` is not the correct flag for `ray exec`

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/conf/__init__.py
samuelstanton: If a user wants this behavior, they can put it in the config file. If they don't want it, they have to go in and modify the plugin source. Why should the plugin be installing packages and modifying my `.bashrc` at runtime?

samuelstanton: Unless I'm mistaken, it's completely redundant with the [example config](https://hydra.cc/docs/plugins/ray_launcher) in the documentation.  As a user, it was very strange to remove the corresponding lines in the config and still have the launcher reinstalling packages.

omry: Commenting below, but @jieru-hu mayhave more context.

Installing things: The plugin is preparing your remove environment and ensuring it matches your local environment in terms of versions (having a version mismatch between the environment that launches the jobs and the environment that runs them is a great way to have the most fascinating bugs.
Also, even though you may choose to bake all the dependencies into your AMI or docker image, some users will not want to do that.

The change of the .bashrc indeed seems weird, I don't think we should be doing it.

You never have to change the plugin source to override configs. I suggest you go over the Structured Configs tutorial.

samuelstanton: Unless there's something I'm missing, if I have this in my config, that _should_ override the default EnvConfig, but that doesn't seem to be what actually happens. 
```
env_setup:
  pip_packages: {}
  commands: []
```

I think [this line](https://github.com/facebookresearch/hydra/blob/9e1039941ee2b714d043e15145fc358f4ab6e9b8/plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py#L63) is to blame. Maybe that is what should be changed?

omry: Generally speaking, dictionaries are being merged. you can't delete things from dicts using merge.
Well, the pip packages there are required. If you really want to delete them,I guess we could support an explicit assignment of None:
```
pip_packages:
  hydra_core: null
  omegaconf: null
  ...
```
Supporting it would take a minor change to the code that generates the setup commands (filter anything that is None).

Can you explain why you don't think you need those packages in your remote environment?

samuelstanton: I have no doubt they're required -- but if you're running a  containerized app there's really no reason for them to be installed at runtime. I understand the versioning rationale... version discrepancy bugs suck. If you think keeping them there makes for a better experience for the average user I'm def not interested in dying on this hill. To me it seemed very strange that if I knew what I was doing I couldn't override the behavior without literally digging through the code and commenting the offending lines out.

samuelstanton: actually I just remembered why I did this. To test my changes I was installing the plugin from source in the container. I wanted to be sure it wasn't reinstalling from pypi.

samuelstanton: It's also possible (though maybe unlikely) that the user doesn't have pip in their container (maybe pip3, for example). In that case the head node setup fails.

omry:
Somehow our discussion about the config changes disappeared once you reverted.
I still think we should avoid changing bashrc.

Side note: can you rebase on top of master? there seems to be merge conflicts.

samuelstanton:
Can you see the discussion again?

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/conf/__init__.py
samuelstanton: @omry So are we agreed that I make `EnvSetupConf.commands` an empty ListConfig and leave `EnvSetupConf.pip_packages` as is?

omry: In CI, the plugin is testing the proper code and not the pip package (in fact, our Amazon image does not even have access to the Internet for security reasons).
I am fine with adding support for removing mandatory pip packages by overriding them to null (which is the None in yaml).
Essentially, it should be as easy filtering any package with a None version in the code.

You should be able to override the commands to an empty list from your side. What I said about dictionaries does not apply to lists (Lists are always replacing the source list).

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py
```diff
@@ -107,7 +107,7 @@ def ray_new_dir(yaml_path: str, new_dir: str, docker: bool) -> None:
     """
     args = ["ray", "exec"]
     if docker:
-        args += "--docker"
+        args += "--run-env=docker"
```
jieru-hu, line 110: args.append() instead :)

jieru-hu:
Thanks for the PR!

1. could you add a docker example application [here](https://github.com/facebookresearch/hydra/tree/master/plugins/hydra_ray_launcher/examples)?
2.  ray run_env takes in 3 arguments ` --run-env [auto|host|docker]`, I think we should support all there arguments instead of hard coding docker. You can add an enum [here](https://github.com/facebookresearch/hydra/blob/master/plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/conf/__init__.py), something like
```
class RayRunEnv(Enum):
    auto = "auto"
    host = "host"
    docker = "docker"
```

and add an extra node `run_env` in hydra.launcher.ray
```
@dataclass
class RayAWSConf(RayConf):
    cluster: RayClusterConf = RayClusterConf()
    run_env: RayRunEnv = RayRunEnv.docker
```

jieru-hu:
> sorry for all the small commits. Is there a way for me to run the CI tests without pushing to this PR? The tests don't seem to pass on master when I run them locally.

Could you provide more info on what issues are you seeing? (you will need to set all the AWS variables accordingly of course.)

jieru-hu:
Also feel free to join our chat https://hydra-framework.zulipchat.com/login/, might be easier to chat there :)

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
```diff
@@ -147,7 +151,7 @@ def launch_jobs(
 
             ray_exec(
                 launcher.ray_yaml_path,
-                launcher.docker_enabled,
+                launcher.ray_cfg.run_env,
```
omry: no type error here? ray_cfg.run_env is the Enum, but the function wants a strings.

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
```diff
@@ -59,11 +59,15 @@ def launch(
     assert launcher.task_function is not None
 
     setup_commands = launcher.env_setup.commands
+    package_filter = filter(
```
omry, line 62: I think packages or package_list are better names.
```suggestion
    packages  = filter(
```

omry (changes requested):
marking as request changes to clear my queue.

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/conf/__init__.py
samuelstanton: It looks like if you don't prepend the conda path to `PATH` in `.bashrc` then the head node setup fails because the `PyYAML` install fails
```
Collecting PyYAML>=5.1.*
  Downloading https://files.pythonhosted.org/packages/64/c2/b80047c7ac2478f9501676c988a5411ed5572f35d1beff9cae07d321512c/PyYAML-5.3.1.tar.gz (269kB)
     |████████████████████████████████| 276kB 7.8MB/s
Building wheels for collected packages: PyYAML
  Building wheel for PyYAML (setup.py) ... done
  Created wheel for PyYAML: filename=PyYAML-5.3.1-cp36-cp36m-linux_x86_64.whl size=44621 sha256=3834687cdd8b382c59c8a02769f50093619ad10ecc392500ff6fbb274524eed2
  Stored in directory: /home/ubuntu/.cache/pip/wheels/a7/c1/ea/cf5bd31012e735dc1dfea3131a2d5eae7978b251083d6247bd
Successfully built PyYAML
Installing collected packages: typing-extensions, dataclasses, PyYAML, omegaconf
  Found existing installation: PyYAML 3.12
ERROR: Cannot uninstall 'PyYAML'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
```

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
```diff
@@ -59,12 +59,12 @@ def launch(
     assert launcher.task_function is not None
 
     setup_commands = launcher.env_setup.commands
+    packages = filter(
+        lambda x: x[1] is not None, launcher.env_setup.pip_packages.items()
+    )
```
jieru-hu, line 64: why would this be `None`?

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
```diff
@@ -115,19 +115,20 @@ def launch(
...
         ) as remote_tmp_dir:
 
             ray_rsync_up(
                 launcher.ray_yaml_path, os.path.join(local_tmp_dir, ""), remote_tmp_dir
             )
 
             script_path = os.path.join(os.path.dirname(__file__), "_remote_invoke.py")
-            ray_rsync_up(launcher.ray_yaml_path, script_path, remote_tmp_dir)
+            remote_script_path = os.path.join(remote_tmp_dir, "_remote_invoke.py")
+            ray_rsync_up(launcher.ray_yaml_path, script_path, remote_script_path)
```
jieru-hu, line 131: why this change? this only rsync up one certain file, there's other stuff in the dir that needs to be rsyned if I remember it correctly..

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py
```diff
@@ -121,15 +116,16 @@ def ray_down(yaml_path: str) -> None:
     _run_command(["ray", "down", "-y", yaml_path])
 
 
-def ray_up(yaml_path: str) -> None:
-    _run_command(["ray", "up", "-y", yaml_path])
+def ray_up(yaml_path: str, no_config_cache: bool) -> None:
+    args = ["ray", "up", "-y", yaml_path]
+    if no_config_cache:
+        args.append("--no-config-cache")
```
jieru-hu, line 122: do we always need `--no-config-cache`? if not we should expose this in the launcher config?

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
omry: This is a new feature:
it's a way to remove an existing a package from the packages dict.

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py
omry: Also, what is --no-config-cache for?

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
samuelstanton: that change was made to fix a FileNotFoundError I was getting in my dockerized app.

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py
samuelstanton: I added the option because the cached AWS config from running the integration tests was interfering with running jobs with my normal AWS profile. We probably could just always pass the flag, I just assumed we'd want to make it an option instead of hardcoding it

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
jieru-hu: I'm not sure this is a proper fix. What was the exact exception in the docker app?

samuelstanton: It happened during the `ray exec` call. `<remote_tmp_dir/_remote_invoke.py` no such file or directory

samuelstanton: I don't think this line is the problem. If you look at the circleCI output, the node can find `_remote_invoke.py` just fine and all this line did before was copy that specific script

```
[2021-01-07 00:55:26,487][HYDRA] Running command: ['ray', 'exec', '--run-env=auto', '/tmp/tmpb7uftols.yaml', 'python /tmp/tmp.AjwSf1fZyK/_remote_invoke.py /tmp/tmp.AjwSf1fZyK']
[2021-01-07 00:55:34,167][HYDRA] Output: 2021-01-07 00:55:28,634	INFO services.py:1090 -- View the Ray dashboard at http://127.0.0.1:8266
  File "/tmp/tmp.AjwSf1fZyK/_remote_invoke.py", line 87, in <module>
    launch_jobs(sys.argv[1])
  File "/tmp/tmp.AjwSf1fZyK/_remote_invoke.py", line 62, in launch_jobs
    result = [ray.get(run) for run in runs]
  File "/tmp/tmp.AjwSf1fZyK/_remote_invoke.py", line 62, in <listcomp>
    result = [ray.get(run) for run in runs]
  File "/home/ubuntu/anaconda3/envs/hydra_3.8.5/lib/python3.8/site-packages/ray/worker.py", line 1452, in get
    raise value.as_instanceof_cause()
ray.exceptions.RayTaskError(FileNotFoundError): ray::_run_job() (pid=6899, ip=172.31.35.143)
  File "python/ray/_raylet.pyx", line 482, in ray._raylet.execute_task
  File "/home/ubuntu/anaconda3/envs/hydra_3.8.5/lib/python3.8/site-packages/hydra_plugins/hydra_ray_launcher/_launcher_util.py", line 40, in _run_job
    return run_job(
  File "/home/ubuntu/anaconda3/envs/hydra_3.8.5/lib/python3.8/site-packages/hydra/core/utils.py", line 126, in run_job
  File "/tmp/hydra_test/no_config.py", line 9, in experiment
  ```

Comment (samuelstanton) [reply plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py:131]: this error is being through _inside_ the `_remote_invoke.py` script, looking for something in an entirely different directory.

**Comment by jieru-hu** [comment]
latest merge of master seems to have messed up the PR's `files changed`. Could you fix that?

feel free to force push this PR as I've mentioned in our chats :)

**Comment by jieru-hu** [comment]
The `run_env=False` logging we see in CI is due to [this](https://github.com/facebookresearch/hydra/blob/master/plugins/hydra_ray_launcher/tests/test_ray_aws_launcher.py#L255-L256)
Since this PR changes the interface of several commonly used methods, please make sure you updated the code throughout the plugin code accordingly.

File: plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py (hunk 9 above)
Comment (jieru-hu) [reply plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py:122]: Could you provide a repro or stacktrace for this? 
Ray's documentation doesn't have much info on this option, I'd like to understand a bit more here :)

Comment (jieru-hu) [reply plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py:122]: also if we are exposing the `--no-config-cache` option through the config, we need to think about should we only apply it in `ray up`

since this option can be used in combo with other major command like `exec`, `submit` etc
https://docs.ray.io/en/releases-1.1.0/package-ref.html#cmdoption-ray-up-no-config-cache

Comment (jieru-hu) [reply plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py:122]: >I added the option because the cached AWS config from running the integration tests was interfering with running jobs with my normal AWS profile. We probably could just always pass the flag, I just assumed we'd want to make it an option instead of hardcoding it


I'm not sure I understand the issue here. Could you be more specific?

**Comment by jieru-hu** [comment]
Thanks for updating the PR!

1. Please add an example application for docker.
2. Please update the ray launcher plugin doc as well.
3. I'm not sure if `--no-config-cache` should be needed or added to the config (with reasons mentioned in previous comments.). Could you address the comments?

**Comment by samuelstanton** [comment]
> Thanks for updating the PR!
> 
> 1. Please add an example application for docker.
> 2. Please update the ray launcher plugin doc as well.
> 3. I'm not sure if `--no-config-cache` should be needed or added to the config (with reasons mentioned in previous comments.). Could you address the comments?

Prioritizing some other work right now, but will get to these at some point.

File: requirements/requirements.txt [hunk 10]
```diff
@@ -1,3 +1,3 @@
-omegaconf>=2.1.0dev13
+omegaconf==2.1.0dev13
```
omry, line 1: Please undo and rebase on master.
correct version is ==2.1.0dev16.

omry (changes requested):

requirements/requirements.txt
samuelstanton: The change to the omegaconf dependency is breaking. Pinning the version temporarily so that my code still runs. Here's the error you get with the latest version of `omegaconf`

```
[2021-01-22 17:15:15,075][HYDRA] Running command: ray exec --run-env=auto /tmp/tmpxbaa8z1i.yaml python /tmp/tmp.oza1lASz0e/_remote_invoke.py /tmp/tmp.oza1lASz0e
[2021-01-22 17:15:18,224][HYDRA] Output: Traceback (most recent call last):
  File "/opt/venv/lib/python3.8/site-packages/omegaconf/omegaconf.py", line 793, in open_dict
    OmegaConf.set_struct(config, False)
  File "/opt/venv/lib/python3.8/site-packages/omegaconf/omegaconf.py", line 483, in set_struct
    conf._set_flag("struct", value)
  File "/opt/venv/lib/python3.8/site-packages/omegaconf/base.py", line 102, in _set_flag
    self._invalidate_flags_cache()
  File "/opt/venv/lib/python3.8/site-packages/omegaconf/base.py", line 512, in _invalidate_flags_cache
    if self.__dict__["_flags_cache"] is not None:
KeyError: '_flags_cache'
```

jieru-hu: I wonder if this is caused by a version mismatch on the AWS instance vs your laptop; probably a good idea to double check to make sure the lib versions are consistent.

omry: Yes. is sounds like you have the wrong version of OmegaConf on the remote side.

samuelstanton: I thought of that. Made sure I was installing the same version in the docker container.

omry:
By the way, I released `hydra-1.1.0dev3` half an hour ago.

requirements/requirements.txt
samuelstanton: Once I clear an impending deadline I'll have time to change this back and add the other things that have been requested. But at the moment I need things that were working to keep working.

omry: no worries.

samuelstanton: actually, let me just make this change on a private branch instead of the PR

omry: That would be even better :)

samuelstanton:
~~1. Please add an example application for docker.~~
~~2. Please update the ray launcher plugin doc as well.~~
> 3. I'm not sure if `--no-config-cache` should be needed or added to the config (with reasons mentioned in previous comments.). Could you address the comments?

I haven't reproduced the behavior that prompted the addition of the `--no-config-cache` option. I'll spend a little more time on it, but if it turns out to difficult to recreate I suppose there isn't much harm in removing the option. It could always be added back in a later PR if the problem arises again.

samuelstanton:
I went ahead and removed the `no_config_cache` option. Are there any other requested changes?

plugins/hydra_ray_launcher/examples/docker/config.yaml
```diff
@@ -0,0 +1,48 @@
+task: 1
```
jieru-hu, line 1: Thank you for adding an example!

Could we only add configs that are actually needed for the example to work?

Some of the configs here are already in teh default config so shouldnt be needed to be put here again

plugins/hydra_ray_launcher/examples/docker/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+      cluster:
+        cluster_name: example-cluster
+        min_workers: 0
+        max_workers: 0
+        initial_workers: 0
+        autoscaling_mode: default
+        target_utilization_fraction: 0.8
+        idle_timeout_minutes: 5
+        docker:
+          image: 'samuelstanton/ray-plugin-example'
```
jieru-hu, line 25: could we use a public docker image, maybe one from the ray team's examples?

plugins/hydra_ray_launcher/examples/docker/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+      commands: []
+    ray:
+      cluster:
+        cluster_name: example-cluster
+        min_workers: 0
+        max_workers: 0
+        initial_workers: 0
+        autoscaling_mode: default
+        target_utilization_fraction: 0.8
+        idle_timeout_minutes: 5
```
jieru-hu, line 23: do these actually need to be added here?

plugins/hydra_ray_launcher/examples/docker/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+          container_name: 'hydra-container'
+          pull_before_run: true
+          run_options: []
+        provider:
+          type: aws
+          region: us-east-2
+          availability_zone: us-east-2a,us-east-2b
+          cache_stopped_nodes: false
+          key_pair:
+            key_name: hydra
```
jieru-hu, line 35: These are in default config, I don't think we need to add them here.

plugins/hydra_ray_launcher/examples/docker/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+        auth:
+          ssh_user: ubuntu
+        head_node:
+          IamInstanceProfile:
+            Arn: arn:aws:iam::<AWS_ACCOUNT_ID>:instance-profile/ray-head-v1
+          InstanceType: m4.large
```
jieru-hu, line 41: we do not specify these in our other examples, it should just work. Is this required for the docker example to work?

plugins/hydra_ray_launcher/examples/docker/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+          ImageId: ami-010bc10395b6826fb
+        worker_nodes:
+          IamInstanceProfile:
+            Arn: arn:aws:iam::<AWS_ACCOUNT_ID>:instance-profile/ray-worker-v1
+          InstanceType: m4.large
+          ImageId: ami-010bc10395b6826fb
```
jieru-hu, line 47: Can we use a public AMI image? If not, is it possible to make this one public?

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_config.py
```diff
@@ -112,6 +112,12 @@ class EnvSetupConf:
     )
 
 
+class RayRunEnv(Enum):
+    auto = "auto"
```
jieru-hu, line 116: Could you add a link to Ray's public doc here?

website/docs/plugins/ray_launcher.md
```diff
@@ -190,6 +190,13 @@ Loaded cached provider configuration
 ```
 </details>
 
+<details><summary>Dockerized app</summary>
+
+```commandline
+python docker_app.py --multirun hydra/launcher=ray_aws task=1,2,3
+```
```
jieru-hu, line 197: Since this is collapsed, it doesnt hurt to add more output related to running on a docker here :)

jieru-hu (changes requested):

plugins/hydra_ray_launcher/examples/docker/config.yaml
samuelstanton: this image is public :). I don't think you can use any old image because you need hydra and its dependencies installed. 

https://hub.docker.com/repository/docker/samuelstanton/ray-plugin-example

plugins/hydra_ray_launcher/examples/docker/config.yaml
samuelstanton: This is the `ubuntu-18-04-docker` community AMI

plugins/hydra_ray_launcher/examples/docker/config.yaml
samuelstanton: There's a balance between keeping the example simple by not including redundant elements of the config, and making the example easy for the user to adapt. It is very likely that some users will not want to use the `us-east-2` region, and they shouldn't have to dig through the source to figure out how to change it.

plugins/hydra_ray_launcher/examples/docker/config.yaml
samuelstanton: these are important options (IMO). It makes sense to use the example to show the user where/how to configure them. Users that don't care will be copy/pasting the config in any case.

plugins/hydra_ray_launcher/examples/docker/config.yaml
samuelstanton: generally speaking I've suppressed redundant config fields that won't usually need to be changed. There are some fields that are worth keeping IMO because it allows the user to easily adapt the example to their requirements.

plugins/hydra_ray_launcher/examples/docker/config.yaml
samuelstanton: I guess it's not required. I'll leave InstanceType for didactic purposes but remove the other field

samuelstanton:
I think it's worth pointing out that the PR as it currently exists passes the unit tests and fixes a known bug in master that completely prevents the dockerized use case. The PR has already received a lot of time and attention relative to its size. I'm all for making sure everything is up to spec, but it's already an improvement over the status quo in its current form.

omry:
@samuelstanton:

> I think it's worth pointing out that the PR as it currently exists passes the unit tests and fixes a known bug in master that completely prevents the dockerized use case. The PR has already received a lot of time and attention relative to its size. I'm all for making sure everything is up to spec, but it's already an improvement over the status quo in its current form.

This is definitely an improvement, but as we will need to maintain it moving forward I want to make sure we do not create technical debt even if it means this PR will be delayed.
Some of the changes you are introducing are establishing a different norm for users on how to do things and we need to make sure we agree with those things.

Other changes like using a personal docker image in the example are a problem because:
1. We do not control it so we will not be able to maintain it.
2. You are requiring Hydra to be preinstalled on your image, which is in conflict with our existing requirements from the AMI. 

These changes deserve discussion. I totally appreciate you putting the time into this but if you want to to accept the changes we need to be onboard with every change.

I know @jieru-hu asked for the example. An alternative is for you to send the PR without the example and let us add the example later.

I hope you understand.

samuelstanton:
Makes total sense. Since I expect to use this feature in the future, I want to make sure it is as easy to maintain as possible as well. 

> 1. We do not control it so we will not be able to maintain it.

If you're concerned about the docker image being hosted on my personal Dockerhub account, I've included the Dockerfile in `examples/docker`. If there is an official Facebook or Hydra DockerHub account, the image could be hosted there. It should be a very simple fix.

> 2. You are requiring Hydra to be preinstalled on your image, which is in conflict with our existing requirements from the AMI.

Frankly I find this concern a little confusing. The AMI and the docker image are separate entities. The AMI simply determines the base environment in which the docker container will run. There is no need for a Hydra installation at all in the base environment in the dockerized setting.

In my example, I provided an installation of Hydra in the docker  image, but it could always be overriden by another installation at runtime. My guess is this concern probably gets back to our previous discussion about the merits of installing Hydra specifically at runtime to avoid version discrepancies. If installing Hydra at runtime is a design pattern you want to strictly enforce in every example, we can simplify the example by removing the Hydra installation in the Dockerfile. 

If these are the only real concerns with the example at this point, I think we can work them out. If there are other norms that the example violates that we haven't yet discussed, maybe it would be better for you guys to come up with your own. Let me know how you would like to proceed.

jieru-hu:
Hi @samuelstanton 
Thanks for you understanding and taking the time to add the example. 

In the docker example we are making some assumptions about the user (for example users are installing Hydra from source in their local machine as well and they are on python 3.8). Because of these assumptions, it's likely the example will not just work like what we try to achieve with the other examples available for Ray launcher. 

I think as this point it's probably better for us to add our own docker example. I think the rest of the code looks OK. Could you update your PR without the example and the doc change? thanks.

samuelstanton:
sure thing

samuelstanton:
Not sure what's up with the Windows tests. Tried rebasing off master but no luck.

omry:
don't worry about it, we will take a look and let you know if you need to do anything.
sometimes CI can be a bit flaky, especially on Windows :).

samuelstanton:
Looks like all checks are passing :). Any chance we can get this merged?

website/docs/plugins/ray_launcher.md
```diff
@@ -190,6 +190,7 @@ Loaded cached provider configuration
 ```
 </details>
 
+
```
jieru-hu, line 193: please remove this file from the change list

jieru-hu:
sorry for the delay - could you rebase from master to run the latest tests again? Otherwise the PR looks good, we  can  merge after all tests pass. thanks.

PR #1251, opened 2021-01-01
[configen] Add optional default flags dict per module.

romesco:
## Motivation

See https://github.com/pytorch/hydra-torch/pull/56 for context.

For configs from pytorch, it is common for `ListConfigs` to need conversion into `list` before instantiating objects. This motivates including the `_convert_` field in the autogenerated configs themselves to instruct hydra on how to handle `ListConfigs` during instantiation. We implement this field in `configen.yaml` on a _per module_ basis.

Yes

## Test Plan

Tests ensuring proper generation when `_convert_` is included and when it is omitted (how things are currently) to come.


tools/configen/configen/config.py
```diff
@@ -10,6 +10,7 @@
 class ModuleConf:
     name: str = MISSING
     classes: List[str] = MISSING
+    _convert_: Optional[str] = None
```
omry, line 13: Let's add a default_flags dict:
Let's add a defaults_flag dict instead. this also support recursive.
```suggestion
class ModuleConf:
    name: str = MISSING
    classes: List[str] = MISSING
    default_flags: Dict[str, Any] = field(default_factory=dict)
```

tools/configen/configen/templates/module.j2
```diff
@@ -10,7 +10,10 @@ from dataclasses import dataclass, field
 @dataclass
 class {{class_name}}Conf:
     _target_: str = "{{ classes_map[class_name].target }}"
+{% if classes_map[class_name].convert %}
+    _convert_: 'all'
+{% endif %}
```
omry, line 15: Any value in the default_flags dict  is copied:
loop on the default_flags dict and and add things.
key should be copied. value should be copied, and don't forget the type.

result should be.
```python
_convert_ : str = "all"
```

It should work correctly for other values of _convert_ (like "none" or "partial").

omry (changes requested):
1. We should support all flags and it should be easy to add additional flags).
2. the value of the flags from the config should be used.

tools/configen/configen/config.py
```diff
@@ -10,7 +10,7 @@
 class ModuleConf:
     name: str = MISSING
     classes: List[str] = MISSING
-    _convert_: Optional[str] = None
+    default_flags: Dict[str, Any] = field(default_factory=dict)
```
omry, line 13: Obviously this change would have to go with some other changes.

romesco: Yeah I'm looking into it now.

romesco: done

tools/configen/configen/templates/module.j2
```diff
@@ -10,9 +10,9 @@ from dataclasses import dataclass, field
 @dataclass
 class {{class_name}}Conf:
     _target_: str = "{{ classes_map[class_name].target }}"
-{% if classes_map[class_name].convert %}
-    _convert_: 'all'
-{% endif %}
+{% for flag in classes_map[class_name].default_flags %}
+    {{ flag.name }}: {{ flag.type_str }} = "{{ flag.default }}"
```
omry, line 14: flags can be boolean as well. (like _recursive_).
I think this should be implemented by adding the flags to the parameters like other parameters.

omry (commented):
it will be easier to code this if you start by creating a few tests that are testing the various new scenarios.

tools/configen/conf/configen.yaml
```diff
@@ -15,7 +15,10 @@ configen:
   # list of modules to generate configs for
   modules:
     - name: configen.samples.my_module
-      _convert_: 'all'
+      # Optional dict defining hydra flags for all classes in this module
+      default_flags:
+        _convert_ : 'all'
+        _recursive_ : True
```
omry, line 21: Let's not change the sample config output. Can you document it in a comment?

tools/configen/configen/configen.py
```diff
@@ -138,13 +133,23 @@ def generate_module(cfg: ConfigenConf, module: ModuleConf) -> str:
...
+    flag_type_dict = {
+        "_convert_": "str",
+        "_recursive_": "bool",
+    }
+    def_flags: List[Parameter] = []
     for name, default in module.default_flags.items():
+        try:
+            flag_type = flag_type_dict[name]
+        except RuntimeError:
+            print(f"{name} flag is currently unsupported by configen.")
```
omry, line 146:
```
log.error(f"{name} is not a supported flag")
```
but actually, just raise an exception. no need to continue working if the user config is bad.

tools/configen/configen/configen.py
```diff
@@ -138,13 +133,23 @@ def generate_module(cfg: ConfigenConf, module: ModuleConf) -> str:
...
+
+    flag_type_dict = {
+        "_convert_": "str",
+        "_recursive_": "bool",
+    }
+    def_flags: List[Parameter] = []
     for name, default in module.default_flags.items():
+        try:
+            flag_type = flag_type_dict[name]
+        except RuntimeError:
```
omry, line 145: what is this RuntimeError? can you be more specific? KeyError?

tools/configen/configen/configen.py
```diff
@@ -132,11 +133,33 @@ def generate_module(cfg: ConfigenConf, module: ModuleConf) -> str:
     classes_map: Dict[str, ClassInfo] = {}
     imports = set()
     string_imports: Set[str] = set()
+
```
omry, line 136: can you extract this logic into a function?

tools/configen/tests/test_generate.py
```diff
@@ -100,9 +100,38 @@ def test_generated_code() -> None:
...
+    if generated != expected:
+        print(diff)
+        assert False, f"Mismatch between {expected_file} and generated code"
+
+
 @pytest.mark.parametrize(
     "classname, params, args, kwargs, expected",
     [
+        pytest.param("Empty", {}, [], {}, Empty(), id="Empty"),
         pytest.param("Empty", {}, [], {}, Empty(), id="Empty"),
```
omry, line 135: This looks exactly like the following line.

tools/configen/tests/test_generate.py
```diff
@@ -100,9 +100,38 @@ def test_generated_code() -> None:
         assert False, f"Mismatch between {expected_file} and generated code"
 
 
+def test_generated_code_with_default_flags() -> None:
```
omry, line 103: Okay, after seeing the corresponding generated code, I think this is too much.
We don't need to re-test everything from scratch. this strategy would lead to an exponential explosion in the number of tests.

Instead, test the new logic directly.
add a few generated class cases, one with convert, one with recursive and one with both.

omry (changes requested):

tools/configen/conf/configen.yaml
romesco: yes

tools/configen/tests/test_generate.py
romesco: I'm not sure why this made it in there. Removing.

tools/configen/configen/configen.py
romesco: which logic?

omry: The 20 or so new lines in this loop that are handling the flags.

tools/configen/tests/test_generate.py
romesco: done

tools/configen/configen/config.py
```diff
@@ -10,6 +10,7 @@
 class ModuleConf:
     name: str = MISSING
     classes: List[str] = MISSING
+    default_flags: Optional[Dict[str, Any]] = field(default_factory=dict)
```
omry, line 13: Here is a thought, instead of using a dict and then restricting manually at runtime to supported types, we can just use a dataclass directly.
dunno what I was thinking.
```python
@dataclass
class Flags:
  _recursive_ : Optional[bool] =  None
  _convert_ : Optional[str] =  None
```

For convert, you should actually use the ConvertMode Enum from Hydra directly here.

```python
  _convert_ : Optional[ConvertMode] =  None
```

tools/configen/conf/configen.yaml
```diff
@@ -15,6 +15,11 @@ configen:
   # list of modules to generate configs for
   modules:
     - name: configen.samples.my_module
+      # Optional dict defining hydra flags for all classes in this module:
+      # default_flags:
+      #   _convert_ : 'all'
+      #   _recursive_ : True
```
omry, line 21:
```suggestion
      #   _convert_: all
      #   _recursive_: True
```

Since this is yaml, you do not need the quotes around strings.

tools/configen/configen/configen.py
```diff
@@ -128,15 +128,45 @@ def is_incompatible(type_: Type[Any]) -> bool:
     return True
 
 
+def get_default_flags(module: ModuleConf) -> List[Parameter]:
```
omry, line 131: The suggestion to use a dataclass will simplify this function quite a bit.

tools/configen/configen/configen.py
```diff
@@ -128,15 +128,45 @@ def is_incompatible(type_: Type[Any]) -> bool:
...
+    return def_flags
+
+
 def generate_module(cfg: ConfigenConf, module: ModuleConf) -> str:
     classes_map: Dict[str, ClassInfo] = {}
     imports = set()
     string_imports: Set[str] = set()
+
+    default_flags = get_default_flags(module)
+
```
omry, line 163: So much cleaner thank trying to force it into an already too large function, isn't it? :)

omry (commented):
Starting to look good, one more suggestion for consideration.

romesco:
Made the requested changes. Definitely seems much closer to the _right_ way.

tools/configen/configen/config.py
```diff
@@ -1,16 +1,23 @@
...
+    _convert_: Optional[ConvertMode] = None
+    _recursive_: Optional[bool] = None
+
+
 @dataclass
 class ModuleConf:
     name: str = MISSING
     classes: List[str] = MISSING
-    default_flags: Optional[Dict[str, Any]] = field(default_factory=dict)
+    default_flags: Optional[Flags] = None
```
omry, line 20: I think you can go with 
```
default_flags: Flags = Flags()
```

and it will populate with the default values (None) for each of the flags.

tools/configen/configen/configen.py
```diff
@@ -128,28 +129,23 @@ def is_incompatible(type_: Type[Any]) -> bool:
...
-            Parameter(
-                name=name,
-                type_str=flag_type,
-                default=default,
+    if module.default_flags:
+        for name, default in module.default_flags.items():
+            flag_type = Flags.__annotations__[name].__args__[0].__name__
+
+            if name == "_convert_":
+                imports_set.add(ConvertMode)
```
omry: I believe the code can handle both string and ConvertMode as the values for convert.
so far, the generated dataclasses did not depend on Hydra. let's keep it this way and use a string for the _convert_ type in the generated code.

tools/configen/configen/config.py
romesco: done

tools/configen/configen/configen.py
romesco: done

tools/configen/configen/configen.py
```diff
@@ -129,16 +128,18 @@ def is_incompatible(type_: Type[Any]) -> bool:
...
         for name, default in module.default_flags.items():
             if default:
                 flag_type = Flags.__annotations__[name].__args__[0].__name__
+                if "ConvertMode" in flag_type:
+                    flag_type = "str"
+                    default = str(default).strip("ConvertMode.").lower()
 
-                if name == "_convert_":
-                    imports_set.add(ConvertMode)
+                default = f'"{default}"' if flag_type == "str" else default
```
omry: no need for all of it.
There are exactly two flags in module.default_flags.{_recursive_, _convert_}.
just access them directly. no need for a generic loop and no need to deduce the type automatically.

Also, you can use default.name (and I don't think you want to lowercase).

tools/configen/configen/config.py
```diff
@@ -17,7 +17,7 @@ class Flags:
 class ModuleConf:
     name: str = MISSING
     classes: List[str] = MISSING
-    default_flags: Optional[Flags] = None
+    default_flags: Optional[Flags] = Flags()
```
omry, line 20: I think this can be non optional (and that you still does not need to specify those in the config).

tools/configen/configen/configen.py
romesco: I thought the point was to write this generically so it can be added to in the future?

Also, it does need to be lower. It errors out if you pass 'ALL' for example.

tools/configen/configen/config.py
romesco: I think you're right. Just tested it. Is this because all the field in `Flags` are optional?

tools/configen/configen/configen.py
```diff
@@ -128,15 +128,43 @@ def is_incompatible(type_: Type[Any]) -> bool:
...
+
+                def_flags.append(
+                    Parameter(
+                        name=name,
+                        type_str=flag_type,
+                        default=default,
+                    )
+                )
+
```
omry, line 151: I was thinking something along the lines of :
```python
if module.default_flags._recursive_ is not None:
  def_flags.append(Parameter(name="_recursive_", type_str=bool, default=module.default_flags._recursive_ ))

if module.default_flags._convert_ is not None:
  def_flags.append(Parameter(name="_convert_", type_str=str, default=module.default_flags._convert_.name))
```

(I didn't run it so there may need to be some tweaks).

tools/configen/configen/config.py
omry: The final config is the merge of the  top level config dataclass and the config from the config file.
the dataclass is providing the default values.

tools/configen/configen/configen.py
omry: That was the initial idea, but once we have a dataclass describing the flags it's cleaner to just be explicit.

What error?

tools/configen/configen/configen.py
romesco: done with tweaks

tools/configen/configen/configen.py
romesco:
```
 hydra.errors.InstantiationException: Error instantiating/calling 'torchvision.transforms.transforms.Compose' : Unsupported _convert_ value: ALL
``

Comment (omry) [reply tools/configen/configen/configen.py]: Is that the full error message?
This is really weird. ConvertMode is defined as:
```
class ConvertMode(Enum):
    NONE = 0
    PARTIAL = 1
    ALL = 2
```

All upper case.
I am expecting to b e able to use the string ALL as a valid value without having to lowercase it (which should result in an error).

Comment (omry) [reply tools/configen/configen/configen.py]: Okay, this is from _pop_convert_mode().
Can you also make it case insensitive? this is some custom logic. (normally OmegaConf enums are case sensitive).
In this case the implementation actually use lowercase even though the Enum is upper case, which is an error.

Comment (romesco) [reply tools/configen/configen/configen.py]: Yeah. Thanks for tracking this down. I wasn't sure whether it was write to make up for this in configen or dive into hydras internals. Change is done.

File: tools/configen/configen/configen.py [hunk 20]
```diff
@@ -128,15 +128,44 @@ def is_incompatible(type_: Type[Any]) -> bool:
...
+def get_default_flags(module: ModuleConf) -> List[Parameter]:
+
+    def_flags: List[Parameter] = []
+
+    if module.default_flags._convert_ is not None:
+        def_flags.append(
+            Parameter(
+                name="_convert_",
+                type_str="str",
+                default=f"""\"{str(module.default_flags._convert_.name)}\"""",
```
omry, line 140: 1. name is already a string.
2. no need for a multiline string:

```suggestion
                default=f"\"{module.default_flags._convert_.name}\",
```

omry (commented):
looking good.
last thing, see comment.

tools/configen/configen/configen.py
romesco: good catch. This was left over from the previous formatting requirements.

omry (approved):
Great, thanks!

PR #1130, opened 2020-11-10
Implemented disabling logging configuration & added config file

briankosw:
## Motivation

Implemented disabling the configuration for Python logging, as described in #1021. Added `notset.yaml` configuration file for hydra/hydra_logging and hydra/job_logging.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebookresearch/hydra/blob/master/CONTRIBUTING.md)?

(Yes)/No

## Test Plan
I tested my implementation locally by comparing the root logger's state before and after `hydra.main`. I wasn't sure how to write that and add it to the test suite.

## Related Issues and PRs
N/A

PR #1263, opened 2021-01-04
Change default log level from INFO to WARNING

### LINKED ISSUE #1262: [Bug?] Hydra turns on global logging to the info level
## Description
Extending on https://github.com/facebookresearch/hydra/issues/1012, Hydra doesn't simply duplicate logs, but it also turns on logging of _all_ loggers globally to the INFO level:
https://github.com/facebookresearch/hydra/blob/bc3567da6d06e9e0e437de4d4bddf9c0a9d23516/hydra/core/utils.py#L46
This leads me to get all kinds of spam from other used packages... 

Is this actually a desired behavior? If so downstream packages will have to add:
```
 {"hydra/job_logging": "disabled"},
 {"hydra/hydra_logging": "disabled"},
```
to all defaults lists to circumvent Hydra updating the global logger, but I'm not sure if I'm losing important logging info by doing this. Am I missing something?

## Checklist

## To reproduce
Clone [Mephisto](https://github.com/facebookresearch/Mephisto).
Run an example script, like `python examples/parlai_chat_task_demo/parlai_test_script.py`
See logs like the following:
```
...
[2021-01-04 11:33:53,609][sh.command][INFO] - <Command '/bin/rm -rf /Users/jju/mephisto/data/data/runs/NO_PROJECT/1034/build/router'>: starting process
[2021-01-04 11:33:53,617][sh.command][INFO] - <Command '/bin/rm -rf /Users/jju/mephisto/data/data/runs/NO_PROJECT/1034/build/router', pid 98262>: process started
[2021-01-04 11:33:53,897][sh.command][INFO] - <Command '/bin/rm -rf /Users/jju/mephisto/data/data/runs/NO_PROJECT/1034/build/router', pid 98262>: process completed
...
```

## Expected Behavior
I wouldn't expect using Hydra would cause me to see info logs of other modules like `sh`.

## System information
- **Hydra Version** :  1.0.0
- **Python version** :  3.8.3
- **Virtual environment type and version** : conda 4.8.3
- **Operating system** : Mac OS 10.15.7

omry:
Closes #1262.

The default log level for Python apps is WARNING ([per Python docs](https://docs.python.org/3/howto/logging.html)). Hydra no longer sets the log level of the root logger to ensure it does not change this behavior.
This is a breaking change, people relying on the previous behavior can configure the log level on their own side by adding something like this to their config:

```yaml
hydra:
  job_logging:
    root:
      level: INFO
```

### DIFF
--- a/examples/configure_hydra/logging/my_app.py
+++ b/examples/configure_hydra/logging/my_app.py
@@ -10,7 +10,11 @@ log = logging.getLogger(__name__)
 
 @hydra.main(config_path="conf", config_name="config")
 def my_app(_cfg: DictConfig) -> None:
-    log.info("Info level message")
+    # By default, only logs of warning or higher level will be printed
+    log.debug("DEBUG level message")
+    log.info("INFO level message")
+    log.warning("WARN level message")
+    log.error("ERROR level message")
 
 
 if __name__ == "__main__":
--- a/examples/tutorials/basic/running_your_hydra_app/4_logging/my_app.py
+++ b/examples/tutorials/basic/running_your_hydra_app/4_logging/my_app.py
@@ -11,6 +11,8 @@ log = logging.getLogger(__name__)
 
 @hydra.main()
 def my_app(_cfg: DictConfig) -> None:
+    log.error("Error level message")
+    log.warning("Warning level message")
     log.info("Info level message")
     log.debug("Debug level message")
 
--- a/hydra/conf/hydra/hydra_logging/default.yaml
+++ b/hydra/conf/hydra/hydra_logging/default.yaml
@@ -10,11 +10,12 @@ handlers:
     formatter: simple
     stream: ext://sys.stdout
 root:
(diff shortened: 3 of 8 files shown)


omry:
@JackUrb, thinking about his some more. I think am will not proceed with this change.

Hydra is configuring the logging so you don't have to. The default log level in Python (WARNING) is not very useful in my opinion.
If you would like to control the log level you can do it by overriding the root log log level in hydra/job_logging. You can also do it for individual loggers.
```yaml
loggers:
  sh:
    level: WARNING
```

omry:
I am abandoning this.
Hydra is configuring the logging to to something that is useful for the majority of the applications by default.
I think INFO level for the root logger is a useful default.

Applications that does like the default behavior has two options:
1. Customize the logging configuration to match their needs.
2. Completely disable the logging configuration (see #1130 to see how, this will be easier in 1.1)

PR #1269, opened 2021-01-05
Jgbos submitit launcher setup

omry:
@jgbos, something like this.


omry:
Closing as this was not meant as a real PR, just a way to share the code.

PR #1287, opened 2021-01-10
Enable %f pattern in now resolver

forcecore:
## Motivation

I sometimes manually run multiple instances of Hydra configured programs within a fraction of a second. In that case, their output directory's timestamp collide.

The user manual points to python2 strftime document which supports %f pattern (micro seconds), which would be one way of solving this issue.
https://docs.python.org/2/library/datetime.html#strftime-strptime-behavior
However, in Python 3,
https://docs.python.org/3/library/time.html?highlight=strftime#time.strftime
time.strftime does not implement %f format and datetime.stftime does.

Yes

## Test Plan

Simple tutorial codes in the user manual would do the job

main.py:
```
#!/usr/bin/env python

import logging
import hydra

log = logging.getLogger(__name__)


@hydra.main(config_name='config')
def main(cfg):
    log.info(OmegaConf.to_yaml(cfg))
    log.debug("hello world!")


if __name__ == "__main__":
    main()
```

config.yaml:
```
db:
  driver: mysql
  user: omry
  password: secret

hydra:
  run:
    dir: ./outputs/${now:%Y%m%d_%H%M%S_%f}
```

## Related Issues and PRs

None related

### DIFF
--- a/tests/test_hydra.py
+++ b/tests/test_hydra.py
@@ -730,6 +730,24 @@ def test_local_run_workdir(
     )
 
 
+@pytest.mark.parametrize(
+    "task_config",
+    [
+        ({"hydra": {"run": {"dir": "${now:%Y%m%d_%H%M%S_%f}"}}}),
+    ],
+)
+def test_run_dir_microseconds(tmpdir: Path, task_config: DictConfig) -> None:
+    cfg = OmegaConf.create(task_config)
+    assert isinstance(cfg, DictConfig)
+    integration_test(
+        tmpdir=tmpdir,
+        task_config=cfg,
+        overrides=[],
+        prints="str('%f' not in os.getcwd())",
+        expected_outputs="True",
+    )
+
+
 def test_hydra_env_set_with_config(tmpdir: Path) -> None:
     cfg = OmegaConf.create({"hydra": {"job": {"env_set": {"foo": "bar"}}}})
     integration_test(
--- a/hydra/core/utils.py
+++ b/hydra/core/utils.py
@@ -7,9 +7,9 @@ import sys
 import warnings
 from contextlib import contextmanager
 from dataclasses import dataclass
+from datetime import datetime
 from os.path import basename, dirname, splitext
 from pathlib import Path
-from time import localtime, strftime
 from typing import Any, Dict, Optional, Sequence, Tuple, Union, cast
 
 from omegaconf import DictConfig, OmegaConf, open_dict, read_write
@@ -151,7 +151,7 @@ def setup_globals() -> None:
             pass
 
     # please add documentation when you add a new resolver
-    register("now", lambda pattern: strftime(pattern, localtime()))
+    register("now", lambda pattern: datetime.now().strftime(pattern))
     register(
         "hydra",
         lambda path: OmegaConf.select(cast(DictConfig, HydraConfig.get()), path),
--- a/news/1287.bugfix
+++ b/news/1287.bugfix
@@ -1,1 +1,1 @@
-Bug fix to make %f strftime directive to work for ${now:} resolver
+Add support for %f directive (microseconds) to the ${now:} resolver
--- a/news/1287.bugfix
+++ b/news/1287.bugfix
@@ -1,0 +1,1 @@
+Bug fix to make %f strftime directive to work for ${now:} resolver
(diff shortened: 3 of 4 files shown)


shagunsodhani:
Thanks @forcecore Could you add a test for checking that %f is correctly resolved? Maybe we could add this to the docs at https://hydra.cc/docs/next/configure_hydra/workdir

forcecore:
@shagunsodhani 
I think I might write a test that uses %f and assert %f is not found in the run dir.
However, I'm unsure where to put the test. Could you recommend me a file to put the test code?

Maybe along with
test_hydra.py:    get_run_output(["tests/test_apps/run_dir_test/my_app.py"])
and write a new app like this?
"tests/test_apps/run_dir_test_microseconds/my_app.py"

shagunsodhani:
Maybe something like this https://github.com/facebookresearch/hydra/blob/master/tests/test_hydra.py#L718 ?

tests/test_hydra.py
```diff
@@ -72,6 +72,20 @@ def test_run_dir() -> None:
...
+@pytest.mark.parametrize("rundir", ["\\${now:%Y%m%d_%H%M%S_%f}"])
+def test_run_dir_microseconds(tmpdir: Path, rundir: str) -> None:
+    rundir = str(tmpdir / rundir)
+    cmd = [
+        "tests/test_apps/run_dir_test/my_app_microseconds.py",
+        f"hydra.run.dir={rundir}",
+    ]
+    print(" ".join(cmd))
+    out, _err = get_run_output(cmd)
+    print(out)
```
omry, line 84: remove the prints please.

tests/test_hydra.py
```diff
@@ -72,6 +72,20 @@ def test_run_dir() -> None:
     get_run_output(["tests/test_apps/run_dir_test/my_app.py"])
 
 
+@pytest.mark.parametrize("rundir", ["\\${now:%Y%m%d_%H%M%S_%f}"])
+def test_run_dir_microseconds(tmpdir: Path, rundir: str) -> None:
+    rundir = str(tmpdir / rundir)
+    cmd = [
+        "tests/test_apps/run_dir_test/my_app_microseconds.py",
+        f"hydra.run.dir={rundir}",
+    ]
```
omry, line 81: You can use integration_test() instead of this decicated app.
Take a look at how it's being used in other places.

forcecore: It seems to expect something exact, but with microseconds I'm unsure how timestamp will go?

omry: You could print something like:
'%f' not in os.getcwd() and ensures that it's "True".

forcecore: Then I might try assert_regex_match, then it seems to only have use cases with dedicated programs

forcecore:
```
You could print something like:
'%f' not in os.getcwd() and ensures that it's "True".
```

OK I see

forcecore:
I think the new test works. It fails with the old code and passes with the new one.

tests/test_hydra.py
```diff
@@ -730,6 +730,24 @@ def test_local_run_workdir(
...
+    ],
+)
+def test_run_dir_microseconds(tmpdir: Path, task_config: DictConfig) -> None:
+    cfg = OmegaConf.create(task_config)
+    assert isinstance(cfg, DictConfig)
+    integration_test(
+        tmpdir=tmpdir,
+        task_config=cfg,
+        overrides=[],
+        prints="str('%f' not in os.getcwd())",
```
omry, line 746: curious why you need str() here.

omry (changes requested):
1. Please update the doc at https://hydra.cc/docs/configure_hydra/intro
2. Please add a news fragment file. See [this](https://hydra.cc/docs/development/documentation#news-entries).

tests/test_hydra.py
forcecore: print() function appends \n, and the expression becomes str('\f' not in os.getcwd())+ "\n". Without str(), the expression fails.

forcecore: File "/home/xxx/usr/venv-3.9/lib/python3.9/site-packages/hydra/core/utils.py", line 129, in run_job
  File "/tmp/pytest-of-xxx/pytest-7/test_run_dir_microseconds_task0/task.py", line 10, in experiment
    f.write('%f' not in os.getcwd() + "\n")

omry: > TypeError: write() argument must be str, not bool

Okay, so this is the real reason.

forcecore:
* Document updated to point to Python 3's datetime.strftime
* Added "1287.bugfix" with content Fixed hydra.job.id and hydra.job.num not getting passed to jobs in multirun

omry:
Wrong doc link, should be https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior

forcecore:
Doc link updated

omry (approved):
Awesome, thanks!

PR #1293, opened 2021-01-12
remove type-ignore for optuna 2.4.0

samuelstanton:
## Motivation

type-ignore statements are causing the linter to fail.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebookresearch/hydra/blob/master/CONTRIBUTING.md)?

Yes

## Test Plan

N/A

PR #1294, opened 2021-01-12
Add multi-objective optimization support to Optuna sweeper

toshihikoyanase:
## Motivation

As suggested by @omry in this comment (https://github.com/facebookresearch/hydra/pull/1293#issuecomment-758536228), this PR enables Optuna-sweeper users to execute multi-objective optimization.

Please note that this PR replaces the `direction` property of `OptunaConfig` with `directions`. Do you think I should keep it for backward compatibility?

**TODO**
- [x] Update unit tests
- [x] Update documents

The updated document is as follows:
image

Yes

## Test Plan

- Add unit tests to `test_optuna_sweeper_plugin.py`
- Add an example of multi-objective optimization

## Related Issues and PRs

A follow-up of #1293


plugins/hydra_optuna_sweeper/example/conf/config.yaml
```diff
@@ -4,7 +4,7 @@ defaults:
 hydra:
   sweeper:
     optuna_config:
-      direction: minimize
+      directions: [minimize]
```
shagunsodhani, line 7: This seems to be the only breaking change in the config - for people who are doing single objective training.

How about enabling some backward compatibility by supporting both `direction` and `directions` field. In the plugin, we ignore `direction` if `directions` is setup. Else, we set `directions = [direction]`

toshihikoyanase: Thank you for your suggestion. That sounds reasonable. I reverted `OptunaConfig.direction` in the commit https://github.com/facebookresearch/hydra/pull/1294/commits/07bcf4cd6416428792d2851229996217a0e89b98 according to your advice.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
```diff
@@ -200,18 +204,39 @@ def sweep(self, arguments: List[str]) -> None:
...
             for trial, ret in zip(trials, returns):
-                study._tell(trial, optuna.trial.TrialState.COMPLETE, [ret.return_value])
+                if len(self.optuna_config.directions) == 1:
+                    study._tell(
+                        trial, optuna.trial.TrialState.COMPLETE, [ret.return_value]
+                    )
+                else:
+                    study._tell(
+                        trial, optuna.trial.TrialState.COMPLETE, ret.return_value
+                    )
```
omry, line 214: Will Optuna raise a sensible error if the dimension of the return value does not match the dimension of the directions?

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
```diff
@@ -52,7 +54,10 @@ class DistributionConfig:
 class OptunaConfig:
 
     # Direction of optimization
+    # For single objective optimization
     direction: Direction = Direction.minimize
+    # For multi-objective optimization. If exists, direction will be ignored.
+    directions: Optional[List[Direction]] = None
```
omry, line 60: Another approach is to make direction effectively:
```python
direction: Union[Direction, List[Direction] = Direction.minimize
```

Since OmegaConf currently does not support Union (it's planned for the next version), you can use Any for now and perform the type checks yourself.
```python
# Union[Direction, List[Direction]
direction: Any = Direction.minimize
```

Once OmegaConf adds support for Union, we can follow up and take advantage of it here.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
toshihikoyanase: Thank you for pointing it out. The current `Study_tell` just report given values without checking them. I added some exception handling of return values including the dimension mismatch and type incompatibility in https://github.com/facebookresearch/hydra/pull/1294/commits/f1eb00018e05f0aaace8a51a85ffddcff4d7959e.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
toshihikoyanase: That makes sense.
I removed the `directions` according to your suggestion.
BTW, I sometimes got `direction` value as `str` instead of `Direction`, and I need to add type checking as follows.

```python
            if isinstance(self.optuna_config.direction, str):
                directions = [self.optuna_config.direction]
            else:
                directions = [self.optuna_config.direction.name]
```
(https://github.com/toshihikoyanase/hydra/blob/f1eb00018e05f0aaace8a51a85ffddcff4d7959e/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py#L181-L184)

Do you think it Is an expected behavior?

plugins/hydra_optuna_sweeper/example/conf/mo-config.yaml
```diff
@@ -0,0 +1,28 @@
+defaults:
+  - override hydra/sweeper: optuna
```
jieru-hu, line 2: since this is a different example, maybe consider separate them into separate folders?

omry: I agree.
Can you also add a test covering the new example app (and the existing one if it's not already there)?

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
```diff
@@ -160,6 +160,8 @@ def sweep(self, arguments: List[str]) -> None:
             "tpe": "optuna.samplers.TPESampler",
             "random": "optuna.samplers.RandomSampler",
             "cmaes": "optuna.samplers.CmaEsSampler",
+            "nsgaii": "optuna.samplers.NSGAIISampler",
+            "motpe": "optuna.samplers.MOTPESampler",
```
omry, line 164: Some of the samplers are taking parameters:
e.g:
https://optuna.readthedocs.io/en/stable/reference/generated/optuna.samplers.CmaEsSampler.html#

```python
CmaEsSampler(x0=None, sigma0=None, n_startup_trials=1, independent_sampler=None, warn_independent_sampling=True, seed=None, *, consider_pruned_trials=False, restart_strategy=None, inc_popsize=2)[
```

We can have a followup PR to expose those to the user.
If we are okay depending on Hydra 1.1 features (not yet released), we should look at using recursive defaults and having a netsted config group for the samplers.

That would be an interest use case to try out the new capabilities coming in Hydra 1.1.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
```diff
@@ -169,17 +171,29 @@ def sweep(self, arguments: List[str]) -> None:
         sampler_class = get_class(samplers[self.optuna_config.sampler.name])
         sampler = sampler_class(seed=self.optuna_config.seed)
 
+        directions: List[str]
+        if isinstance(self.optuna_config.direction, ListConfig):
```
omry, line 175: This is a better (of course with the proper import):

```suggestion
        if isinstance(self.optuna_config.direction, MutableSequence):
```

omry (changes requested):

plugins/hydra_optuna_sweeper/example/conf/mo-config.yaml
toshihikoyanase: Thank you for your suggestion. That completely makes sense.
I separated the config directories in the commit https://github.com/facebookresearch/hydra/pull/1294/commits/6103edb13c5dde34f323060d2f311aa06291e9de and added the example app to the test in the commit https://github.com/facebookresearch/hydra/pull/1294/commits/d6286ab9a3801b30376bf34eeb5a726b0de575d8.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
toshihikoyanase: I applied your suggestion in the commit https://github.com/facebookresearch/hydra/pull/1294/commits/c20054f3bdf22ed6585b4b9d382e9b70d3fbe679. Thank you!

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
toshihikoyanase: > We can have a followup PR to expose those to the user.
> If we are okay depending on Hydra 1.1 features (not yet released), we should look at using recursive defaults and having a netsted config group for the samplers.

Thank you for letting me know. I agree with you.

I think Optuna users want to specify arguments of samplers. For example, [`population_size`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.samplers.NSGAIISampler.html#optuna.samplers.NSGAIISampler) of `NSGAIISampler` should be changed depending on `n_trials` because it will be identical to `RandomSampler` if the `population_size` is larger than `n_trials`.

In addition to that, we have still a few users according to the [pypistats](https://pypistats.org/packages/hydra-optuna-sweeper), and I think we need to attract new users by adding new functionalities.

omry: > In addition to that, we have still a few users according to the pypistats, and I think we need to attract new users by adding new functionalities.

I think it will mostly take awareness.
Not many people are using HPO plugins with Hydra in general, so I am not surprised that this new HPO plugin is also not getting a lot of usage.

plugins/hydra_optuna_sweeper/example/mo.py
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
omry, line 1: for clarity, can you rename it (and corresponding config dirs etc) to multi-objective.py?

plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
```diff
@@ -148,3 +148,41 @@ def test_optuna_example(with_commandline: bool, tmpdir: Path) -> None:
...
+            assert trial["params"]["y"] % 1 == 0
+            assert trial["values"][0] % 1 == 0
+            assert trial["values"][1] % 1 == 0
+    else:
+        assert len(returns["pareto_front"]) == 13
+        for trial in returns["pareto_front"]:
+            assert trial["params"]["x"] % 1 in {0, 0.5}
+            assert trial["params"]["y"] % 1 in {0, 0.5}
+            assert trial["values"][0] % 1 == 0
+            assert trial["values"][1] % 1 in {0, 0.25, 0.5}
```
omry, line 188: Feels a bit fragile.
How sure are you of these tests?

plugins/hydra_optuna_sweeper/example/mo.py
toshihikoyanase: Thank you for your suggestion. I applied it in https://github.com/facebookresearch/hydra/pull/1294/commits/65e8e7e9ecff6681c20ba4a8bb273f32f1c9957b.

plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
toshihikoyanase: Thank you for pointing it out. I think the assertion of the parameter values must pass if the sweeper parses the configurations (both command-line and configuration file) correctly.

In terms of the assertion of the values, I found that they checked the implementation of Binh and Korn benchmark function. I rewrite them to test if returned values are actually Pareto solutions. The corresponding commit is https://github.com/facebookresearch/hydra/pull/1294/commits/3b66471b0aca440d42e4aaaa4ee22f8ebd6d4bec.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
```diff
@@ -200,18 +214,68 @@ def sweep(self, arguments: List[str]) -> None:
...
+                        study._tell(trial, optuna.trial.TrialState.FAIL, None)
+                        log.warning(
+                            f"Return value must be a float value. Got {ret.return_value}."
+                        )
+                else:
+                    if not isinstance(ret.return_value, (list, tuple)):
+                        study._tell(trial, optuna.trial.TrialState.FAIL, None)
+                        log.warning(
+                            f"Return value is not a sequence. Got {ret.return_value}."
+                        )
```
jieru-hu, line 234: Do we need to return early here or raise an Error? Or the application will error out itself?

toshihikoyanase: Thank you for pointing it out. We need to raise errors because the user code should be fixed to continue optimization correctly. I applied your suggestion in the commit https://github.com/facebookresearch/hydra/pull/1294/commits/78b6f351fffbefe328f26653c392618cb94edc8f.

toshihikoyanase:
I added a section about the multi-objective optimization example to the document. The PR is review-ready now. PTAL.

plugins/hydra_optuna_sweeper/example/multi-objective-conf/config.yaml
```diff
@@ -0,0 +1,28 @@
+defaults:
+  - override hydra/sweeper: optuna
```
shagunsodhani, line 2: just confirming that we do not plan to backport this to 1.0 correct?

website/docs/plugins/optuna_sweeper.md
```diff
@@ -208,3 +208,46 @@ If `log` is `false`, the parameter is mapped to [`UniformDistribution`](https://
...
+    step: 0.5
+```
+
+To run optimization, run the following command in the `plugins/hydra_optuna_sweeper` directoy:
+
+```commandline
+python example/multi-objective.py --multirun
+```
+
+For this problem, no single solution exists that simultaneously minimizes both objectives. Instead, we obtained a set of solutions, namely Pareto optimal solutions, that shapes the trade-off relationship between the objectives. In the following figure, the blue dots show the Pareto optimal solutions in the optimization results.
```
shagunsodhani, line 251: Is `Pareto optimal solutions` something common in multi-objective optimization? If not, we may want to add a small link somewhere?

plugins/hydra_optuna_sweeper/example/multi-objective-conf/config.yaml
omry: good point. If this plugin is targeting Hydra 1.1 setup.py should reflect that.

website/docs/plugins/optuna_sweeper.md
```diff
@@ -208,3 +208,46 @@ If `log` is `false`, the parameter is mapped to [`UniformDistribution`](https://
...
 The parameters are mapped to [`CategoricalDistribution`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.distributions.CategoricalDistribution.html).
+
+
+## Example 2:  Multi-Objective Optimization
+
+We include another example in the directory. `example/multi-objective.py` implements a simple benchmark function which has two objective values. We'd like to minimize both objectives simultaneously.
+
+You can discover the Optuna sweeper parameters with:
+
+```yaml title="python example/multi-objective.py hydra/sweeper=optuna --cfg hydra -p hydra.sweeper"
```
omry, line 219: use <details> tag to fold this output.
See example in other plugins.

omry (changes requested):

website/docs/plugins/optuna_sweeper.md
```diff
@@ -208,3 +208,46 @@ If `log` is `false`, the parameter is mapped to [`UniformDistribution`](https://
...
+    high: 5
+    step: 0.5
+  y:
+    type: float
+    low: 0
+    high: 3
+    step: 0.5
+```
+
+To run optimization, run the following command in the `plugins/hydra_optuna_sweeper` directoy:
```
Crissman, line 245:
```suggestion
To run the optimization, use the following command in the `plugins/hydra_optuna_sweeper` directory:
```

website/docs/plugins/optuna_sweeper.md
Crissman, line 251:
```suggestion
For problems with trade-offs between two different objectives, there may be no single solution that simultaneously minimizes both objectives. Instead, we obtained a set of solutions, namely Pareto optimal solutions, that show the best trade-offs possible between the objectives. In the following figure, the blue dots show the Pareto optimal solutions in the optimization results.
```

Crissman (changes requested):
Some English edits.

plugins/hydra_optuna_sweeper/example/multi-objective-conf/config.yaml
toshihikoyanase: Thank you for your question. This plugin will use the Hydra 1.1 feature as suggested in https://github.com/facebookresearch/hydra/pull/1294#discussion_r557767722, but it still works with Hydra 1.0 for now.

I think we can backport this change to 1.0, but do we have any difficulties?

website/docs/plugins/optuna_sweeper.md
toshihikoyanase: Thank you for pointing it out. In my understanding, it is a common term in the domain of multi-objective optimization. For example, it appears in the [Wikipedia article](https://en.wikipedia.org/wiki/Multi-objective_optimization).

On the other hand, it is not as common as the word `the best solution`, and I should add a link as you suggested. Do you think the link to the [Wikipedia article](https://en.wikipedia.org/wiki/Multi-objective_optimization)?

In addition, I used `pareto_front` and `Pareto solutions` in the key name of the result dictionary and a log message. I'm not sure if it is user-friendly or not. I guess many users are familiar with the word `best solutions`. Do you think I should replace them with `best_solutions` and `Best solutions`?

https://github.com/toshihikoyanase/hydra/blob/b98efa709ca5f43314e8d8bbec7d8f52eed01dcb/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py#L263-L265

website/docs/plugins/optuna_sweeper.md
toshihikoyanase: I applied the `details` tag in the commit https://github.com/facebookresearch/hydra/pull/1294/commits/0f774bc2d1fad317b1a3bb1876df760642372159. Thank you!
image

toshihikoyanase:
@Crissman Thank you for your suggestions! I applied them.

plugins/hydra_optuna_sweeper/example/multi-objective-conf/config.yaml
omry: For now we will not have much difficulties but the above line ('override hydra/sweeper: optuna') will not work in 1.0.
other features we are likely to use in next iterations will make actual use of Hydra 1.1 capabilities and will not be portable.
I suggest that we limit the plugin to 1.1.
maintaining two versions of the plugin at this early stage is not worth it in my opinion. People that wants to use it can use Hydra 1.1 dev release and provide some additional testing for it.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
```diff
@@ -200,18 +214,58 @@ def sweep(self, arguments: List[str]) -> None:
             returns = self.launcher.launch(overrides, initial_job_idx=self.job_idx)
             self.job_idx += len(returns)
             for trial, ret in zip(trials, returns):
-                study._tell(trial, optuna.trial.TrialState.COMPLETE, [ret.return_value])
+                values: Optional[List[float]] = None
+                state: optuna.trial.TrialState = optuna.trial.TrialState.COMPLETE
+                func_error: Optional[Exception] = None
+                try:
+                    if len(directions) == 1:
+                        values = [float(ret.return_value)]
```
omry, line 222: can you raise a proper error if the user does not return a floating point value here?
blindly casting to float is not great.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
```diff
@@ -200,18 +214,58 @@ def sweep(self, arguments: List[str]) -> None:
...
+                    else:
+                        if not isinstance(ret.return_value, (list, tuple)):
+                            raise ValueError(
+                                f"Return value must be a float value. Got {ret.return_value}."
+                            )
+                        values = [float(v) for v in ret.return_value]
```
omry, line 228: error out of one if the return values is not a float.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
```diff
@@ -200,18 +214,58 @@ def sweep(self, arguments: List[str]) -> None:
...
+                            raise ValueError(
+                                "The number of the values and the number of the objectives are"
+                                f" mismatched. Expect {len(directions)}, but actually "
+                                f"{len(values)}."
+                            )
+                except Exception as e:
+                    state = optuna.trial.TrialState.FAIL
+                    func_error = e
+                finally:
+                    study._tell(trial, state, values)
```
omry, line 239: That's a bit creative usage of finally.
I think it's better to do:
```python
try:
  ...
  study._tell(trial, state, values)
except Exception as e:
  study._tell(trial, optuna.trial.TrialState.FAIL, values)
  raise e
  
```

website/docs/plugins/optuna_sweeper.md
```diff
@@ -6,7 +6,7 @@ sidebar_label: Optuna Sweeper plugin
 
 import GithubLink,{ExampleGithubLink} from "@site/src/components/GithubLink"
 
-[![PyPI](https://img.shields.io/pypi/v/hydra-optuna-sweeper)](https://img.shields.io/pypi/v/hydra-optuna-sweeper)
+[![PyPI](https://img.shields.io/pypi/v/hydra-optuna-sweeper)](https://pypi.org/project/hydra-optuna-sweeper/)
```
omry, line 9: If we decide to officially support Hydra 1.1 and newer here we should mention it early in the readme to avoid surprises by users installing the plugin and having their Hydra upgraded to 1.1.

omry (changes requested):

plugins/hydra_optuna_sweeper/example/multi-objective-conf/config.yaml
toshihikoyanase: Thank you for your explanation. I was not aware of the use of Hydra 1.1 features.

> maintaining two versions of the plugin at this early stage is not worth it in my opinion.

I agree. The development resource is limited, and it sounds reasonable to me.

> People that wants to use it can use Hydra 1.1 dev release and provide some additional testing for it.

Ah, you already released [`v1.1.0.dev3`](https://pypi.org/project/hydra-core/1.1.0.dev3/) at PyPI. Users can easily use it with `pip install --pre hydra`.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
toshihikoyanase: That makes sense. I fixed it in https://github.com/facebookresearch/hydra/pull/1294/commits/22e4e528af58f6321b4c03814fc5e95974cae4ec.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
toshihikoyanase: Thank you for pointing it out. I raised `ValueError` here in https://github.com/facebookresearch/hydra/pull/1294/commits/30b50e48e4a0d073ea7c892b4cddae8808ff33e6.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
toshihikoyanase: Thanks. I also fix it in https://github.com/facebookresearch/hydra/pull/1294/commits/30b50e48e4a0d073ea7c892b4cddae8808ff33e6.

website/docs/plugins/optuna_sweeper.md
toshihikoyanase: Thank you for your suggestion.
I set the version limitation of the hydra-core as `hydra-core>1.0` in the commit https://github.com/facebookresearch/hydra/pull/1294/commits/9b231efab9fe2dd347847f45d4da80810805f32d. And then, I explained the version limitation of Hydra in the installation section in https://github.com/facebookresearch/hydra/pull/1294/commits/d71aec8aec783312c1c2d8ae6ec291ae71e01588.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
```diff
@@ -216,30 +216,38 @@ def sweep(self, arguments: List[str]) -> None:
             for trial, ret in zip(trials, returns):
                 values: Optional[List[float]] = None
                 state: optuna.trial.TrialState = optuna.trial.TrialState.COMPLETE
-                func_error: Optional[Exception] = None
                 try:
                     if len(directions) == 1:
-                        values = [float(ret.return_value)]
+                        try:
+                            values = [float(ret.return_value)]
+                        except Exception:
```
omry, line 223: too broad. ValueError?

omry: Pro tip about re-raising exceptions (which is not critical here but good to know):

The following will preserve the current stack trace instead of starting a new one:
```
raise raise ValueError(...).with_traceback(sys.exc_info()[2])
```

This is not making a big deal here because the error is happening just two lines above the exception, but in many cases the error is caught far away from where it happens and rethrowing a new exception makes it hard to know what went wrong without careful debugging.

No need to apply in this case.

omry (changes requested):

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
toshihikoyanase: > too broad. ValueError?

That must be fixed. Thanks. I replaced `Exception` with a pair of `ValueError` and `TypeError`.  I added `TypeError` because `float(None)` raises it in https://github.com/facebookresearch/hydra/pull/1294/commits/66b136d5378b20c42928fa69822c5d441fc96419.

> `raise raise ValueError(...).with_traceback(sys.exc_info()[2])`

Thank you for sharing the tips. I applied the suggestion in the commit https://github.com/facebookresearch/hydra/pull/1294/commits/66b136d5378b20c42928fa69822c5d441fc96419 because users can see the line of the error cause without opening the source code. If you find it verbose, I'll revert the change.

Previous
```console
  File "/prj/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py", line 251, in sweep
    raise e
  File "/prj/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py", line 226, in sweep
    f"Return value must be float-castable. Got '{ret.return_value}'."
ValueError: Return value must be float-castable. Got 'None'.
``` 

https://github.com/facebookresearch/hydra/pull/1294/commits/66b136d5378b20c42928fa69822c5d441fc96419
```console
  File "/prj/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py", line 227, in sweep
  File "/prj/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py", line 223, in sweep
    values = [float(ret.return_value)]
ValueError: Return value must be float-castable. Got 'None'.
```

plugins/hydra_optuna_sweeper/setup.py
```diff
@@ -32,7 +32,7 @@ def get_long_description() -> str:
         "Development Status :: 4 - Beta",
     ],
     install_requires=[
-        "hydra-core",
+        "hydra-core>1.0",
```
omry, line 35: This expresses the intention better in my opinion:
```
hydra-core>=1.1
```

omry (approved):
Overall looks good.
I have one request to simplify one area in the code, once it's done we can merge this.

omry (changes requested):
ah, accidentally approved. meant to request changes.

plugins/hydra_optuna_sweeper/setup.py
toshihikoyanase: Thank you for your suggestion, but I got the following installation error with `hydra-core>=1.1` when I ran the nox's `"lint_plugins-3.8` task.

```console
ERROR: Could not find a version that satisfies the requirement hydra-core>=1.1 (from hydra-optuna-sweeper)              ERROR: No matching distribution found for hydra-core>=1.1                                                               
WARNING: You are using pip version 20.3.1; however, version 21.0.1 is available.                                        
You should consider upgrading via the '/home/yanase/pfn/code/hydra/.nox/lint_plugins-3-8/bin/python -m pip install --upgrade pip' command.                                                                                                      
nox > Session lint_plugins-3.8 failed. 
```

I think this is because `hydra-core==1.1.0` has not been released yet, and I reverted the change with TODO comment (https://github.com/facebookresearch/hydra/pull/1294/commits/0fe11b699ce60e0622e4570a1b331caa3e74e500).

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
```diff
@@ -200,18 +214,66 @@ def sweep(self, arguments: List[str]) -> None:
...
+                            raise ValueError(
+                                "Return value must be a list or tuple of float-castable values."
+                                f" Got '{ret.return_value}'."
+                            )
+                        try:
+                            values = [float(v) for v in ret.return_value]
+                        except (ValueError, TypeError):
+                            raise ValueError(
+                                "Return value must be a list or tuple of float-castable values."
+                                f" Got '{ret.return_value}'."
```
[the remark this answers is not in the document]
toshihikoyanase: Thank you for pointing it out.
I think `if not isinstance(ret.return_value, (list, tuple))` is not required because the try-except from L233-L239 raises the same error without it. I removed it in the commit https://github.com/facebookresearch/hydra/pull/1294/commits/7d1b40fc0e11631bad01eb144e1e8f97f39b58cd.

website/docs/plugins/optuna_sweeper.md
toshihikoyanase: Regarding the `Pareto optimal solutions`, I add a link to a Wikipedia article about [Pareto efficiency](https://en.wikipedia.org/wiki/Pareto_efficiency) in the commit https://github.com/facebookresearch/hydra/pull/1294/commits/976a916a0ec601400e9ac1469652eebf83c80f30.

> I guess many users are familiar with the word best solutions or solutions. Do you think I should replace them with best_solutions or solutions?

I replaced `pareto_front` that is a key name in the result yaml with `solutions` in the commit (https://github.com/facebookresearch/hydra/pull/1294/commits/d27267c1739b4a8b20238e3b502e78116a6abe5d).

PTAL.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
```diff
@@ -77,6 +80,12 @@ class OptunaConfig:
     # Random seed of sampler
     seed: Optional[int] = None
 
+    def __post_init__(self) -> None:
+        if isinstance(self.direction, list) and len(self.direction) > 1:
+            if self.sampler == Sampler.tpe:
+                # The default sampler of multi-objective optimization is NSGAIISampler
+                self.sampler = Sampler.nsgaii
```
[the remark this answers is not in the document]
toshihikoyanase: > we could introduce a opt_mode [single|multi] config, each with it's own defaults list.
> The multi one defaults list can override the default sample to nsgaii.

Sounds great! I'll work on it as a follow-up PR.

plugins/hydra_optuna_sweeper/example/multi-objective-conf/config.yaml
toshihikoyanase: > We are not using any yet. but I mentioned that we can configure the sampler with Hydra.

Oh, I misunderstood. `override hydra/sweeper: optuna` is just used in the example config.

> Going with 1.1 does hurt the visibility of the plugin in the short term, but will also allow us to focus on making it better, leveraging 1.1 features.

Agreed. I'd like to focus on the improvement with the Hydra 1.1 features as you mentioned. I'll start from the [recursive defaults](https://github.com/facebookresearch/hydra/pull/1294#discussion_r568890895).

omry:
Looking good.
Can you add a news fragment in news directory?

toshihikoyanase:
I added a news fragment for this PR in the commit https://github.com/facebookresearch/hydra/pull/1294/commits/313e20ba86aa8bbc4e6d3336d991dba50611d455.
I appreciate your careful reviews. They definitely improved the implementation.

omry:
Awesome, thanks!
I am merging this in, I think the best followup would be to try to utilize recursive defaults to configure the plugin in a more powerful way (including samplers, and maybe even explicit separation of single and multi objective optimization modes in a config group.

PR #1317, opened 2021-01-22
Add special check for zipfile.Path

jieru-hu:
## Motivation

When runnning in `par` format we get a `zipfile.Path` instaed of `pathlib.Path`, as a result tests fail in fbcode.

This change is to fix that. 

Yes/No

## Test Plan

CI here

tested on both platforms as well (007 + 009)

## Related Issues and PRs

### DIFF
--- a/hydra/_internal/core_plugins/importlib_resources_config_source.py
+++ b/hydra/_internal/core_plugins/importlib_resources_config_source.py
@@ -4,6 +4,8 @@ import sys
 import zipfile
 from typing import List, Optional
 
+from pkg_resources import parse_version
+
 from omegaconf import OmegaConf
 
 from hydra.core.object_type import ObjectType
@@ -41,19 +43,20 @@ class ImportlibResourcesConfigSource(ConfigSource):
         if not res.exists():
             raise ConfigLoadError(f"Config not found : {normalized_config_path}")
 
-        if hasattr(zipfile, "Path") and isinstance(res, zipfile.Path):  # type: ignore
-            with res.open() as f:
-                header_text = f.read(512)
-                header_text = header_text.decode("utf-8")
-                header = ConfigSource._get_header_dict(header_text)
-                f.seek(0)
-                cfg = OmegaConf.load(f)
-        else:
-            with res.open(encoding="utf-8") as f:
-                header_text = f.read(512)
-                header = ConfigSource._get_header_dict(header_text)
-                f.seek(0)
-                cfg = OmegaConf.load(f)
+        vi = sys.version_info
+        python_version = f"{vi.major}.{vi.minor}.{vi.micro}"
+        zipfile_path = parse_version(python_version) >= parse_version(
+            "3.8"
+        ) and isinstance(res, zipfile.Path)
+
+        f = res.open() if zipfile_path else res.open(encoding="utf-8")
+        header_text = f.read(512)
+        if zipfile_path:
+            header_text = header_text.decode("utf-8")
+        header = ConfigSource._get_header_dict(header_text)
+        f.seek(0)
+        cfg = OmegaConf.load(f)
+        f.close()
 
         self._update_package_in_header(
             header=header,
--- a/hydra/_internal/core_plugins/importlib_resources_config_source.py
+++ b/hydra/_internal/core_plugins/importlib_resources_config_source.py
@@ -1,6 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 import os
 import sys
+import zipfile
 from typing import List, Optional
 
 from omegaconf import OmegaConf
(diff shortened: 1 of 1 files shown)


hydra/_internal/core_plugins/importlib_resources_config_source.py
```diff
@@ -40,23 +41,32 @@ def load_config(
...
-                provider=self.provider,
-                header=header,
-            )
+        if isinstance(res, zipfile.Path):  # type: ignore
+            with res.open() as f:
+                header_text = f.read(512)
+                header_text = header_text.decode("utf-8")
+                header = ConfigSource._get_header_dict(header_text)
+                f.seek(0)
+                cfg = OmegaConf.load(f)
```
jieru-hu, line 50: I've tested in fbcode with unicode (Chinese) in the config file, OmegaConf can handle it as well, as a result we only need to decode the header text

omry: 1. looks like zipfile.Path was [introduced in Python 3.8](https://docs.python.org/3/library/zipfile.html). It will be cleaner to test that python is 3.8 or newer instead of checking for the presence of zipfile.Path,
2. Can you minimize the duplicate code between the two branches?
3. I really want to find a way to test this code.

omry (changes requested):

hydra/_internal/core_plugins/importlib_resources_config_source.py
```diff
@@ -40,23 +43,33 @@ def load_config(
...
-            f.seek(0)
-            cfg = OmegaConf.load(f)
-            return ConfigResult(
-                config=self._embed_config(cfg, header["package"]),
-                path=f"{self.scheme()}://{self.path}",
-                provider=self.provider,
-                header=header,
-            )
+        vi = sys.version_info
+        python_version = f"{vi.major}.{vi.minor}.{vi.micro}"
```
omry, line 47:
```
sys.version_info[0:2] < (3,8)
```

hydra/_internal/core_plugins/importlib_resources_config_source.py
```diff
@@ -40,23 +43,33 @@ def load_config(
...
+        zipfile_path = parse_version(python_version) >= parse_version(
+            "3.8"
+        ) and isinstance(res, zipfile.Path)
+
+        f = res.open() if zipfile_path else res.open(encoding="utf-8")
```
omry, line 52: I prefer
```
if foo:
  res = ...
else:
  res = ...
```
in this case as it makes it more obvious what is happening.
It will also allow you to place additional logic in each branch.

hydra/_internal/core_plugins/importlib_resources_config_source.py
jieru-hu: nice!

hydra/_internal/core_plugins/importlib_resources_config_source.py
```diff
@@ -40,23 +41,32 @@ def load_config(
...
-            )
+        zipfile_path = sys.version_info[0:2] >= (3, 8) and isinstance(res, zipfile.Path)
+        if zipfile_path:
+            f = res.open()
+        else:
+            f = res.open(encoding="utf-8")
+
+        header_text = f.read(512)
+        if zipfile_path:
+            header_text = header_text.decode("utf-8")
```
omry, line 52:
```suggestion
        zipfile_path = sys.version_info[0:2] >= (3, 8) and isinstance(res, zipfile.Path)
        if zipfile_path:
            f = res.open()
            header_text = f.read(512)            
        else:
            f = res.open(encoding="utf-8")
            header_text = f.read(512).decode("utf-8")
```

hydra/_internal/core_plugins/importlib_resources_config_source.py
```diff
@@ -40,23 +41,32 @@ def load_config(
...
+        header = ConfigSource._get_header_dict(header_text)
+        f.seek(0)
+        cfg = OmegaConf.load(f)
+        f.close()
```
omry, line 56: close in a finally block.
an exception between open and close and you wouldn't be closing the file at all.

hydra/_internal/core_plugins/importlib_resources_config_source.py
```diff
@@ -40,23 +41,32 @@ def load_config(
         if not res.exists():
             raise ConfigLoadError(f"Config not found : {normalized_config_path}")
 
-        with res.open(encoding="utf-8") as f:
-            header_text = f.read(512)
+        zipfile_path = sys.version_info[0:2] >= (3, 8) and isinstance(res, zipfile.Path)
+        try:
+            if zipfile_path:
```
omry, line 46:
```suggestion
        try:
            if sys.version_info[0:2] >= (3, 8) and isinstance(res, zipfile.Path):
```

omry: Maybe, your call :)

omry (approved):

hydra/_internal/core_plugins/importlib_resources_config_source.py
jieru-hu: fixed, sorry shoulda spotted that myself :)

jieru-hu:
tested in fbcode again on platform007 & 009, all passed. merging now.

omry:
Looks like we forgot about the test.
I still want a test for this.

jieru-hu:
> Looks like we forgot about the test.
> I still want a test for this.

will add soon :)

PR #1374, opened 2021-02-03
Update nevergrad default optim to NGOpt

jrapin:
## Motivation

`NGOpt` is the best algorithm overall, it adapts to the current settings.

Yes

## Test Plan

N/A

## Related Issues and PRs

N/A

### DIFF
--- a/plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/config.py
+++ b/plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/config.py
@@ -42,11 +42,11 @@ class OptimConf:
     #   - "CMA" very good algorithm, but may require a significant budget (> 120)
     #   - "TwoPointsDE": an algorithm good in a wide range of settings, for significant
     #     budgets (> 120).
-    #   - "Shiwa" an algorithm aiming at identifying the best optimizer given your input
-    #     definition (work in progress, it may still be ill-suited for low budget)
+    #   - "NGOpt" an algorithm aiming at identifying the best optimizer given your input
+    #     definition (updated regularly)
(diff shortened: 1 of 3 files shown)


jrapin:
@omry initially I was also planning to add a way to catch jobs errors and send infinity in that case instead of raising (if it does not happen on all jobs). Is there a way to make something like this work?

omry:
1. Please add a news fragment in the news directory inside the plugin.
2. Not currently, this would require a change in all launchers. It does make sense to support partial failure though, can you file a separate enhancement request?

omry (changes requested):
This is an important change. can you add a news fragment file?

omry (approved):

omry:
Thanks @jrapin.
Do you want a new version released as well or can this wait for something more significant to come along?

jrapin:
> Thanks @jrapin.
> Do you want a new version released as well or can this wait for something more significant to come along?

Up to you. I'd say it's probably significant enough (NGOpt is supposed to be significantly better than OnePlusOne), but there is no hurry anyway (and my biggest concern was the documentation, which is fixed)

omry:
I can give you release access to the plugin.
@shagunsodhani, do you mind documenting the current plugin release process (with towncrier news update)?

jrapin:
> I can give you release access to the plugin.
> @shagunsodhani, do you mind documenting the current plugin release process (with towncrier news update)?

Could be useful indeed, although I would not push without your go (except emergency, which I don't expect to happen :D)
And indeed, I'd need to know how you do the releases

omry:
Invited you as a maintainer to https://pypi.org/project/hydra-nevergrad-sweeper/

I also invited you and @gwenzek as maintainers to the submitit plugin.

PR #1375, opened 2021-02-03
[Submitit-Plugin] Add support for SLURM parameters `cpus_per_gpu`, `gpus_per_task`, `mem_per_gpu` and `mem_per_cpu`

### LINKED ISSUE #1366: [Feature Request] [Submitit-Plugin] (Potentially a bug) Impossible to set certain flags in submitit launcher
# 🚀 Feature Request

## Motivation

This could be read as a feature request or a bug report, I'm not sure how you want to consider it, I'm going with feature request. Currently theres some flags that you cannot set in the submitit launcher. I practice "gpu centric" scheduling, so I like to specify `mem_per_gpu` and `cpus_per_gpu` and then I can just use `gpus_per_task` to always get the optimal settings. 

For example on "mystery cluster" (you know the one), we can use 10 cpus and 64GB RAM per GPU. These settings allow me to only have to change `gpus_per_task` for example if I need to have 2 GPUs for 2 different models, then I'll automatically get 20 cpus and 128GB RAM for each of the tasks without having to change all of the settings. I've already PRed stuff related to this into submitit.

The problem occurs when you try to set, via additional parameters, something like `mem_per_gpu`. You can't set `mem` and `mem_per_gpu`, slurm just crashes when you do that. Similarly, if you try to set `cpus_per_gpu` via `additional_parameters` you'll wind up setting that in addition to `cpus_per_task`.

## Pitch

I tried making a simple patch that fixes this, but it hits issues with the `AutoExecutor` which I never updated in submitit to be aware of the options which conflict with each other. I think in general auto is missing some of the recent work in submitit, and it feels like a semi-hacky workaround to me. 

Submitit already has pretty good validation logic so my pitch is to (1) let submitit handle everything by calling the correct executor for the job instead of using Auto. This is going to require a revamp of how the parameters are named/passed however and will likely be a breaking change to the API. (2) It would be nice if we could allow people to pass whatever parameters are supported by submitit without needing to update the hydra schema each time. These are my two major goals, so they sound reasonable/feasible? Point (2) may not be possible I guess.

Other options:

* Use auto but try to include our own validation logic (needs exploration), I don't like this because it requires us to duplicate logic that submitit has already and also it may not even be possible depending on what Auto decides to do (looked iffy on my quick pass)
* Punt this to submitit to revamp the auto executor to do the right stuff and be more flexible. I don't like this because I think the auto executor is more of a convenience hack and more complex/flexible code would be calling the executors directly (this is what I was doing previously with my own submitit hydra wrapper thing). I also think its going to be a more more complex PR.

Yes, but I want feedback first on the best way to go about it

Queuecumber:
## Motivation

It's currently impossible to set certain advanced scheduling options, mainly around GPUs, using the hydra submitit plugin because they conflict with each other. In order to make it work, some of the base options need to be made `Optional`.

This PR makes two updates to the submitit laucher:

1. It makes the following parameters optional: [`cpus_per_task`, `gpus_per_node`, `mem_gb`] which conflict with the advanced options.
2. It adds in some parameters that are supported in the slurm executor but were missing from the hydra plugin.

Yes

## Test Plan

I've confirmed that the generated sbatch is correct based on the given options

## Related Issues and PRs

Closes #1366

### DIFF
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
@@ -14,13 +14,13 @@ class BaseQueueConf:
     # maximum time for the job in minutes
     timeout_min: int = 60
     # number of cpus to use for each task
-    cpus_per_task: int = 1
+    cpus_per_task: Optional[int] = None
     # number of gpus to use on each node
-    gpus_per_node: int = 0
+    gpus_per_node: Optional[int] = None
     # number of tasks to spawn on each node
     tasks_per_node: int = 1
     # memory to reserve for the job on each node (in GB)
-    mem_gb: int = 4
+    mem_gb: Optional[int] = None
     # number of nodes to use for the job
     nodes: int = 1
     # name of the job
@@ -46,6 +46,10 @@ class SlurmQueueConf(BaseQueueConf):
     comment: Optional[str] = None
     constraint: Optional[str] = None
     exclude: Optional[str] = None
+    cpus_per_gpu: Optional[int] = None
+    gpus_per_task: Optional[int] = None
+    mem_per_gpu: Optional[str] = None
+    mem_per_cpu: Optional[str] = None
 
     # Following parameters are submitit specifics
     #
--- a/plugins/hydra_submitit_launcher/news/1366.feature
+++ b/plugins/hydra_submitit_launcher/news/1366.feature
@@ -1,1 +1,1 @@
-Add support for advanced GPU scheduling parameters
+Add support for SLURM parameters `cpus_per_gpu`, `gpus_per_task`, `mem_per_gpu` and `mem_per_cpu`
--- a/plugins/hydra_submitit_launcher/news/1366.feature
+++ b/plugins/hydra_submitit_launcher/news/1366.feature
@@ -1,0 +1,1 @@
+Add support for advanced GPU scheduling parameters
--- a/website/docs/plugins/submitit_launcher.md
+++ b/website/docs/plugins/submitit_launcher.md
@@ -33,29 +33,41 @@ the launching host and the target host.
 Submitit actually implements 2 different launchers: `submitit_slurm` to run on a SLURM cluster, and `submitit_local` for basic local tests.
 
 You can discover the SLURM Launcher parameters with:
+
+<details><summary>Discover SLURM launcher's config</summary>
+
 ```yaml title="$ python your_app.py hydra/launcher=submitit_slurm --cfg hydra -p hydra.launcher"
 # @package hydra.launcher
-_target_: hydra_plugins.hydra_submitit_launcher.submitit_launcher.SlurmLauncher
-submitit_folder: ${hydra.sweep.dir}/.submitit/%j
+submitit_folder: $&#123;hydra.sweep.dir/.submitit/%j
 timeout_min: 60
-cpus_per_task: 1
-gpus_per_node: 0
+cpus_per_task: null
+gpus_per_node: null
 tasks_per_node: 1
-mem_gb: 4
+mem_gb: null
 nodes: 1
 name: ${hydra.job.name}
+_target_: hydra_plugins.hydra_submitit_launcher.submitit_launcher.SlurmLauncher
 partition: null
 comment: null
 constraint: null
 exclude: null
+cpus_per_gpu: null
+gpus_per_task: null
+mem_per_gpu: null
+mem_per_cpu: null
 signal_delay_s: 120
 max_num_timeout: 0
 additional_parameters: {}
 array_parallelism: 256
+setup: null
 
 ```
+</details>
 
 Similarly, you can discover the local launcher parameters with:
+
+<details><summary>Discover local launcher's config</summary>
+
 ```yaml title="$ python example/my_app.py hydra/launcher=submitit_local --cfg hydra -p hydra.launcher"
 # @package hydra.launcher
 _target_: hydra_plugins.hydra_submitit_launcher.submitit_launcher.LocalLauncher
@@ -68,6 +80,7 @@ mem_gb: 4
 nodes: 1
 name: ${hydra.job.name}
 ```
+</details>
 
 You can set all these parameters in your configuration file and/or override them in the commandline: 
 ```text
--- a/website/docs/plugins/submitit_launcher.md
+++ b/website/docs/plugins/submitit_launcher.md
@@ -60,7 +60,7 @@ array_parallelism: 256
 setup: null
 ```
 </details>
-<details><summary>Discover the local launcher parameters <b>(Expand)</b></summary>
+<details><summary>Discover the Local Launcher parameters <b>(Expand)</b></summary>
 
 ```yaml title="$ python example/my_app.py hydra/launcher=submitit_local --cfg hydra -p hydra.launcher"
 # @package hydra.launcher
--- a/website/docs/plugins/submitit_launcher.md
+++ b/website/docs/plugins/submitit_launcher.md
@@ -30,13 +30,11 @@ defaults:
 Note that this plugin expects a valid environment in the target host. usually this means a shared file system between
 the launching host and the target host.
 
-Submitit actually implements 2 different launchers: `submitit_slurm` to run on a SLURM cluster, and `submitit_local` for basic local tests.
+The Submitit Plugin implements 2 different launchers: `submitit_slurm` to run on a SLURM cluster, and `submitit_local` for basic local tests.
 
-You can discover the SLURM Launcher parameters with:
+<details><summary>Discover the SLURM Launcher parameters <b>(Expand)</b></summary>
 
-<details><summary>$ python your_app.py hydra/launcher=submitit_slurm --cfg hydra -p hydra.launcher <b>(Expand)<b/></summary>
-
-```yaml
+```yaml title="$ python your_app.py hydra/launcher=submitit_slurm --cfg hydra -p hydra.launcher"
 # @package hydra.launcher
 submitit_folder: $&#123;hydra.sweep.dir/.submitit/%j
 timeout_min: 60
@@ -62,12 +60,9 @@ array_parallelism: 256
 setup: null
 ```
 </details>
+<details><summary>Discover the local launcher parameters <b>(Expand)</b></summary>
 
-Similarly, you can discover the local launcher parameters with:
-
-<details><summary>$ python example/my_app.py hydra/launcher=submitit_local --cfg hydra -p hydra.launcher <b>(Expand)<b/></summary>
-
-```yaml
+```yaml title="$ python example/my_app.py hydra/launcher=submitit_local --cfg hydra -p hydra.launcher"
 # @package hydra.launcher
 _target_: hydra_plugins.hydra_submitit_launcher.submitit_launcher.LocalLauncher
 submitit_folder: ${hydra.sweep.dir}/.submitit/%j
@@ -81,7 +76,9 @@ name: ${hydra.job.name}
 ```
 </details>
 
-You can set all these parameters in your configuration file and/or override them in the commandline: 
+<br/>
+You can set all these parameters in your configuration file and/or override them in the command-line:
+
 ```text
 python foo.py --multirun hydra/launcher=submitit_slurm hydra.launcher.timeout_min=3
 ```
--- a/website/docs/plugins/submitit_launcher.md
+++ b/website/docs/plugins/submitit_launcher.md
@@ -34,7 +34,7 @@ Submitit actually implements 2 different launchers: `submitit_slurm` to run on a
 
 You can discover the SLURM Launcher parameters with:
 
-<details><summary>$ python your_app.py hydra/launcher=submitit_slurm --cfg hydra -p hydra.launcher (click to expand)</summary>
+<details><summary>$ python your_app.py hydra/launcher=submitit_slurm --cfg hydra -p hydra.launcher <b>(Expand)<b/></summary>
 
 ```yaml
 # @package hydra.launcher
@@ -65,7 +65,7 @@ setup: null
 
 Similarly, you can discover the local launcher parameters with:
 
-<details><summary>$ python example/my_app.py hydra/launcher=submitit_local --cfg hydra -p hydra.launcher (click to expand)</summary>
+<details><summary>$ python example/my_app.py hydra/launcher=submitit_local --cfg hydra -p hydra.launcher <b>(Expand)<b/></summary>
 
 ```yaml
 # @package hydra.launcher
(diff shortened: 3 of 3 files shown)


Queuecumber:
CC @jrapin

Queuecumber:
Not sure why the ray launcher test is failing

jieru-hu:
> Not sure why the ray launcher test is failing

taking a look..

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
```diff
@@ -14,13 +14,13 @@ class BaseQueueConf:
...
-    cpus_per_task: int = 1
+    cpus_per_task: Optional[int] = None
     # number of gpus to use on each node
-    gpus_per_node: int = 0
+    gpus_per_node: Optional[int] = None
     # number of tasks to spawn on each node
     tasks_per_node: int = 1
     # memory to reserve for the job on each node (in GB)
-    mem_gb: int = 4
+    mem_gb: Optional[int] = None
```
omry, line 23: If those parameters should not always be passed, why are they in the Base class instead of in the subclass SlurmQueueConf?

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -111,7 +111,7 @@ def launch(
             **{
                 f"{self._EXECUTOR}_{x}": y
                 for x, y in params.items()
-                if x in specific_init_keys
+                if x in specific_init_keys and y is not None
```
omry, line 114: I would rather avoid it.
The config should contain what we need and nothing more.

Queuecumber: There's a second issue here where the AutoExecutor won't accept `None` values because of some of type checking it does. If we simply don't pass them it works as expected which is why I'm filtering them out here

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
Queuecumber: That's a good question, I'm happy to move them if it makes sense but we can wait to hear from @jrapin tomorrow to see if a reorg makes sense

omry: Let's do it and see if everything holds.
Once we do - we can eliminate the None filtering. (None can be a perfectly legitimate value of some future parameter).

Queuecumber: I think we still need the none filtering because it's AutoExecutor that has the issue and everything ends up there. So even if we move to slurm it still goes through auto which complains if mem_gb is set to None instead of an int (but works fine if we just don't pass mem_gb). 

Basically the issue is that submitit is designed to have optional values be not passed at all vs assigned None

omry: So maybe AutoExecutor should be fixed to accept none for mutually exclusive parameters and do the proper validation internally?

Queuecumber: The issue with that is that the validation is done in the slurmexecutor already and would need to be duplicated in auto. We could make auto less picky about params and delegate that authority to the specific executor that gets used however. I think @jrapin didn't like the idea and indicated that the missing-value-instead-of-None thing was an intentional decision, so let's wait for his feedback

omry: So I am a bit confused.
If AutoExecutor is in the business of validating the input, it should be okay for it to apply specific logic to mutually exclusive fields.
If it's not, it should also not care if any of the input fields are None and should just pass them as is to the underlying executor.

Happy to wait for some feedback from jrapin.

Queuecumber: My understanding is that the AutoExecutor does two things

1. Route the call to the correct executor
2. Remove executor prefixes (like "slurm_") so that parameter names are correct.

It also seems to make sure that certain parameters which are considered shared are the correct type that the underlying executor expects but does no further validation. 

It could be argued that the type checking is simply incorrect since, using mem as an example, it checks for int but the slurm executor expects an optional int

Queuecumber:
One other thing that I'd like to be considered: what exactly is the function of the local submitit executor in hydra? Is it doing something different from running your script without multirun?

omry:
> One other thing that I'd like to be considered: what exactly is the function of the local submitit executor in hydra? Is it doing something different from running your script without multirun?

Most remote executors has a local mode. (Ray, Submitit, Flow). The local mode is usually meant to behave somewhat like the remote one but work locally. The primary purpose is to allow for easier testing (In fact, there is no actual slurm testing at all for the Submitit Hydra Launcher plugin right now, and all testing is done using the local executor.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
jrapin: `BaseQueueConf`  contains the parameters that are shared between different executors, and supported by all of the executors, specific configurations derive from it because they are build on top of these shared parameters, adding different options.

> It could be argued that the type checking is simply incorrect since, using mem as an example, it checks for int but the slurm executor expects an optional int

mem is str, but it checks `mem_gb` is int, which it should be, so nothing incorrect here.

> If AutoExecutor is in the business of validating the input, it should be okay for it to apply specific logic to mutually exclusive fields.

It does not validate the inputs, slurm does, we don't want to duplicate validation.

> Basically the issue is that submitit is designed to have optional values be not passed at all vs assigned None

Exactly. So far the behavior of submitit is that if those base parameters are passed, they shoul be used. This behavior *could* be changed though so that `None` are ignored, I can check what Guillaume thinks of it. In any case, if you want to avoid incompatibilities I would advise to filter as well, at least for the time being.

Queuecumber: OK and as for 

> None can be a perfectly legitimate value of some future parameter

Is it safe to assume, for now, that `None` will not be a valid value or will be equivalent in meaning to not passing the parameter?

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
gwenzek: > The config should contain what we need and nothing more.

I agree with that.

>  AutoExecutor won't accept None values

we can also fix this on AutoExecutor side, see eg: https://github.com/facebookincubator/submitit/pull/1607

website/docs/plugins/submitit_launcher.md
```diff
@@ -35,23 +35,28 @@ Submitit actually implements 2 different launchers: `submitit_slurm` to run on a
 You can discover the SLURM Launcher parameters with:
 ```yaml title="$ python your_app.py hydra/launcher=submitit_slurm --cfg hydra -p hydra.launcher"
 # @package hydra.launcher
-_target_: hydra_plugins.hydra_submitit_launcher.submitit_launcher.SlurmLauncher
 submitit_folder: ${hydra.sweep.dir}/.submitit/%j
 timeout_min: 60
-cpus_per_task: 1
-gpus_per_node: 0
+cpus_per_task: null
```
gwenzek, line 40: is this mandatory ? Do users have to type `cpus_per_task: null` ?
It feels really hostile.
I think a minimal example should only have `mem_gb`, and timeout_min set.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
Queuecumber: > The config should contain what we need and nothing more.

The config it still correct though right? It's just that hydra marks optional parameters with `None` if they aren't used and submitit wants unused optionals to not be passed and this code reconciles the difference

> we can also fix this on AutoExecutor side

Do you want to though? I think the contract with users right now is very clear that optional parameters are not passed rather than assigned `None`

website/docs/plugins/submitit_launcher.md
Queuecumber: This isn't an example and they don't have to explicitly type `null` this is just showing what happens if you dump the config node in order to discover the possible values and their defaults

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
omry: In all other places in Hydra None is a legitimate value that can be used in configs (null in yaml).
The philosophy of Hydra that the config should be exactly what is needed by the app (to the possible extent).
Manipulating the config in code is not as flexible as composing the right thing in the first place.

I don't mind making None mean something special in this plugin but I expect it will confuse users of the plugin.
If we are planning on making a change in AutoExecutor anyway I would just wait for it. You are of course free to use your local hacked version of the plugin in the mean time.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
omry: > The config it still correct though right? It's just that hydra marks optional parameters with None if they aren't used and submitit wants unused optionals to not be passed and this code reconciles the difference

If the config contains None when it shouldn't it's not a correct config.
We either make None a legitimate input  or remove it from the config (probably by moving the field to the appropriate subclass).

> Do you want to though? I think the contract with users right now is very clear that optional parameters are not passed rather than assigned None

My opinion is that that we should be using the SlurmExecutor directly but I am yielding to the submitit team here.
What to do or not to do in AutoExecutor is their call.

Queuecumber: > If the config contains None when it shouldn't it's not a correct config.
We either make None a legitimate input or remove it from the config (probably by moving the field to the appropriate subclass).

Right so either we make an exception for this plugin because `None` means something different for submitit, or we fix on the autoexecutor side, I'm happy to push on both fronts

> My opinion is that that we should be using the SlurmExecutor directly but I am yielding to the submitit team here.
What to do or not to do in AutoExecutor is their call.

Agreed, using the executor directly makes the most sense to me but deferring to them seems like the right call here.

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
Queuecumber: I'm not clear on how we would make it work otherwise

The way I'm looking at it is just that `None` in hydra `==` missing value in submitit, so it's just an API translation.

Queuecumber: > I don't mind making None mean something special in this plugin but I expect it will confuse users of the plugin.
If we are planning on making a change in AutoExecutor anyway I would just wait for it. You are of course free to use your local hacked version of the plugin in the mean time.

If we fix it in AutoExecutor, that's fine and I can rebase this PR to one that just adds in the new parameters and makes the old ones optional.

jrapin:
> One other thing that I'd like to be considered: what exactly is the function of the local submitit executor in hydra? Is it doing something different from running your script without multirun?

It does a couple of stuff more which are similar to slurm behaviors:
- spawns multiprocess (if tasks/node > 1) and sets up corresponding environment variables
- allows to test preemption/timeout handling

plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
Queuecumber: With facebookincubator/submitit#1607 merged I'll update this PR to remove the None filtering

Queuecumber: I think I can't make a push on that change until submitit makes a release with facebookincubator/submitit#1607 is that correct @omry ?

omry: Hard to say, but it's easier to just do it once they release an updated version.

Queuecumber:
OK I think this is good to go now

website/docs/plugins/submitit_launcher.md
```diff
@@ -35,23 +35,28 @@ Submitit actually implements 2 different launchers: `submitit_slurm` to run on a
 You can discover the SLURM Launcher parameters with:
```
omry, line 35: can you fold this block like in the Ray Launcher docs?
see https://hydra.cc/docs/plugins/ray_launcher

Queuecumber: Done, somehow it turned out to be more complicated than I anticipated due to some formatting quirks of react

jrapin (approved):

Queuecumber:
Sorry about the test errors from before, it needed a rebase

plugins/hydra_submitit_launcher/news/1366.feature
```diff
@@ -0,0 +1 @@
+Add support for advanced GPU scheduling parameters
```
omry, line 1: can you be more specific? maybe:
```suggestion
Add support for SLURM parameters `cpus_per_gpu`, `gpus_per_task`, `mem_per_gpu` and `mem_per_cpu`
```

omry:
updated the doc a bit and commented on the news fragment.
FYI: this is in the master branch and the public version should be released from the 1.0_branch.

If you are interested in releasing a version everyone can consume, cherry pick this change to the 1.0_branch in a different PR first.
If you want to release it from master (for use with Hydra 1.1 dev releases) we can make a launcher dev release from master.

Queuecumber:
I would actually recommend keeping it in 1.1, but im interested in your thoughts on this. It might be a substantial enough change to the configuration that if people are relying on sane defaults it will break their config

omry:
I am fine with not releasing this version until we have a better reason to release it.
If you are okay with installing from master we can keep it at that until we release a formal version (probably with Hydra 1.1 if nothing urgent comes along).

Queuecumber:
well I wouldn't mind a dev release but it's probably not worth doing on just for this small feature

Queuecumber:
whats your timeline on 1.1 BTW?

omry:
no deadline, but I am estimating a release candidate in 2-3 months.

Queuecumber:
Ok that makes sense, I'm loving the new features, I even got my wife hooked on it

omry:
Thanks for the PR @Queuecumber.

Queuecumber:
Happy to help

Issue #1386, opened 2021-02-07
[Bug][Ax-plugin]Plugin raises SearchSpaceExhausted Exception when the search space is exhausted.


shagunsodhani:
# 🐛 Bug
## Description

Same as title.

## Checklist
- [ y] I checked on the latest version of Hydra
- [ y] I created a minimal repro (See [this](https://stackoverflow.com/help/minimal-reproducible-example) for tips).

## To reproduce

```
cd hydra/plugins/hydra_ax_sweeper
python tests/apps/polynomial.py -m polynomial.y=1 'polynomial.x=range(2,4)'
```

** Stack trace/error message **
```
[2021-02-07 09:03:38,001][HYDRA] AxSweeper is optimizing the following parameters: 
polynomial.x: choice=[2, 3]
polynomial.y: fixed=1
[INFO 02-07 09:03:38] ax.service.utils.instantiation: Inferred value type of ParameterType.INT for parameter polynomial.x. If that is not the expected value type, you can explicity specify 'value_type' ('int', 'float', 'bool' or 'str') in parameter dict.
[INFO 02-07 09:03:38] ax.modelbridge.dispatch_utils: Using Sobol generation strategy.
Traceback (most recent call last):
  File "/private/home/sodhani/projects/hydra/hydra/_internal/utils.py", line 212, in run_and_report
    return func()
  File "/private/home/sodhani/projects/hydra/hydra/_internal/utils.py", line 373, in <lambda>
    lambda: hydra.multirun(
  File "/private/home/sodhani/projects/hydra/hydra/_internal/hydra.py", line 136, in multirun
    return sweeper.sweep(arguments=task_overrides)
  File "/private/home/sodhani/projects/hydra/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/ax_sweeper.py", line 29, in sweep
    return self.sweeper.sweep(arguments)
  File "/private/home/sodhani/projects/hydra/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py", line 142, in sweep
    batch_of_trials = get_one_batch_of_trials(
  File "/private/home/sodhani/projects/hydra/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py", line 79, in get_one_batch_of_trials
    parameters, trial_index = ax_client.get_next_trial()
[...]

PR #1387, opened 2021-02-07
Correctly handle SearchSpaceExhaustedException from Ax

shagunsodhani:
## Motivation

Ax-Plugin raises SearchSpaceExhausted Exception when the search space is exhausted.

Yes

## Test Plan

Added test.

## Related Issues and PRs

#1386

### DIFF
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
@@ -27,7 +27,10 @@ class Trial:
     trial_index: int
 
 
-BatchOfTrialType = List[Trial]
+@dataclass
+class TrialBatch:
+    list_of_trials: List[Trial]
+    is_search_space_exhausted: bool
 
 
 def encoder_parameters_into_string(parameters: List[Dict[str, Any]]) -> str:
@@ -63,15 +66,13 @@ def get_one_batch_of_trials(
     parallelism: Tuple[int, int],
     num_trials_so_far: int,
     num_max_trials_to_do: int,
-) -> Tuple[BatchOfTrialType, bool, str]:
-    """Produce a batch of trials that can be run in parallel. It also
-    checks if the search space is exhausted. In case it is exhausted,
-    an exception message from Ax is also returned."""
+) -> TrialBatch:
+    """Returns a TrialBatch that contains a list of trials that can be
+    run in parallel. TrialBatch also flags if the search space is exhausted."""
 
     is_search_space_exhausted = False
     # Ax throws an exception if the search space is exhausted. We catch
     # the exception and set the flag to True
-    exception_msg = ""
     (num_trials, max_parallelism_setting) = parallelism
     if max_parallelism_setting == -1:
         # Special case, we can group all the trials into one batch
@@ -82,22 +83,24 @@ def get_one_batch_of_trials(
             # Given that num_trials is also -1, we can run all the trials in parallel.
             max_parallelism_setting = num_max_trials_to_do
 
-    batch_of_trials = []
+    list_of_trials = []
     for _ in range(max_parallelism_setting):
         try:
             parameters, trial_index = ax_client.get_next_trial()
-            batch_of_trials.append(
+            list_of_trials.append(
                 Trial(
                     overrides=map_params_to_arg_list(params=parameters),
                     trial_index=trial_index,
                 )
             )
-        except SearchSpaceExhausted as exception:
+        except SearchSpaceExhausted:
             is_search_space_exhausted = True
-            exception_msg = str(exception)
             break
 
-    return (batch_of_trials, is_search_space_exhausted, exception_msg)
+    return TrialBatch(
+        list_of_trials=list_of_trials,
+        is_search_space_exhausted=is_search_space_exhausted,
+    )
 
 
 class CoreAxSweeper(Sweeper):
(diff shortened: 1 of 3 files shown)


plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -76,14 +84,20 @@ def get_one_batch_of_trials(
...
+        except SearchSpaceExhausted as exception:
+            is_search_space_exhausted = True
+            exception_msg = str(exception)
+            break
+        else:
+            batch_of_trials.append(
+                Trial(
+                    overrides=map_params_to_arg_list(params=parameters),
+                    trial_index=trial_index,
+                )
```
omry, line 98: I never liked try, except else.
why not something like this:
```suggestion
        try:
            parameters, trial_index = ax_client.get_next_trial()
            batch_of_trials.append(
                Trial(
                    overrides=map_params_to_arg_list(params=parameters),
                    trial_index=trial_index,
                )            
        except SearchSpaceExhausted as exception:
            is_search_space_exhausted = True
            exception_msg = str(exception)
            break
```
?

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -167,6 +189,13 @@ def sweep(self, arguments: List[str]) -> None:
                     num_trials_left = -1
                     break
 
+                if is_search_space_exhausted:
+                    log.info(
+                        "Ax has exhausted the search space and raised "
+                        f"the exception: {search_space_exhausted_exception_msg}"
+                    )
```
omry, line 196: If this is business as usual, why even tell the user about it?

shagunsodhani: It may not be apparent to the end user that the search terminate early because the search space was exhausted.

omry: Ok.
Saying that this is an exception seems misleading still.
I understand that this is the underlying API, but the user message can just say that the search space is exhausted.
if the message from Ax is useful maybe we can use it directly.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -76,14 +84,20 @@ def get_one_batch_of_trials(
 
     batch_of_trials = []
     for _ in range(max_parallelism_setting):
-        parameters, trial_index = ax_client.get_next_trial()
-        batch_of_trials.append(
-            Trial(
-                overrides=map_params_to_arg_list(params=parameters),
-                trial_index=trial_index,
+        try:
```
omry, line 87: Since the exception is terminating the search, why not let it fly to the calling function and handle it there?

shagunsodhani: we would loose all the batches that we generated in this function (before the exception was raised).  For example, say the search space exhausted after 5 trials and num_max_trials_to_do = 10. So directly raising the exception means we do not evaluate the first 5 trials.

omry:
@bernardbeckerman, can you help reviewing this?

bernardbeckerman:
Looks good to me. I'm relatively new to the Ax code base, and would like to run this by one of my colleagues quickly next week, if this can wait that long. I feel like if Ax exhausts the search space, it should pat itself on the back and say "job well done! here are your best params." and not necessarily produce an error. There's probably a good reason why that's not the case, I'd just like to have full context.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
shagunsodhani: Hmm good point. I will wait for the follow up to https://github.com/facebookresearch/hydra/pull/1387#issuecomment-778514608 and if Ax will still raise the exception, I will update the message here.

omry:
> Looks good to me. I'm relatively new to the Ax code base, and would like to run this by one of my colleagues quickly next week, if this can wait that long. I feel like if Ax exhausts the search space, it should pat itself on the back and say "job well done! here are your best params." and not necessarily produce an error. There's probably a good reason why that's not the case, I'd just like to have full context.

Yes, this can wait until we can some clarity from the Ax team.
I agree with your intuition.

bernardbeckerman:
Just spoke with Lena offline and this makes sense to me now. Feel free to merge when you are ready.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
omry: I see, I missed the sweep in between because it was not showing in the diff.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -62,8 +63,15 @@ def get_one_batch_of_trials(
     parallelism: Tuple[int, int],
     num_trials_so_far: int,
     num_max_trials_to_do: int,
-) -> BatchOfTrialType:
-    """Produce a batch of trials that can be run in parallel"""
+) -> Tuple[BatchOfTrialType, bool, str]:
```
omry, line 66: use a small dataclass for the return value.
returning a complex tuple makes things more cumbersome in my opinion. (especially with your tendency for long variable names :) ).

omry (approved):
lgtm

omry (commented):
address my comment before merging.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
shagunsodhani: Done!

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -131,34 +148,40 @@ def sweep(self, arguments: List[str]) -> None:
         max_parallelism = ax_client.get_max_parallelism()
         current_parallelism_index = 0
         # Index to track the parallelism value we are using right now.
+        is_search_space_exhausted = False
+        # Ax throws an exception if the search space is exhausted. We catch
+        # the exception and set the flag to True
+
```
omry, line 154: generally speaking:
prefer putting the comment above the variable.

PR #1395, opened 2021-02-12
[Ax] implements `is_noisy` parameter for errorless measurements

### LINKED ISSUE #1352: [AxPlugin-Bug] Ax optimization does not return the optimal value
ax optimization doesnt actually select the lowest target value for recommendation

i run a job locally with the following ax config ("ax_local.yaml"):

```
hydra:
  sweeper:
    max_batch_size: 1
    ax_config:
      max_trials: 64
      params:
        p1:
          value_type: float
          type: range
          bounds: [0.0,8.0]
        p2:
          value_type: float
          type: range
          bounds: [0.0,5.0]
```

the job is launched like so:

```
python my_script.py -m --config-dir ~/configs --config-name my_config hydra/sweeper=ax +sweep_config=ax_local
```

my main method returns a value (which is what is presumably optimized) and also prints this score (along with params)

after the optimization is finished i get "best parameters" method. If i cross reference the parameters to the actual scores printed out during the optimization, i see that these there are several scores that are much lower that are available.

additionally, when "new best params" message is printed out occasionally, the score that it lists is not the actual score that was returned by the main method (but something reasonably close), for example:

New best value: 8.640574264487782, best parameters: ...

but when I find my own print out for these parameters i see that the score is actually 11.79782511729088

bernardbeckerman:
Closes #1352

## Motivation

Generally, Ax expects the function being optimized to return a tuple of `(measurement_value, measurement_uncertainty)`. If the return value is instead a scalar, Ax by default treats the uncertainty as an unknown and infers its value as the experiment progresses. For exact functions, this is not the desired behavior, so users may now leverage the `is_errorless` parameter. When `is_noisy` is `False`, the uncertainty of each scalar measurement is set to zero, whereas `True` (the default) leads Ax to infer the uncertainty as discussed above. Note that returning a `(measurement_value, measurement_uncertainty)` tuple will override this behavior in either circumstance.

Yes

## Test Plan

Tested by logging that hydra follows the correct codepath when run with `is_errorless: true/false`. Currently, all new code is covered by previous unit tests since the `banana` example has `is_errorless = True` and all others follow the default which is false.

## Related Issues and PRs

https://github.com/facebookresearch/hydra/issues/1352

### DIFF
--- a/plugins/hydra_ax_sweeper/example/conf/config.yaml
+++ b/plugins/hydra_ax_sweeper/example/conf/config.yaml
@@ -26,6 +26,10 @@ hydra:
         # An Epoch is defined as a batch of trials executed in parallel
         max_epochs_without_improvement: 10
 
+      # the synthetic measurements in this example have 0 error
+      # which is specified by setting is_errorless to true
+      is_errorless: true
+
       params:
         banana.x:
           type: range
--- a/plugins/hydra_ax_sweeper/example/conf/config.yaml
+++ b/plugins/hydra_ax_sweeper/example/conf/config.yaml
@@ -27,8 +27,8 @@ hydra:
         max_epochs_without_improvement: 10
 
       # the synthetic measurements in this example have 0 error
-      # which is specified by setting is_errorless to true
-      is_errorless: true
+      # which is specified by setting is_noisy to false
+      is_noisy: False
 
       params:
         banana.x:
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
@@ -227,10 +227,10 @@ class CoreAxSweeper(Sweeper):
                 # if true (default), the error of each measurement is inferred by Ax.
                 # if false, the error of each measurement is set to 0.
                 if isinstance(val, (int, float)):
-                    if not self.is_noisy:
-                        val = (val, 0)
+                    if self.is_noisy:
+                        val = (val, None) # specify unknown noise
                     else:
-                        val = (val, None)
+                        val = (val, 0) # specify no noise
                 ax_client.complete_trial(
                     trial_index=batch[idx].trial_index, raw_data=val
                 )
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
@@ -126,6 +126,7 @@ class CoreAxSweeper(Sweeper):
         self.sweep_dir: str
         self.job_idx: Optional[int] = None
         self.max_batch_size = max_batch_size
+        self.is_errorless: bool = ax_config.is_errorless
 
     def setup(
         self,
@@ -218,7 +219,18 @@ class CoreAxSweeper(Sweeper):
             )
             self.job_idx += len(rets)
             for idx in range(len(batch)):
-                val = rets[idx].return_value
+                val: Any = rets[idx].return_value
+                # Ax expects a measured value (int or float), which can optionally
+                # be given in a tuple along with the error of that measurement
+                assert isinstance(val, (int, float, tuple))
+                # is_errorless specifies how Ax should behave when not given an error value.
+                # if false (default), the error of each measurement is inferred by Ax.
+                # if true, the error of each measurement is set to 0.
+                if isinstance(val, (int, float)):
+                    if self.is_errorless:
+                        val = (val, 0)
+                    else:
+                        val = (val, None)
                 ax_client.complete_trial(
                     trial_index=batch[idx].trial_index, raw_data=val
                 )
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
@@ -228,9 +228,9 @@ class CoreAxSweeper(Sweeper):
                 # if false, the error of each measurement is set to 0.
                 if isinstance(val, (int, float)):
                     if self.is_noisy:
-                        val = (val, None) # specify unknown noise
+                        val = (val, None)  # specify unknown noise
                     else:
-                        val = (val, 0) # specify no noise
+                        val = (val, 0)  # specify no noise
                 ax_client.complete_trial(
                     trial_index=batch[idx].trial_index, raw_data=val
                 )
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
@@ -126,7 +126,7 @@ class CoreAxSweeper(Sweeper):
         self.sweep_dir: str
         self.job_idx: Optional[int] = None
         self.max_batch_size = max_batch_size
-        self.is_errorless: bool = ax_config.is_errorless
+        self.is_noisy: bool = ax_config.is_noisy
 
     def setup(
         self,
@@ -223,11 +223,11 @@ class CoreAxSweeper(Sweeper):
                 # Ax expects a measured value (int or float), which can optionally
                 # be given in a tuple along with the error of that measurement
                 assert isinstance(val, (int, float, tuple))
-                # is_errorless specifies how Ax should behave when not given an error value.
-                # if false (default), the error of each measurement is inferred by Ax.
-                # if true, the error of each measurement is set to 0.
+                # is_noisy specifies how Ax should behave when not given an error value.
+                # if true (default), the error of each measurement is inferred by Ax.
+                # if false, the error of each measurement is set to 0.
                 if isinstance(val, (int, float)):
-                    if self.is_errorless:
+                    if not self.is_noisy:
                         val = (val, 0)
                     else:
                         val = (val, None)
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
@@ -50,7 +50,7 @@ class AxConfig:
     experiment: ExperimentConfig = ExperimentConfig()
     client: ClientConfig = ClientConfig()
     params: Dict[str, Any] = field(default_factory=dict)
-    is_errorless: bool = False
+    is_noisy: bool = True
 
 
 @dataclass
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
@@ -50,6 +50,7 @@ class AxConfig:
     experiment: ExperimentConfig = ExperimentConfig()
     client: ClientConfig = ClientConfig()
     params: Dict[str, Any] = field(default_factory=dict)
+    is_errorless: bool = False
 
 
 @dataclass
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
@@ -50,6 +50,8 @@ class AxConfig:
     experiment: ExperimentConfig = ExperimentConfig()
     client: ClientConfig = ClientConfig()
     params: Dict[str, Any] = field(default_factory=dict)
+    # is_noisy = True indicates measurements have unknown uncertainty
+    # is_noisy = False indicates measurements have an uncertainty of zero
     is_noisy: bool = True
 
 
--- a/plugins/hydra_ax_sweeper/news/1395.feature
+++ b/plugins/hydra_ax_sweeper/news/1395.feature
@@ -1,1 +1,1 @@
-Support for errorless measurements in an Ax experiment via `is_noisy` input parameter
+Support for deterministic functions in an Ax experiment via `is_noisy` input parameter
--- a/plugins/hydra_ax_sweeper/news/1395.feature
+++ b/plugins/hydra_ax_sweeper/news/1395.feature
@@ -1,0 +1,1 @@
+Support for errorless measurements in an Ax experiment via `is_errorless` input parameter
--- a/plugins/hydra_ax_sweeper/news/1395.feature
+++ b/plugins/hydra_ax_sweeper/news/1395.feature
@@ -1,1 +1,1 @@
-Support for errorless measurements in an Ax experiment via `is_errorless` input parameter
+Support for errorless measurements in an Ax experiment via `is_noisy` input parameter
--- a/website/docs/plugins/ax_sweeper.md
+++ b/website/docs/plugins/ax_sweeper.md
@@ -35,7 +35,7 @@ The return value of the function should be the value that we want to optimize.
 To compute the best parameters for the Banana function, clone the code and run the following command in the `plugins/hydra_ax_sweeper` directory:
 
 ```
-python example/banana.py -m 'banana.x=int(interval(-5, 5))' 'banana.y=interval(-5, 10.1)' hydra.sweeper.ax_config.is_noisy=false
+python example/banana.py -m 'banana.x=int(interval(-5, 5))' 'banana.y=interval(-5, 10.1)'
 ```
 
 The output of a run looks like:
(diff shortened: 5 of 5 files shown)


bernardbeckerman:
If a test specific to this change is desired, I may need some coaching, since I am somewhat new to pytest and am unsure how to test this without superfluous log statements and/or testing for specific output/"best parameter" values.

ldworkin:
> If a test specific to this change is desired, I may need some coaching, since I am somewhat new to pytest and am unsure how to test this without superfluous log statements and/or testing for specific output/"best parameter" values.

Yea, agreed, this is tricky. A log statement is one way to go (combined with something like [this](https://github.com/facebookresearch/hydra/blob/master/plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py#L213)) but I'm not sure its ideal. And it seems really difficult to somehow create a synthetic problem with behavior like the one we saw in the motivation issue.

website/docs/plugins/ax_sweeper.md
```diff
@@ -42,22 +42,22 @@ The output of a run looks like:
...
+[__main__][INFO] - Banana_Function(x=2, y=-0.9878678928129379)=2488.8826116154178
+[HYDRA]        #1 : banana.x=-1 banana.y=7.700736676901579
+[__main__][INFO] - Banana_Function(x=-1, y=7.700736676901579)=4493.987201317402
+[HYDRA]        #2 : banana.x=-1 banana.y=-3.901284400280565
+[__main__][INFO] - Banana_Function(x=-1, y=-3.901284400280565)=2406.2588772433614
+[HYDRA]        #3 : banana.x=-1 banana.y=0.20855206744745391
+[__main__][INFO] - Banana_Function(x=-1, y=0.20855206744745391)=66.63898299416996
+[HYDRA]        #4 : banana.x=4 banana.y=-4.557481761090457
+[__main__][INFO] - Banana_Function(x=4, y=-4.557481761090457)=42270.00563575668
+[HYDRA] New best value: 66.63898299416996, best parameters: {'banana.x': -1, 'banana.y': 0.20855206744745391}
```
shagunsodhani, line 60: curious why did this output change so much?

shagunsodhani: additionally, can we truncate the values to 3 places of decimal?

website/docs/plugins/ax_sweeper.md
```diff
@@ -81,10 +81,12 @@ In general, the plugin supports setting all the Ax supported [Parameters](https:
...
 * `values` - Required only for the `choice` parameters. It should be a list of values.
-* `value` - Required only for the `fixed` parameters. It should be a single value. 
+* `value` - Required only for the `fixed` parameters. It should be a single value.
 
 Note that if you want to sample integers in the range `-5` to `5`, you need to specify the range as `int(interval(-5, 5))` (in the command line) or `[-5, 5]` (in config). If you want to sample floats in range `-5` to `5`, you need to specify the range as `interval(-5, 5)` (in the command line) or `[-5.0, 5.0]` (in config).
 
+Generally, Ax expects the function being optimized to return a tuple of `(measurement_value, measurement_uncertainty)`. If the return value is instead a scalar, Ax by default treats the uncertainty as an unknown and infers its value as the experiment progresses. The example above uses an exact function, so the `is_exact` parameter is set to `true`. As a result, the uncertainty of each scalar measurement is set to 0. Note that returning a `(measurement_value, measurement_uncertainty)` tuple will override this behavior.
```
shagunsodhani, line 88: ``is_exact` parameter is set to `true`` -> is this referencing the change in the example config? probably better to override and show in the docs (eg here `python example/banana.py -m 'banana.x=int(interval(-5, 5))' 'banana.y=interval(-5, 10.1)'`)

website/docs/plugins/ax_sweeper.md
```diff
@@ -107,5 +109,6 @@ ax_config:
   client:
     verbose_logging: false
     random_seed: null
+  is_exact: false
```
shagunsodhani, line 112: Probably not a big deal but being exact is not the same as being certain. Ax cares about the uncertainty measure so maybe we could call it `is_certain` (or something else).

shagunsodhani (commented):
Thanks for the PR :) 

We would also want to add a news entry for this change as described [here](https://hydra.cc/docs/development/documentation#news-entries)

This will also be a good opportunity to release a new version of the plugin.

shagunsodhani:
> > If a test specific to this change is desired, I may need some coaching, since I am somewhat new to pytest and am unsure how to test this without superfluous log statements and/or testing for specific output/"best parameter" values.
> 
> Yea, agreed, this is tricky. A log statement is one way to go (combined with something like [this](https://github.com/facebookresearch/hydra/blob/master/plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py#L213)) but I'm not sure its ideal. And it seems really difficult to somehow create a synthetic problem with behavior like the one we saw in the motivation issue.

How about something like [this](https://github.com/facebookresearch/hydra/blob/master/plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py#L90) where we run two versions, with and without the flag and ensure that the values matches with the respective expected values. If the number of trials are small, we should not have similar values.  There was a discussion about reproducibility of results with the Ax team. I will cc you on that discussion.

website/docs/plugins/ax_sweeper.md
bernardbeckerman: good point. maybe `is_errorless`?

website/docs/plugins/ax_sweeper.md
bernardbeckerman: > curious why did this output change so much?

Trials is the correct term here rather than arms. Other than that, it just looks like some whitespace and an increase in sig figs, and I'm not sure why that changed. As for the actual numerical values changing, I suppose that could be due to randomness in each run, as well as the change implemented here, namely, setting SEM to 0.

> additionally, can we truncate the values to 3 places of decimal?

I could be wrong but this sounds like something that needs to happen in the Ax codebase, which I think should be decoupled from this PR and done later. I can truncate manually if you like in the mean time.

shagunsodhani: Regarding truncating, I meant manually truncating in the docs (not the code etc). Sorry I should have clarified it earlier.

website/docs/plugins/ax_sweeper.md
shagunsodhani: Is there a term that Ax docs/codebase uses that we can reuse?

bernardbeckerman: good point! changed to `is_errorless`.

bernardbeckerman: The parameter that is being set to 0 is called `SEM` which stands for "standard error of the mean" which can be truncated to "error" in this context, which is why I believe `is_errorless` fits.

bernardbeckerman (commented):
> How about something like this where we run two versions, with and without the flag and ensure that the values matches with the respective expected values. If the number of trials are small, we should not have similar values. There was a discussion about reproducibility of results with the Ax team. I will cc you on that discussion.

Just to make sure I understand, are you saying run a test to ensure that the output is different when the `is_errorless` flag is toggled? If I've got this wrong, feel free to skip the remainder of this comment. If I'm understanding correctly, I think this could produce problems/probabilistically failing tests since consecutive runs with the same inputs may not produce the same output.

plugins/hydra_ax_sweeper/example/conf/config.yaml
```diff
@@ -26,6 +26,8 @@ hydra:
         # An Epoch is defined as a batch of trials executed in parallel
         max_epochs_without_improvement: 10
 
+      is_errorless: true
+
```
omry, line 30: Can you add a comment here explaining what this does? (this is an example people will look at).

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -191,7 +192,13 @@ def sweep_over_batches(
             )
             self.job_idx += len(rets)
             for idx in range(len(batch)):
-                val = rets[idx].return_value
+                val: Any = rets[idx].return_value
+                assert isinstance(val, (int, float, tuple))
```
omry, line 196: Add a comment to this assert. I think it will trip some people.
something like:
```python
assert isinstance(val, (int, float, tuple)), f"Return value type {type(val).__name__} is not supported by Ax"
```

omry (commented):
See inline comments.
Also:
It would be good to undo the whitespace changes (it's probably your IDE stripping whitespace from end of lines), but I don't care too much about that.

About testing:

I guess the behavior we are really interested in is to see that the reported best result is exactly matching the actual best result when is_errorless is set to True.
this seems testable.

bernardbeckerman:
@omry @shagunsodhani I understand you want a guarantee that Ax will return the best parameter set if `is_errorless = True`, especially after seeing that this wasn't the case before this PR. I agree this would be a useful test, and I think it would be better to implement it within Ax itself, which is where that contract should be maintained, rather than in hydra. We've filed this on our task board (T84973028 - apologies to anyone who can't follow this facebook task 😅) to be addressed in a future sprint. As for guaranteeing this behavior in the mean time, it should be noted that this desired behavior is and has always been the behavior of Ax.

website/docs/plugins/ax_sweeper.md
```diff
@@ -1,4 +1,4 @@
----
+OB---
```
omry, line 1: Probably unintentional?

website/docs/plugins/ax_sweeper.md
```diff
@@ -35,29 +35,29 @@ The return value of the function should be the value that we want to optimize.
 To compute the best parameters for the Banana function, clone the code and run the following command in the `plugins/hydra_ax_sweeper` directory:
 
 ```
-python example/banana.py -m 'banana.x=int(interval(-5, 5))' 'banana.y=interval(-5, 10.1)'
+python example/banana.py -m 'banana.x=int(interval(-5, 5))' 'banana.y=interval(-5, 10.1)' 'hydra.sweeper.ax_config.is_errorless=True'
```
omry, line 38: 1. no need to quote regular overrides (you need to quote the functions because the shell will not like ().
2. use lowercase true.

omry (changes requested):
Thanks @bernardbeckerman, Your comment about testing it in Ax makes sense to me.
See a couple of nits. otherwise this is good to go.

omry:
Looks like there are some conflicts. Can you fix?

omry:
Looks like you ran into some issues with the conflicts. Sent you a direct message.

shagunsodhani:
I probably would have preferred `is_noisy` over `is_errorless` but `is_errorless` is fine as well.

omry:
I like is_noisy as well.

bernardbeckerman:
ah shoot looks like i undid some more recent changes with my effort to resolve this. will have to go through and edit manually tomorrow.

website/docs/plugins/ax_sweeper.md
```diff
@@ -85,6 +85,8 @@ In general, the plugin supports setting all the Ax supported [Parameters](https:
 
 Note that if you want to sample integers in the range `-5` to `5`, you need to specify the range as `int(interval(-5, 5))` (in the command line) or `[-5, 5]` (in config). If you want to sample floats in range `-5` to `5`, you need to specify the range as `interval(-5, 5)` (in the command line) or `[-5.0, 5.0]` (in config).
 
+Generally, Ax expects the function being optimized to return a tuple of `(measurement_value, measurement_uncertainty)`. If the return value is instead a scalar, Ax by default treats the uncertainty as an unknown and infers its value as the experiment progresses. The example above uses an exact function, so the `is_errorless` parameter is set to `True`. As a result, the uncertainty of each scalar measurement is set to 0. Note that returning a `(measurement_value, measurement_uncertainty)` tuple will override this behavior.
+
```
omry, line 89: Maybe something like:
```
The Ax Sweeper assumes the optimized function is a noisy function with unknown measurement uncertainty.
This can be changed by overriding the `is_errorless` parameter to True, which is equivalent to a unction that is always certain about it's measurements.

Functions that can provide actual  measurement uncertainty can do so by returning the tuple `(measurement_value, measurement_uncertainty)` instead of a scalar value.
```

I also think we need to do a better job at describing what the measurement_uncertainty actually is and how to compute it. Is this [SEM (Standard deviation)](https://en.wikipedia.org/wiki/Standard_error)?

omry (commented):
1. Thoughts on is_noisy?
2. What did you drop? looks pretty good to me.

website/docs/plugins/ax_sweeper.md
bernardbeckerman: Yes, SEM is as described in that article. Also I like your rephrasing.

bernardbeckerman:
1. I think I like `is_noisy` better - will update.
2. I think I had fixed the issue by the time you looked at the PR.

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -218,7 +219,18 @@ def sweep_over_batches(
...
+                # be given in a tuple along with the error of that measurement
+                assert isinstance(val, (int, float, tuple))
+                # is_noisy specifies how Ax should behave when not given an error value.
+                # if true (default), the error of each measurement is inferred by Ax.
+                # if false, the error of each measurement is set to 0.
+                if isinstance(val, (int, float)):
+                    if not self.is_noisy:
+                        val = (val, 0)
+                    else:
+                        val = (val, None)
```
omry, line 233: This seems a bit clearer:
```suggestion
                    if self.is_noisy:
                        val = (val, None) # specify unknown noise
                    else:
                        val = (val, 0) # specify no noise
```

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/config.py
```diff
@@ -50,6 +50,7 @@ class AxConfig:
     experiment: ExperimentConfig = ExperimentConfig()
     client: ClientConfig = ClientConfig()
     params: Dict[str, Any] = field(default_factory=dict)
+    is_noisy: bool = True
```
omry, line 53: Can you add a comment here?
maybe with a link to the Ax docs if there is anything good to link to?

plugins/hydra_ax_sweeper/news/1395.feature
```diff
@@ -0,0 +1 @@
+Support for errorless measurements in an Ax experiment via `is_noisy` input parameter
```
omry, line 1:
```suggestion
Support for deterministic functions in an Ax experiment via `is_noisy` input parameter
```

website/docs/plugins/ax_sweeper.md
```diff
@@ -28,36 +28,36 @@ defaults:
...
-implements the [Rosenbrock function (aka Banana function)](https://en.wikipedia.org/wiki/Rosenbrock_function). 
+implements the [Rosenbrock function (aka Banana function)](https://en.wikipedia.org/wiki/Rosenbrock_function).
 The return value of the function should be the value that we want to optimize.
 
 
 To compute the best parameters for the Banana function, clone the code and run the following command in the `plugins/hydra_ax_sweeper` directory:
 
 ```
-python example/banana.py -m 'banana.x=int(interval(-5, 5))' 'banana.y=interval(-5, 10.1)'
+python example/banana.py -m 'banana.x=int(interval(-5, 5))' 'banana.y=interval(-5, 10.1)' hydra.sweeper.ax_config.is_noisy=false
```
omry, line 38: The default for this example is is_noisy=false, no need to override it here.

omry (commented):
looking good, a few nits.

website/docs/plugins/ax_sweeper.md
bernardbeckerman: replicated here for clarity per [this comment](https://github.com/facebookresearch/hydra/pull/1395#discussion_r575588955) from @shagunsodhani

omry: Either way works. I just think it's not adding clarity to the example.
@shagunsodhani, what do you think?

shagunsodhani: Ahh I suggested it earlier because we were/are setting the default of `is_noisy` to `True`. Given that we are overriding it in the config, I think we can remove it.

bernardbeckerman: But that was always the case. @shagunsodhani your comment:

> probably better to override and show in the docs

suggests that it's better that the user sees this option explicitly used in the docs.

bernardbeckerman: I think we should just ship this.

omry (changes requested):
> I think we should just ship this.

Can you respond to my suggestions first? (in  other files too).
Nothing major, but I think you missed them.

bernardbeckerman:
@omry I definitely did! Thanks for pointing it out.

omry (approved):
Thanks @bernardbeckerman!

PR #1378, opened 2021-02-04
Upgrade docusaurus-plugin-internaldocs-fb

jknoxville:
## Motivation

Updating the facebook internaldocs plugin.
I noticed hydra is on a really old version, because I always update the monorepo, but this one's source of truth is GitHub.

Things this will bring:
* No more certificate dialogs in safari for viewers of the open source site.
* Transclude internal markdown files inside external ones.
* Use code snippets to embed real code from the repo. [readme](https://www.npmjs.com/package/remark-code-snippets)
* And more...

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebookresearch/hydra/blob/master/CONTRIBUTING.md)?

Yes

## Test Plan

I have run:
```
yarn
yarn build
```
inside website/ and it appears to work fine.

Also
```
yarn start
```

as an fb employee and verified that the internal documentation banner still displays.

PR #1399, opened 2021-02-15
Use internaldocs-fb preset

jknoxville:
## Motivation

I saw @omry 's comment about code-snippets not working and remembered that you need to use this preset.

Enables the facebook docusaurus preset, which brings various plugins, including internal/external content helper functions and code-snippets.

The reason the preset replaces the docs preset, is that it has to add to the remark (markdown) plugins used by the content plugin.

You can do achieve this without using this preset if you want, by reproducing [this](https://www.internalfb.com/intern/diffusion/FBS/browsefile/master/xplat/staticdocs/packages/docusaurus-plugin-internaldocs-fb/docusaurus-preset.js?commit=d6f28e365058b4fac570fff87e97ba986d967516&lines=17-20) in your docusaurus.config.js, but this preset makes it easier. It also acts as a single point of integration for all facebook plugins, so you'll just need to bump this preset version to get all the new features.

### Have you read the [Contributing Guidelines on pull requests?

Yes

## Test Plan

Example of using the code-snippets plugin

## Related Issues and PRs

https://github.com/facebookresearch/hydra/pull/1378

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -19,7 +19,6 @@ module.exports = {
             current: "https://github.com/facebookresearch/hydra/blob/master/",
         },
     },    
-    plugins: [require.resolve('docusaurus-plugin-internaldocs-fb')],
     themeConfig: {
         googleAnalytics: {
             trackingID: 'UA-149862507-1',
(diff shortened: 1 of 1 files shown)


omry (approved):
Thanks John, we can give this one a shot and see how it goes.

PR #1440, opened 2021-03-02
Support for force-add override syntax (++key=value)

### LINKED ISSUE #1049: Add OR Override-if-exists
# 🚀 Feature Request

Add option to "Add-OR-Override-if-exists" to command line params

e.g.
```
python myscript.py !gpus=2 # gpus will be set to two, disregarding whether it existed in the original config file or not
```

## Motivation

Hydra offers few options to handle params from command line:
```
python myscript.py gpus=2 # override gpus (must exists in config files/structured config earlier!)
```
```
python myscript.py ~gpus # removes gpus from configuration 
```
```
python myscript.py +gpus=2 # add gpus (cannot exists in config files/structured config earlier!)
```

Note +gpus will raise an error if it was already present in the configuration.

So the case, when we want to have gpus set to 2, disregarding whether it was present in yaml file or now, is missing.

omry:
Closes #1049

### DIFF
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -68,40 +68,37 @@ class Overrides:
             is_dict = isinstance(override.value(), dict)
             if is_dict or not is_group:
                 self.config_overrides.append(override)
-            else:
-                if override.is_force_add():
-                    # This could probably be made to work if there is a compelling use case.
-                    raise ConfigCompositionException(
-                        f"force-add of config groups is not supported: '{override.input_line}'"
-                    )
-                elif override.is_delete():
-                    key = override.get_key_element()[1:]
-                    value = override.value()
-                    if value is not None and not isinstance(value, str):
-                        raise ValueError(
-                            f"Config group override deletion value must be a string : {override}"
-                        )
-
-                    self.deletions[key] = Deletion(name=value)
-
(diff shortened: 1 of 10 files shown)


hydra/_internal/defaults_list.py
```diff
@@ -69,7 +69,12 @@ def __init__(self, repo: IConfigRepository, overrides_list: List[Override]) -> N
             if is_dict or not is_group:
                 self.config_overrides.append(override)
             else:
-                if override.is_delete():
+                if override.is_force_add():
```
odelalleau, line 72: Not actually related to this PR, but you could gain some indentation back by using an`elif` here

omry: yup, done.

odelalleau (changes requested):
Just one thing, I suspect that `get_key_element()` in`types.py` also needs to be updated, could you have a look and add a test for it?

omry:
> Just one thing, I suspect that `get_key_element()` in`types.py` also needs to be updated, could you have a look and add a test for it?

nice catch.

odelalleau (approved):
All good once CI gives the green light!

kenahoo:
I tested this for my use case, it seems to work well.  Thanks for looking at it.

Gpapi13:
- [ ]

PR #1441, opened 2021-03-03
RFC - internal callbacks

jieru-hu:
- [Updated design doc](https://docs.google.com/document/d/1IXp0fN8qMz7N6R_FJ8-WiZgtgUgpcNbY2MXqNd2LTAU/edit)
- Will add and fix the unit tests once we agree on the design
- The example apps are added as an example on how the callbacks should be configured, I do not intend to check anything under the examples folder in (since this will remain Hydra internal for now.)


jieru-hu:
Mark this as a draft while we finalize the design.

jieru-hu:
For some reason, the default callbacks configurations are not discovered in python 3.9  (while working in <=3.8). 

Looking into the failures now...

hydra/conf/__init__.py
```diff
@@ -98,13 +98,18 @@ class HydraConf:
             {"hydra_help": "default"},
             {"hydra_logging": "default"},
             {"job_logging": "default"},
+            {"env": "default"},
+            {"callbacks": "default"},
```
omry, line 102: env should be last because it should be allowed to override any of the config groups defined above.

omry: Also, we can set it to None here and then we wouldn't need env/default.yaml.

hydra/conf/hydra/callbacks/default.yaml
```diff
@@ -0,0 +1,3 @@
+# @package hydra
+
+callbacks: {}
```
omry, line 3: This can work as an empty file as well, right?

hydra/conf/__init__.py
jieru-hu: > env should be last because it should be allowed to override any of the config groups defined above.

Yea agreed env should be the last compared to the other config groups.  But I somehow feel it should be before `callbacks`. If we put this before callbacks, we will get to call our default callbacks before custom ones. We can be sure of what configs we are getting, in case the custom callbacks modifies the GlobalHydra configs etc. 

> Also, we can set it to None here and then we wouldn't need env/default.yaml.

I tried to set this to none here, but as a result, I cannot prepend a `default.yaml` config via the SearchPathPlugin.

hydra/conf/hydra/callbacks/default.yaml
jieru-hu: Yes, it works. I've made the Yaml empty. :)

hydra/conf/__init__.py
omry: a specific env file can add some unconditional callbacks for this purpose.

hydra/env/fbcode.yam
```yaml
defaults:
  - hydra/output: fbcode
  - hydra/callbacks/fbcode # not overridable. this does not interact with the choice of the config group hydra/callbacks.
```

However, in some environment you may want to add some default callbacks that the user can override:
```yaml
defaults:
  - hydra/output: fbcode
  - hydra/callbacks: # overrideable
    - foo
    - bar
```

jieru-hu: I see, thanks, makes sense! I've fixed the order.

omry: The new Defaults List is super powerful. It takes a while to really grasp all the things it can do :).

hydra/_internal/callback.py
```diff
@@ -0,0 +1,54 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Callback:
+    def on_run_start(
+        self,
+        config: DictConfig,
+        **kwargs: Any,
+    ) -> None:
```
omry, line 17: 1. can this be more compact if you remove the trailing comma (in all of those).
2. Add docstrings describing each event.

hydra/core/callbacks.py
```diff
@@ -0,0 +1,112 @@
...
+
+from omegaconf import DictConfig
+
+from hydra._internal.callback import Callback
+from hydra.core.singleton import Singleton
+from hydra.core.utils import JobReturn
+from hydra.utils import instantiate
+
+
+class Callbacks(metaclass=Singleton):
```
omry, line 14: Why should this be a singleton?
Looking at how you are using it, I suspect there isn't really a reason for this. making it a normal member of the Hydra object should work just fine.

(You will need to pass it explicitly into run_job()).

hydra/core/callbacks.py
```diff
@@ -0,0 +1,112 @@
...
+        function_name: str,
+        **kwargs: Any,
+    ) -> None:
+        for c in self.callbacks:
+            try:
+                getattr(c, function_name)(**kwargs)
+            except Exception as e:
+                warnings.warn(
+                    f"Exception occurred during {function_name} with callback {c}, error message: {e}"
+                )
```
omry, line 44: What does the exception message looks like? in particular, the callback {c}?

hydra/core/callbacks.py
```diff
@@ -0,0 +1,112 @@
...
+    def instantiate_callbacks(self, config: DictConfig) -> None:
+        if config.hydra.callbacks is None:
+            raise RuntimeError("Hydra Callbacks is not configured")
+
+        assert len(self.callbacks) == 0, "Callbacks already instantiated."
+
+        for params in config.hydra.callbacks.values():
+            self.callbacks.append(instantiate(params))
+
+    def _iterate(
```
omry, line 33: notify or fireEvent sounds better.
```suggestion
    def _notify(
```

hydra/core/utils.py
```diff
@@ -126,11 +128,19 @@ def run_job(
...
+            )
             ret.return_value = task_function(task_cfg)
         ret.task_name = JobRuntime.instance().get("name")
 
         _flush_loggers()
 
+        Callbacks.instance().on_job_end(
+            config=config,
+            job_return=ret,
+        )
```
omry, line 142: No reason for this to spawn 4 lines, right? if you remove the trailing comma it would probably fit in one line.

omry (commented):
some feedback on impl (didn't look at tests yet).

jieru-hu:
@omry 
thanks for the review! Sorry I had to force push (the branch is a bit outdated), all the new changes start from 3e98ffd

hydra/core/callbacks.py
jieru-hu: I've considered passing it explicitly: If we pass in Callbacks to every run_job that means we will have pass it from sweepers then on to all the launchers. 
Meaning we will probably have to change all the sweeper and launcher code. 
for example, we'd have to change [here](https://github.com/facebookresearch/hydra/blob/37cdb2daf75c9a0af37952122405af8f83449772/plugins/hydra_joblib_launcher/hydra_plugins/hydra_joblib_launcher/_core.py#L43-L48) for the joblib launcher

Another alternative is to init callbacks every time we need then, but that doesn't seem as clean (we will have to init callbacks twice 
during the life of a Hydra application)

After trying out a few alternative ways, I think having this as a Singleton is the cleanest way to get around the above. 

What do you think?

hydra/core/callbacks.py
jieru-hu: fixed!

hydra/core/utils.py
jieru-hu: true, fixed this and other places as well.

hydra/core/callbacks.py
jieru-hu: I've fixed the warning message a bit, looks like the following now.
```
[2021-03-23 16:52:57,503][HYDRA] Init DefaultCallback with self.param1=foo, self.param2=bar
/Users/jieru/workspace/hydra-fork/hydra/hydra/core/callbacks.py:42: UserWarning: Exception occurred during on_run_start with callback DefaultCallback, error message: hello world
  warnings.warn(
```

hydra/core/callbacks.py
omry: I think we should it init it once, and pass it down to plugins somehow (e.g. Sweeper -> Launcher -> run_job). Using a singleton here is just accruing technical debt.

What do you think about the following:

Introduce a HydraContext object, it can contain ConfigLoader and the callbacks for now.
We now need to make sure HydraContext is propagated in.

This can either be a breaking change in plugins:
e.g. a change to the signature of setup(), in this case, we can release dev release of existing plugins for Hydra 1.1 users) (and also update the dependencies of Hydra 1.0 plugins to be sure they are marked as Hydra 1.0 only).

Alternatively we can try to make it in steps while maintaining backward compatibility.
In this case I would like to see a plan of how we do it (what happens at each Hydra version).

I am leaning toward just ripping the band-aid and go with option 1, but open to option 2 as well.

hydra/core/callbacks.py
```diff
@@ -0,0 +1,112 @@
...
+        ret = Singleton.instance(Callbacks, *args, **kwargs)
+        assert isinstance(ret, Callbacks)
+        return ret
+
+    def __init__(self) -> None:
+        self.callbacks: List[Callback] = []
+
+    def instantiate_callbacks(self, config: DictConfig) -> None:
+        if config.hydra.callbacks is None:
+            raise RuntimeError("Hydra Callbacks is not configured")
```
omry, line 26: HydraException?

hydra/_internal/callback.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+import logging
```
omry, line 2: I guess we can move this to hydra.experimental instead of hydra._internal.
we can followup with documentation later.

hydra/conf/__init__.py
```diff
@@ -98,13 +98,18 @@ class HydraConf:
             {"hydra_help": "default"},
             {"hydra_logging": "default"},
             {"job_logging": "default"},
+            {"callbacks": "default"},
```
omry, line 101: I think if you use None here you won't need an empty file:
```suggestion
            {"callbacks": None},
```

This is equivalent to using null in the defaults list in yaml.
(documented [here](https://hydra.cc/docs/next/advanced/defaults_list)).

hydra/conf/__init__.py
```diff
@@ -98,13 +98,18 @@ class HydraConf:
...
         ]
     )
 
     # Elements to append to the config search path.
     # Note: This can only be configured in the primary config.
     searchpath: List[str] = field(default_factory=list)
 
+    # Provides environment specific overrides.
+    env: Dict[str, Any] = field(default_factory=dict)
+
```
omry, line 112: What is this for?

tests/test_apps/app_with_callbacks/default_callback/hydra_plugins/env_plugin/env_searchpath_plugin.py
```diff
@@ -0,0 +1,12 @@
...
+from hydra.plugins.search_path_plugin import SearchPathPlugin
+
+
+class ExampleSearchPathPlugin(SearchPathPlugin):
+    def manipulate_search_path(self, search_path: ConfigSearchPath) -> None:
+        search_path.prepend(
+            provider="test-env-defaults",
+            path="pkg://hydra_plugins.env_plugin.conf",
+            anchor="hydra",
+        )
```
omry, line 12: What is an Env Plugin?

Let's pull out the env related stuff to a separate PR. It deserves proper review.

omry (changes requested):

jieru-hu:
Thanks for the review @omry 

> I guess we can move this to hydra.experimental instead of hydra._internal.
we can followup with documentation later.

I feel `_internal` is more fitting for Callback  in that Callbacks are more like Launcher/Sweepers and we do not expect users to call the event methods directly. Could you share more thought on why should fall under experimental?



Also per our offline discussions, 
1) remove the env node change and will follow up with another PR
2) init Callbacks twice now through a Hydra RUN, created a PR to follow up before Hydra 1.1 is officially released.

hydra/_internal/hydra.py
```diff
@@ -90,13 +91,18 @@ def run(
             run_mode=RunMode.RUN,
         )
 
-        return run_job(
+        callbacks = Callbacks(cfg)
```
omry, line 94: One downside of re-initializing is that a callback can no longer share state across different events (e.g, a callback can be a callable class that is updating members across events).

This is still okay for now though.
Are we planning a followup for initializing it once before 1.1?

hydra/core/callbacks.py
```diff
@@ -0,0 +1,51 @@
...
+            self.callbacks.append(instantiate(params))
+
+    def _notify(self, function_name: str, **kwargs: Any) -> None:
+        for c in self.callbacks:
+            try:
+                getattr(c, function_name)(**kwargs)
+            except Exception as e:
+                warnings.warn(
+                    f"Exception occurred during {function_name} with callback {type(c).__name__}, error message: {e}"
+                )
```
omry, line 29: Why not logger.warn() ?

hydra/core/callbacks.py
```diff
@@ -0,0 +1,51 @@
...
+        for params in config.hydra.callbacks.values():
+            self.callbacks.append(instantiate(params))
+
+    def _notify(self, function_name: str, **kwargs: Any) -> None:
+        for c in self.callbacks:
+            try:
+                getattr(c, function_name)(**kwargs)
+            except Exception as e:
+                warnings.warn(
+                    f"Exception occurred during {function_name} with callback {type(c).__name__}, error message: {e}"
```
omry, line 28: Something like this ismore compact and informative.
```suggestion
                    f"Callback {type(c).__name__}.{function_name} raised {type(e).__name__}: {e}"
```

hydra/core/callbacks.py
```diff
@@ -0,0 +1,51 @@
...
+from hydra.core.utils import JobReturn
+from hydra.errors import HydraException
+from hydra.utils import instantiate
+
+
+class Callbacks:
+    def __init__(self, config: DictConfig) -> None:
+        self.callbacks = []
+        if config.hydra.callbacks is None:
+            raise HydraException("Hydra Callbacks is not configured")
```
omry, line 17: What is the purpose of this error?

tests/test_apps/app_with_callbacks/custom_callback/conf/hydra/callbacks/custom_callback.yaml
```diff
@@ -0,0 +1,6 @@
+# @package hydra.callbacks
+
+custom_callback:
+  _target_: my_callbacks_impl.CustomCallback
```
omry, line 4: You can put the callback directly in my_app.

omry (changes requested):

jieru-hu:
Thanks for the review omry. I had to do a force push due to conflicts with master, all the new commits start from fa7e008

> One downside of re-initializing is that a callback can no longer share state across different events (e.g, a callback can be a callable class that is updating members across events).
This is still okay for now though.
Are we planning a followup for initializing it once before 1.1?


Yes, I created [this](https://github.com/facebookresearch/hydra/issues/1498) to track it.
Now that we've decided to only release plugins from master for future release, it is much easier to 
change the set up method in launcher and sweeper. I will look into this as a follow up.


> What is the purpose of this error?

removed, I don't think we need this.


> Why not logger.warn() ?

We call this in `hydra.run` and `hydra.multirun` before `run_job`, at which point the logger is not fully configured yet. Below is an example what the logging looks like with `warnings` and `logging` if an exception is thrown from `on_run_start`
### using warnings
```
[2021-03-31 13:58:00,491][HYDRA] CustomCallback on_run_start
/Users/jieru/workspace/hydra-fork/hydra/hydra/core/callbacks.py:25: UserWarning: Callback CustomCallback.on_run_start raised ValueError: SOMETHING WRONG
  warnings.warn(
```


### using logging
code changes
```
log.error(
                    f"Callback {type(c).__name__}.{function_name} raised {type(e).__name__}: {e}"
                )
```
application output
```
[2021-03-31 13:58:43,922][HYDRA] CustomCallback on_run_start
[2021-03-31 13:58:43,922][HYDRA] Callback CustomCallback.on_run_start raised ValueError: SOMETHING WRONG
```

omry:
> > I guess we can move this to hydra.experimental instead of hydra._internal.
> > we can followup with documentation later.
> 
> I feel `_internal` is more fitting for Callback in that Callbacks are more like Launcher/Sweepers and we do not expect users to call the event methods directly. Could you share more thought on why should fall under experimental?

Callbacks are a very powerful mechanism and I think many users will find it useful for integrating with various systems and for applying other custom logic automatically.

jieru-hu:
> Callbacks are a very powerful mechanism and I think many users will find it useful for integrating with various systems and for applying other custom logic automatically.
The reason to make it experimental and not immediately public is that it's too early to commit on the interface.
Going with _internal is a way for us to not make this a public feature (even easier to break and no need for documentation). However, this is starting to look pretty good and I think we can go with experimental.


OK - how about I will create a follow up PR to move this to experimental? :)

omry:
> > Why not logger.warn() ?
> 
> We call this in `hydra.run` and `hydra.multirun` before `run_job`, at which point the logger is not fully configured yet. Below is an example what the logging looks like with `warnings` and `logging` if an exception is thrown from `on_run_start`
> ### using warnings
> 
> ```
> [2021-03-31 13:58:00,491][HYDRA] CustomCallback on_run_start
> /Users/jieru/workspace/hydra-fork/hydra/hydra/core/callbacks.py:25: UserWarning: Callback CustomCallback.on_run_start raised ValueError: SOMETHING WRONG
>   warnings.warn(
> ```
> 
> ### using logging
> 
> code changes
> 
> ```
> log.error(
>                     f"Callback {type(c).__name__}.{function_name} raised {type(e).__name__}: {e}"
>                 )
> ```
> 
> application output
> 
> ```
> [2021-03-31 13:58:43,922][HYDRA] CustomCallback on_run_start
> [2021-03-31 13:58:43,922][HYDRA] Callback CustomCallback.on_run_start raised ValueError: SOMETHING WRONG
> ```


So, what's wrong with the console output in the logging case?

hydra/core/callbacks.py
```diff
@@ -1,23 +1,19 @@
...
 
 class Callbacks:
     def __init__(self, config: DictConfig) -> None:
         self.callbacks = []
-        if config.hydra.callbacks is None:
-            raise HydraException("Hydra Callbacks is not configured")
-
-        for params in config.hydra.callbacks.values():
-            self.callbacks.append(instantiate(params))
+        if config.hydra.callbacks is not None:
```
omry, line 14: based on the config, it should never be None:
```
# Callbacks configuration
callbacks: Dict[str, Any] = field(default_factory=dict)
```

jieru-hu:
> > > Why not logger.warn() ?
> > 
> > 
> > We call this in `hydra.run` and `hydra.multirun` before `run_job`, at which point the logger is not fully configured yet. Below is an example what the logging looks like with `warnings` and `logging` if an exception is thrown from `on_run_start`
> > ### using warnings
> > ```
> > [2021-03-31 13:58:00,491][HYDRA] CustomCallback on_run_start
> > /Users/jieru/workspace/hydra-fork/hydra/hydra/core/callbacks.py:25: UserWarning: Callback CustomCallback.on_run_start raised ValueError: SOMETHING WRONG
> >   warnings.warn(
> > ```
> > 
> > 
> > ### using logging
> > code changes
> > ```
> > log.error(
> >                     f"Callback {type(c).__name__}.{function_name} raised {type(e).__name__}: {e}"
> >                 )
> > ```
> > 
> > 
> > application output
> > ```
> > [2021-03-31 13:58:43,922][HYDRA] CustomCallback on_run_start
> > [2021-03-31 13:58:43,922][HYDRA] Callback CustomCallback.on_run_start raised ValueError: SOMETHING WRONG
> > ```
> 
> So, what's wrong with the console output in the logging case?

The error reporting is not as obvious as using a warning IMO.

hydra/core/callbacks.py
omry: My error message is more informative.

tests/test_apps/app_with_callbacks/custom_callback/my_callbacks_impl.py
```diff
@@ -0,0 +1,39 @@
...
+
+from omegaconf import DictConfig
+
+from hydra._internal.callback import Callback
+from hydra.core.utils import JobReturn
+
+log = logging.getLogger(__name__)
+
+
+class CustomCallback(Callback):
```
omry, line 14: For simplicity, can you move the callback implementation to tests/test_apps/app_with_callbacks/custom_callback/my_app.py?

omry (changes requested):

jieru-hu:
>based on the config, it should never be None:

fixed!


>My error message is more informative.


Oops, I applied the change, but something lost it... fixed now. sorry.


> For simplicity, can you move the callback implementation to tests/test_apps/app_with_callbacks/custom_callback/my_app.py?

done!

omry (approved):
Looking good overall. See final question.

jieru-hu:
>For non massive diffs (like the Ray one), we can take advantage of this flow and have short discussions in context.
Do we have a followup issue to consider promoting this to experimental?

I added it to my todo list :P, also just created #1523

PR #1445, opened 2021-03-03
Cleanup pytest imports to reduce tests boilerplate

omry:
switching the codebase from using 
```python
import pytest
```
to 
```python
from pytest import mark, param, raises, warns
```

To eliminate usage of `pytest.` everywhere in tests, making lines shorter and code formatting cleaner.


omry:
from @odelalleau ,@jieru:
May likely cause conflicts for tests in outstanding diffs.

PR #1447, opened 2021-03-04
fix fish tests

### LINKED ISSUE #1442: [Bug] Fish shell tests are broken with fish 3.2.0
There are a few expected outputs from fish 3.2.0 that seems to be breaking the tests.

@binshengliu - mind taking a look?
It looks like the trick to start fish without configuration no longer works as it was in previous versions:
```
 # Start fish without user configuration
 set env(HOME) /dev/null
```

This is example output:
```
omry@Coronadev:~/dev/hydra$ HOME=/dev/null fish;
error: Your history will not be saved.
warning-path: Unable to locate data directory derived from $HOME: '/dev/null/.local/share/fish'.
warning-path: The error was 'Not a directory'.
warning-path: Please set $HOME to a directory where you have write access.

error: Your personal settings will not be saved.
warning-path: Unable to locate config directory derived from $HOME: '/dev/null/.config/fish'.
warning-path: The error was 'Not a directory'.
warning-path: Please set $HOME to a directory where you have write access.

Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
omry@Coronadev /h/o/d/hydra (fish)> 
```

Even providing a clean home dir does not achieve the required behavior:
```
omry@Coronadev:~/dev/hydra$ HOME=/tmp/3231232 fish
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
omry@Coronadev /h/o/d/hydra (fish)> 
```

As a temp workaround I am disabling fish shell tests.

binshengliu:
## Motivation

Fix fish completion tests.

The failure was caused by a too strict regexp matching at preprocessing. `key=value` style completion was captured but `value_only` style completion was not. This PR fixes that.

Regarding HOME directory errors, I think we can safely ignore as fish is just complaining about not being able to save personal settings. Fish doesn't have a clean way to ignore configurations and dealing with this error is a bit over the top IMO.

Yes

## Test Plan

NA

## Related Issues and PRs

Fix #1442

### DIFF
--- a/tests/test_completion.py
+++ b/tests/test_completion.py
@@ -179,12 +179,7 @@ class TestRunCompletion:
     @mark.parametrize("prog", [["python", "hydra/test_utils/completion.py"]])
     @mark.parametrize(
         "shell",
-        [
-            "bash",
-            # Temporarily disabled, see https://github.com/facebookresearch/hydra/issues/1442
-            # "fish",
-            "zsh",
-        ],
+        ["bash", "fish", "zsh"],
     )
     def test_shell_integration(
         self,
@@ -227,7 +222,7 @@ class TestRunCompletion:
 
         if shell == "fish":
             # Fish will add a space to an unambiguous completion.
-            expected = [x + " " if re.match(r".*=\w+$", x) else x for x in expected]
+            expected = [x + " " if re.search(r"\w$", x) else x for x in expected]
 
             # Exactly match token end. See
             # https://github.com/fish-shell/fish-shell/issues/6928
--- a/tests/test_completion.py
+++ b/tests/test_completion.py
@@ -221,12 +221,8 @@ class TestRunCompletion:
         )
 
         if shell == "fish":
-            # Fish will add a space to an unambiguous completion.
-            expected = [x + " " if re.search(r"\w$", x) else x for x in expected]
-
-            # Exactly match token end. See
-            # https://github.com/fish-shell/fish-shell/issues/6928
-            expected = [re.escape(x) + "$" for x in expected]
+            # Escape special chars in the output
+            expected = [re.escape(x) for x in expected]
 
         cmd.extend(expected)
         if verbose:
--- a/tests/test_completion.py
+++ b/tests/test_completion.py
@@ -220,10 +220,6 @@ class TestRunCompletion:
             ]
         )
 
-        if shell == "fish":
-            # Escape special chars in the output
-            expected = [re.escape(x) for x in expected]
-
         cmd.extend(expected)
         if verbose:
             print("\nCOMMAND:\n" + " ".join([f"'{x}'" for x in cmd]))
--- a/tests/scripts/test_fish_completion.exp
+++ b/tests/scripts/test_fish_completion.exp
@@ -32,7 +32,7 @@ send "complete --do-complete \"$command\"\r"
 puts "\nMatching:\n"
 foreach arg [lrange $argv 3 end] {
     expect {
-	-re "$arg" {puts "matched '$arg' to '$expect_out(0,string)'\nbuffer: '$expect_out(buffer)'";continue}
-        timeout {puts "Error matching $arg";exit 1}
+	"$arg" {puts "matched '$arg' to '$expect_out(0,string)'\nbuffer: '$expect_out(buffer)'";continue}
+        timeout {puts "Error matching $arg"; exit 1}
     }
 }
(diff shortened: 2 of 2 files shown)


binshengliu:
@omry I can have all the fish tests passed with the proposed fix in my local Linux environment, but I don't have a macOS to debug the failures on it.

Sorry, I won't be able to work on this issue until 16 March but I'm happy to come back after that. (maybe allow me ssh into the CI env for debugging?)

omry:
Thanks @binshengliu, we can definitely wait a few weeks until you have time to take a closer look.

As for ssh access, it opens up security concerns because the CI instances got access to some secrets.

It is possible that you will be able to use your own CircleCI account for that (but from past experience it really eats the free credits quickly).
Try to manage without ssh access, if you get stuck we will figure it out...

omry:
About the warnings:
In general I think it's best to be very precise when testing with expect.

If you look at the bash test, you will notice that I am even consuming the prompt:
```
spawn bash --norc --noprofile
expect >>
```
This makes it easier to test and debug things.
To do this with fish you will need to use a custom fish_greeting function, which again means we need a custom fish.config.

binshengliu:
OK. There are two problems here.

1. macOS tests fail. This should be caused by some compatibility issue not test logic, so access to macOS would be ideal. We can discuss this later.
2. HOME warning. I think you meant `promt` instead of `greeting`? The purpose of setting HOME to /dev/null is exactly to expect a standard prompt. Setting up a temporary HOME and a customized fish.config is another way of doing it. It can get around the HOME writable warning but at the cost of adding some complexity in the test setup and cleanup. I'm ok with it if you prefer the fish.config way.

I'll ping you on zulip when I wrap up my current work.

omry:
Yeah, at some point I also saw fish_greeting.
in any case, sounds good - let's chat once you get more time.

jieru-hu:
@binshengliu 

As for the MacOS access, I've reached out to CircleCI and they recommended for this case, you can request MacOS access from your own account, could you try that? Basically you need to:

1. Create a circleCI account with your github if you have not
2. Follow Hydra in your own circleCI account (this will trigger the tests in your own account instead of Hydra's)
3. Email billing@circleci.com  for macOS OSS access (it is free)

Pls let us know if you run into any issues if you try the above :) 

Thanks again.

binshengliu:
Thanks! Will try that.

omry:
To avoid using all your free credits too quickly I also suggest that you disable all but the core tests on a mac and on a single Python version in the noxfile while you test.

binshengliu:
I'm looking at this issue now. Already contacted CircleCI and waiting for their response.

@jieru-hu Can you give me some instructions on how to achieve what we want once they give me access? I looked at the CircleCI interface and don't really understand how things work.

omry:
once you run a CI job and it fails, you can rerun it with ssh.
There are instructions in the ssh action about how to connect in.
image

[Here](https://circleci.com/docs/2.0/ssh-access-jobs/) is their help link for it.

binshengliu:
OK. It's now disabled for me. I'll see what happens after I have access.

tests/test_completion.py
```diff
@@ -226,12 +221,8 @@ def test_shell_integration(
...
 
         if shell == "fish":
-            # Fish will add a space to an unambiguous completion.
-            expected = [x + " " if re.match(r".*=\w+$", x) else x for x in expected]
-
-            # Exactly match token end. See
-            # https://github.com/fish-shell/fish-shell/issues/6928
-            expected = [re.escape(x) + "$" for x in expected]
+            # Escape special chars in the output
+            expected = [re.escape(x) for x in expected]
```
omry, line 225: what special chars?

omry (approved):
this looks really nice, thanks for taking the time.
one quick question about the escaping.

tests/test_completion.py
binshengliu: That was supposed to protect `.` and possible special chars in regex. Now we don't need to use regex so we can get rid of that branch. I've pushed a new commit to remove that. The code should look cleaner.

omry: Much cleaner indeed.
The fish expect test is now also (almost) identical to the bash expect test (which also did not use -re for matching and did not send tab in the match loop.
Almost because you have a missing space before exit here:
`timeout {puts "Error matching $arg";exit 1}`

How come you were able to remove the tab from the match loop?

binshengliu: Right. I just learned the `complete` command for testing completions and it's also how fish does its own completion tests. It returns a list of candidates without any formatting which makes the test easier.

Just pushed a new commit to make them more similar.

PR #1489, opened 2021-03-24
OmegaConf==2.1.0dev23

omry:
1. OmegaConf register_resolver is deprecated, this switch to the new register_new_resolver().
2. the `env` resolver is deprecated, this switches to `oc.env`.

### DIFF
--- a/plugins/hydra_rq_launcher/hydra_plugins/hydra_rq_launcher/config.py
+++ b/plugins/hydra_rq_launcher/hydra_plugins/hydra_rq_launcher/config.py
@@ -9,15 +9,15 @@ from omegaconf import II
 @dataclass
 class RedisConf:
     # host address via REDIS_HOST environment variable, default: localhost
-    host: str = II("env:REDIS_HOST,localhost")
+    host: str = II("oc.env:REDIS_HOST,localhost")
     # port via REDIS_PORT environment variable, default: 6379
-    port: int = II("env:REDIS_PORT,6379")
+    port: int = II("oc.env:REDIS_PORT,6379")
     # database via REDIS_DB environment variable, default: 0
-    db: Optional[str] = II("env:REDIS_DB,0")
+    db: Optional[str] = II("oc.env:REDIS_DB,0")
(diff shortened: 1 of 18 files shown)


plugins/hydra_rq_launcher/hydra_plugins/hydra_rq_launcher/config.py
```diff
@@ -9,15 +9,15 @@
...
+    port: int = II("oc.env:REDIS_PORT,6379")
     # database via REDIS_DB environment variable, default: 0
-    db: Optional[str] = II("env:REDIS_DB,0")
+    db: Optional[str] = II("oc.env:REDIS_DB,0")
     # password via REDIS_PASSWORD environment variable, default: no password
-    password: str = II("env:REDIS_PASSWORD,null")
+    password: str = II("oc.env:REDIS_PASSWORD,null")
     # switch to run without redis server in single thread, for testing purposes only
-    mock: bool = II("env:REDIS_MOCK,False")
+    mock: bool = II("oc.env:REDIS_MOCK,null")
```
omry, line 20: @odelalleau, here is an example where allowing uninterrupted passthrough for the default value would make things easier.

odelalleau: Yeah, I considered it as well. It's a trade off: on one hand this can enable some shortcuts. On the other hand, it may occasionally lead to tricky bugs due to suddenly feeding a non-string input to something expecting a string.

Just a note that this specific diff is "wrong" in the sense that `mock` is now `Optional[bool]`. For instance in a structured config this would fail:
```python
@dataclass
class Config:
    x: bool = "${oc.env:MISSING,null}"
        
cfg = OmegaConf.structured(Config)
cfg.x  # InterpolationValidationError: Non optional field cannot be assigned None
```

A better fix (IMO) is to provide the string representation of the value you want to use, as would be found in the environment variable: `II("oc.env:MISSING,'False'")`. This is point (iii) in https://github.com/omry/omegaconf/issues/573 => I'll add an example there with a boolean (right now it's just with an integer value)

PR #1495, opened 2021-03-25
fix ray launcher install commands

jieru-hu:
modify the doc to be consistent with other plugin installation.


omry (approved):
check that other plugin pages are consistent too if you haven't.

jieru-hu:
> check that other plugin pages are consistent too if you haven't.

I did :) they are consistent.

PR #1497, opened 2021-03-25
remove optuna from 1.0_branch

jieru-hu:
optuna sweeper is an Hydra 1.1 only plugin and it has been released from Hydra 1.1 from day 1. 

Removing this from 1.0_branch to avoid confusion.


omry:
I actually removed it and accidentally pushed straight to 1.0_branch already.

PR #1502, opened 2021-03-26
Plugins are now obtaining version from hydra_plugins.PLUGIN.__version__

omry:
This is establishing a single source of truth for the plugin's version, both for towncrier and for package creation.
It's using the the build helpers in the core repo to parse the versions like hydra-core.
Sharing code between setup.py files is tricky, so this is a bit ugly.
The alternative (which I have been through) is to write the same function 9 times. That's even less fun.

tested with:
```
omry@Coronadev:~/dev/hydra/plugins/hydra_submitit_launcher$ python setup.py  --version
1.2.0
omry@Coronadev:~/dev/hydra/plugins/hydra_submitit_launcher$ towncrier  --draft
Loading template...
Finding news fragments...
Rendering news fragments...
Draft only -- nothing has been written.
What is seen below is what would be written.

1.2.0 (2021-03-25)
==================


### Features

- Add support for SLURM parameters `cpus_per_gpu`, `gpus_per_task`, `mem_per_gpu` and `mem_per_cpu` ([#1366](https://github.com/facebookresearch/hydra/issues/1366))
```

### DIFF
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/__init__.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/__init__.py
@@ -1,1 +1,3 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+__version__ = "1.1.0rc1"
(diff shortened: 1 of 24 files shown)


omry:
This was done in a different PR.

Issue #1472, opened 2021-03-12
[Feature Request] Additional Configuration for Optuna plugin


monney:
# 🚀 Feature Request

Hi I would like to be able to change the sampler settings from default.
## Motivation

Currently we can only pass what sampler to use but not any settings.

## Pitch

**Describe the solution you'd like**
It would be nice if we could pass in arbitrary sampler configs to be instantiated internally.

**Describe alternatives you've considered**
There are only a couple  hot settings I really care about, I imagine most people only change a few things, you could selectively support just a few options.

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))
Yes

omry:
cc @toshihikoyanase

Hi @monney,
Yes, this makes sense. 

I think we can leverage the new 1.1 support for nested defaults list for this.
It would be great if you initially try to make it without guidance, just based on the 1.1 docs.
[Here](https://hydra.cc/docs/next/advanced/defaults_list) is the doc for the new Defaults List.

monney:
Closed by #1503

PR #1503, opened 2021-03-26
Add Configuration of Optuna Sampler settings

monney:
## Motivation
See #1472

Yes

## Test Plan
Passing current tests for the plugin, will add more if needed.


omry:
Not a detailed review, but it looks really nice!
@monney, can you jump to the chat? I want to ask you a few questions about your experience with recursive defaults.

cc @toshihikoyanase, check this out.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
```diff
@@ -176,39 +175,39 @@ class OptunaSweeperConf:
...
+    package="hydra.sweeper.sampler",
     node=TPESamplerConfig,
     provider="optuna_sweeper",
 )
 
 ConfigStore.instance().store(
     group="hydra/sweeper/sampler",
     name="random",
-    package="hydra.sweeper.optuna_config.sampler",
+    package="hydra.sweeper.sampler",
```
omry, line 186: The default package is `_group_`, which is exactly what you have here. You can now omit package the package from the store calls for the new configs.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
```diff
@@ -161,8 +156,12 @@ class OptunaSweeperConf:
     _target_: str = "hydra_plugins.hydra_optuna_sweeper.optuna_sweeper.OptunaSweeper"
     defaults: List[Any] = field(default_factory=lambda: defaults)
 
-    optuna_config: OptunaConfig = OptunaConfig()
+    # Sampling algorithm
+    # Please refer to the reference for further details
+    # https://optuna.readthedocs.io/en/stable/reference/samplers.html
+    sampler: Any = MISSING
 
+    optuna_config: OptunaConfig = OptunaConfig()
```
omry, line 164: I think we can inline the content of OptunaConfig in this node to simplify the config further.
Maybe this can go into a followup diff as it's another breaking change. what do you think?

plugins/hydra_optuna_sweeper/example/conf/config.yaml
```diff
@@ -4,14 +4,14 @@ defaults:
...
 hydra:
   sweeper:
+    sampler:
+      seed: 123
     optuna_config:
       direction: minimize
       study_name: sphere
       storage: null
       n_trials: 20
       n_jobs: 1
```
omry, line 14: Okay, I think we can completely kill optuna_config here and go for this:
```suggestion
hydra:
  sweeper:
    sampler:
      seed: 123
    direction: minimize
    study_name: sphere
    storage: null
    n_trials: 20
    n_jobs: 1
```

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
```diff
@@ -17,12 +18,86 @@ class Direction(Enum):
...
 
 
-class Sampler(Enum):
-    tpe = 1
-    random = 2
-    cmaes = 3
-    nsgaii = 4
-    motpe = 5
+@dataclass
+class TPESamplerConfig:
```
omry, line 22: You can create a superclass for all of those to get better type safety.
```python
@dataclass
class SamplerConfig:
  _target_: str = MISSING

@dataclass
class TPESamplerConfig(SamplerConfig):
  ...

@dataclass
class OptunaSweeperConf:
    _target_: str = "hydra_plugins.hydra_optuna_sweeper.optuna_sweeper.OptunaSweeper"
    ...
    sampler: SamplerConfig = MISSING
```

website/docs/plugins/optuna_sweeper.md
```diff
@@ -114,8 +127,8 @@ The output is as follows:
 ```commandline
 [HYDRA] Study name: sphere
 [HYDRA] Storage: None
-[HYDRA] Sampler: TPESampler
-[HYDRA] Direction: minimize
+[HYDRA] Sampler: <optuna.samplers._tpe.sampler.TPESampler object at 0x7fd7901d6d30>
```
omry, line 130: @toshihikoyanase, it would be nice to have an informative string representation of the samplers.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
monney: will do

plugins/hydra_optuna_sweeper/example/conf/config.yaml
monney: will do

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
monney: will do

website/docs/plugins/optuna_sweeper.md
monney: This used to print the enum, which was interpretable. That is gone now, Optuna doesn't have a nice repr for their samplers which causes this. I'm not sure what the best solution is to get a nice print out

monney: I ended up using `type(sampler).__name_`

omry: I think it should be straight forward and useful to add a `__repr__` to the samplers (and possibly to other objects).
Printing the type is good enough for now, if Optuna adds the repr we can use it here for more informative representation.

What is interesting to me is that by using recursive instantiation, we have lost the original config and now we have the object.
You can still gain access to the Hydra config though!
See https://github.com/facebookresearch/hydra/blob/master/examples/configure_hydra/job_name/no_config_file_override.py

Try something like:
`HydraConfig.get().sweeper.sampler`.
That should have the DictConfig object. if you print it as it will look like a plain dictionary.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
```diff
@@ -72,33 +158,47 @@ class OptunaConfig:
...
-
-    # Random seed of sampler
-    seed: Optional[int] = None
+    search_space: Dict[str, Any] = field(default_factory=dict)
 
-    def __post_init__(self) -> None:
-        if isinstance(self.direction, list) and len(self.direction) > 1:
-            if self.sampler == Sampler.tpe:
-                # The default sampler of multi-objective optimization is NSGAIISampler
-                self.sampler = Sampler.nsgaii
```
omry, line 87: Okay, so we have lost this logic.

@toshihikoyanase, is it an error to use tpe with multi-objective optimization?
should we fail here? should we issue a warning?
Without introducing a new config group for multi-objective, I don't think we can make the default different (and this seems like an overkill to me).

The example app from @monney is using nsgaii.
We can add a warning or an error if the combination is not supported (but I think it will make more sense if it comes from Optuna itself).

plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
```diff
@@ -162,10 +162,10 @@ def test_optuna_multi_objective_example(with_commandline: bool, tmpdir: Path) ->
...
         "--multirun",
         "hydra.sweep.dir=" + str(tmpdir),
-        "hydra.sweeper.optuna_config.n_trials=20",
-        "hydra.sweeper.optuna_config.n_jobs=1",
-        "hydra.sweeper.optuna_config.sampler=random",
-        "hydra.sweeper.optuna_config.seed=123",
+        "hydra.sweeper.n_trials=20",
+        "hydra.sweeper.n_jobs=1",
+        "hydra/sweeper/sampler=random",
+        "hydra.sweeper.sampler.seed=123",
```
omry, line 168: sweet.

omry (commented):
This is looking really nice.

Add a couple of news fragments into the news directory:
one a feature (support configuring the samplers), and the second an api_change for changing the config structure.

Something like:
1472.feature:
```
Add support for configuring the Samplers
```

1472.api_change:
```
Config structure changes, please refer to documentation page
```

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
monney: @omry good catch. I knowingly dropped this for the reason you outlined. Optuna already triggers an error before the sweep starts if this is set to TPE.
`ValueError: If the study is being used for multi-objective optimization, TPESampler cannot be used.`

monney:
@omry tests failed but they have nothing to do with my code, and didn't fail on the last commit

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
omry: Yeah, it looks good enough for me.

news/1472.api_change
```diff
@@ -0,0 +1 @@
+Optuna sweeper config stucture changes, please refer to the [plugin docs](https://hydra.cc/docs/next/plugins/optuna_sweeper/)
```
omry, line 1: there should be a news directory inside the Optuna sweeper dir (if there isn't make one).
move your two news fragments into it.

omry (commented):
one nit.
Failing tests are not related to your changes, don't worry about them.

omry (approved):
Awesome, thanks!

website/docs/plugins/optuna_sweeper.md
toshihikoyanase: Sorry for the delayed response.

> I ended up using `type(sampler).__name__`
> I think it should be straight forward and useful to add a `__repr__` to the samplers (and possibly to other objects).

Thank you for your discussion. I totally agree with you and create a feature request in the Optuna repository (https://github.com/optuna/optuna/issues/2536).

toshihikoyanase:
@monney Thank you for adding the configuration of samplers. It definitely improves the usability of Optuna sweeper.

@omry Sorry for the delayed response. It looks great to me!

I have a comment of the default sampler for single objective optimization discussed in this thread: https://github.com/facebookresearch/hydra/pull/1503#discussion_r602633069.

 
This PR changed the default sampler of Optuna sweeper from `tpe` to  `nsgaii`. As you discussed, NSGA-II does not raise errors or warnings even if it is applied to a single objective optimization, but it may significantly degrade the search efficiency because NSGA-II works like a quite simple genetic algorithm.

 > Without introducing a new config group for multi-objective, I don't think we can make the default different (and this seems like an overkill to me).

Just a confirmation, but can we somehow overwrite the default sampler by Python code instead of introducing a config group?

omry:
> This PR changed the default sampler of Optuna sweeper from `tpe` to `nsgaii`. As you discussed, NSGA-II does not raise errors or warnings even if it is applied to a single objective optimization, but it may significantly degrade the search efficiency because NSGA-II works like a quite simple genetic algorithm.

[Here](https://github.com/facebookresearch/hydra/pull/1503/files#diff-35bc09021922daf7005966ef790182d0f61fad0c9f04e280d51576ee9228916bR129) is the default sampler, still tpe.
Are you looking at the [example for multi-objective optimization](https://github.com/facebookresearch/hydra/pull/1503/files#diff-b385b1f45a28b7d68afe64f6865d502d4adc08bff00acd056b08942ce96d706eR3)?

> > Without introducing a new config group for multi-objective, I don't think we can make the default different (and this seems like an overkill to me).
> 
> Just a confirmation, but can we somehow overwrite the default sampler by Python code instead of introducing a config group?

Because this is using recursive instantiation, the sampler is already instantiated by the time the sweeper initializes.
You can't change the default from the code. (You can apply logic to replace it if you want, but based on my above comment I think it should not be needed).

toshihikoyanase:
@omy Thank you for your answer. I misunderstood the code, and the default sampler is actually tpe.

The default sampler of multi-objective optimization was changed from NSGA-II to TPE in this PR, right?
And, we got the `ValueError` if we don't set the multi-objective samplers. This is not aligned with Optuna's default settings, but I think it is reasonable for simplicity.

> Because this is using recursive instantiation, the sampler is already instantiated by the time the sweeper initializes. You can't change the default from the code.

I got it.
I just thought I could somehow get raw configurations in the code. It'll make things complicated and it is not a good idea.
Thanks!

omry:
> The default sampler of multi-objective optimization was changed from NSGA-II to TPE in this PR, right?
And, we got the ValueError if we don't set the multi-objective samplers. This is not aligned with Optuna's default settings, but I think it is reasonable for simplicity.

The plugin is depending on Optuna to raise an exception.
The example for multi-objective optimization shows the correct sampler to use.
How does Optuna select the sampler by default normally?

>> Because this is using recursive instantiation, the sampler is already instantiated by the time the sweeper initializes. You ca
n't change the default from the code.
> I got it.
I just thought I could somehow get raw configurations in the code. It'll make things complicated and it is not a good idea.
Thanks!

You can get raw configuration but it's a bit hard (see my comment about HydraConfig). At that point the configuration was already used to instantiate the object.
Even if the object was not already instantiated, the config is the composed config. at that point you can't change the default choice for the config group.

PR #1490, opened 2021-03-24
Update ray example to reflect Hydra 1.1 changes

jieru-hu:
Closes #1471 
Verified the example apps run as expected.

PR #1506, opened 2021-03-26
Add doc for configuring plugin in Hydra 1.1

jieru-hu:
This PR is to follow up on the comment in https://github.com/facebookresearch/hydra/pull/1490

@omry I haven't added tests for the example apps, i want to get your input on the content/examples of the doc first.

### DIFF
--- a/website/docs/patterns/configuring_plugins.md
+++ b/website/docs/patterns/configuring_plugins.md
@@ -1,0 +1,186 @@
+---
+id: configuring_plugins
+title: Configuring Plugins
+---
+
+import {ExampleGithubLink} from "@site/src/components/GithubLink"
+
+Hydra plugins usually comes with sensible defaults which works with minimal configuration.
+There are two primary ways to customize the configuration of a plugin:
+- Overriding it directly in your primary config
+- Extending the config and using it from your primary config.
+	
+The first method is the simplest, but it makes it harder to switch to a different plugin configuration.
+The second method is a bit more complicated, but makes it easy to switch between different plugin configurations.
+
+
+The following methods apply to all Hydra plugins. In the following examples, we will configure a imaginary Launcher plugin
+`MoonLauncher`. The Launcher has two modes: `falcon9`, which actually launches the application to the Moon and 
+`simulation` which simulates a launch.
+
+The config schema for MoonLauncher looks like:
+
+
+
+<div className="row">
+<div className="col col--6">
+
+```python
+@dataclass
+class Falcon9Conf:
+  ton_fuel:  int = 10
+
+
+
+
+```
+</div>
+<div className="col  col--6">
+
+```python
+@dataclass
+class Simulation:
+  ton_fuel:  int = 10
+  window_size:
+    width: 1024
+    height: 768
+
+```
+</div>
+</div>
+
+
+
+### Overriding in primary config
+We can directly override Launcher config in primary config.
+
+<div className="row">
+<div className="col col--4">
+
+```yaml title="config.yaml" 
+a: 1
+
+hydra:
+  launcher:
+    ton_fuel: 2
+
+
+
+
+
+
+
+
+
+```
+</div>
+<div className="col col--4">
+
+```commandline title="command-line override" 
+hydra/launcher=falcon9
+
+
+```
+```yaml title="resulting launcher config"  {3}
+hydra:
+  launcher:
+    ton_fuel: 2
+
+
+
+```
+</div>
+<div className="col col--4">
+
+```commandline title="command-line override" 
+hydra/launcher=simulation
+
+
+```
+```yaml title="resulting launcher config"  {3}
+hydra:
+  launcher:
+    ton_fuel: 2
+    window_size:
+      width: 1024
+      height: 768
+```
+</div>
+</div>
+
+
+
+
+### Extending plugin default config
+
+This section assumes that you are familiar with the contents of [Common Patterns/Extending Configs](patterns/extending_configs.md).
+
+Extending plugin default config has several advantages:
+- Separate configuration concerns, keep primary config clean.
+- Easier to switch between different plugin configurations.
+- Provides flexibility when a Plugin has different modes
+that requires different schema.
+
+
+Say that we want to override certain values for different Launcher mode:
+
+<div className="row">
+<div className="col col--6">
+
+```yaml title="hydra/launcher/my_falcon9" {4}
+defaults:
+  - falcon9
+
+ton_fuel: 2
+
+```
+</div>
+<div className="col col--6">
+
+```yaml title="hydra/sweeper/my_simulation" {5}
+defaults:
+  - simulation
+
+window_size:
+  width: 768
+
+```
+</div>
+</div>
+
+We can easily user command-line overrides to get the configuration needed:
+<div className="row">
+
+<div className="col col--6">
+
+```commandline title="command-line override" 
+hydra/launcher=my_falcon9
+
+
+```
+```yaml title="resulting launcher config"  {3}
+hydra:
+  launcher:
+    ton_fuel: 2
+
+
+
+```
+</div>
+<div className="col col--6">
+
+```commandline title="command-line override" 
+hydra/launcher=my_simulation
+
+
+```
+```yaml title="resulting launcher config"  {5}
+hydra:
+  launcher:
+    ton_fuel: 10
+    window_size:
+      width: 768
+      height: 768
+```
+</div>
+</div>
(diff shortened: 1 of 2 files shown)


website/docs/patterns/configuring_plugins.md
```diff
@@ -0,0 +1,127 @@
...
+To achieve the same configuration, we can also create a separate config for the sweeper which helps keeping the primary config clean.
+
+<div className="row">
+<div className="col col--6">
+
+```yaml title="hydra/sweeper/custom_sweeper.yaml" 
+defaults:
+  - basic
+
+max_batch_size: 1
```
omry, line 113: batch of 2 may make the difference more obvious. took me a moment to see the difference in output.

website/docs/patterns/configuring_plugins.md
```diff
@@ -0,0 +1,127 @@
...
+</div>
+
+
+
+### Extending plugin default config
+
+<ExampleGithubLink text="Example application" to="examples/patterns/configuring_plugins_extend_defaults"/>
+
+To achieve the same configuration, we can also create a separate config for the sweeper which helps keeping the primary config clean.
+
```
omry, line 105: Explain that the advantages are.
(Easier to switch between different plugin configuration, can work if the different configuration has a different schema etc).

The example may work better if we have two parameters in the plugin config.

You can show two different configurations, say - batch of 1 and batch of 2 and show how you are switching between them.

website/docs/patterns/configuring_plugins.md
jieru-hu: True, I've updated the example.

website/docs/patterns/configuring_plugins.md
jieru-hu: > The example may work better if we have two parameters in the plugin config.

Yes - however for the examples I want to use a Plugin out of hydra-core (for simplicity and testing purpose), as a result, the choice is limited :(

omry: You can talk about a hypothetical plugin without a concrete example.

For this example, we will configure and imaginary Launcher plugin called MoonLauncher.
MoonLauncher can launch Hydra applications to the Moon. It also has a simulation mode, where the application is not really launched to the moon.

...
```

Comment (jieru-hu) [reply website/docs/patterns/configuring_plugins.md]: Thanks @omry , nice example :)

Although I still feel a runnable example is easier to comprehend than a hypothetical plugin - if you feel strongly about having the MoonLauncher I can try to update the doc further (and probably delete the new examples I've added in this PR.)

Comment (omry) [reply website/docs/patterns/configuring_plugins.md]: We will have plenty of runnable examples in all the plugins, right?
This is more about explaining the concepts.

Comment (omry) [reply website/docs/patterns/configuring_plugins.md]: This is just applying "Extending Configs" in the context of plugins. We have examples both in Extending Configs and in the plugins themselves.

Comment (jieru-hu) [reply website/docs/patterns/configuring_plugins.md]: @omry 

Sounds good. Just so we are on the same page here: we will get rid of all the BasicSweeper examples here and only use `MoonLauncher` in this page, correct?

Comment (omry) [reply website/docs/patterns/configuring_plugins.md]: I think so. I can see the advantage of using a concrete plugin for the example. But at the same time, it makes it less useful for explaining the general concept of configuring other plugins.
Using an obviously made up plugin will probably make the point that this applies to all Hydra plugins easier to make.

**Review by omry — CHANGES_REQUESTED** [review]
I think you have the ball here.

**Comment by omry** [comment]
What's the plan here?

**Comment by jieru-hu** [comment]
I've been prioritizing non doc changes, will wrap this up soon.

**Comment by omry** [comment]
Sounds good.

File: website/docs/patterns/configuring_plugins.md [hunk 3]
```diff
@@ -0,0 +1,186 @@
...
+    ton_fuel: 2
+
+
+
+```
+</div>
+<div className="col col--4">
+
+```commandline title="command-line override" 
+hydra/launcher=simulation
```
omry, line 96: This has a horizontal scrollbar.
Try to rename the config name for simluation to something shorter (sim?) or to reorgranize the page a bit to give this column more space.

website/docs/patterns/configuring_plugins.md
```diff
@@ -0,0 +1,186 @@
...
+    height: 768
+
+```
+</div>
+</div>
+
+
+
+### Overriding in primary config
+We can directly override Launcher config in primary config.
```
omry, line 55: The problem with this approach, that you example is not exposing - is that if you want to override variables in the simulation (like window_size) in your primary config, it will not be compatible with falcon9 config anymore.
It's worth explaining this.

website/docs/patterns/configuring_plugins.md
```diff
@@ -0,0 +1,186 @@
...
+defaults:
+  - falcon9
+
+ton_fuel: 2
+
+```
+</div>
+<div className="col col--6">
+
+```yaml title="hydra/sweeper/my_simulation" {5}
```
omry, line 140: make a pass on the page and make sure all the line highlights are actually focusing on something relevant.
I did not understand why you highlight line 5 here.

omry: Also, make sure that the boxes are in the same size.
image

omry (changes requested):
Very nice.
See inline comments.

jieru-hu:
>make a pass on the page and make sure all the line highlights are actually focusing on something relevant.
I did not understand why you highlight line 5 here.

line 5 here is where the `my_sim` config overrides (window_size.width) so I highlighted it. 

>Also, make sure that the boxes are in the same size.

Thanks, I missed that one :) 

>The problem with this approach, that you example is not exposing - is that if you want to override variables in the simulation (like window_size) in your primary config, it will not be compatible with falcon9 config anymore.
It's worth explaining this.

Thanks, that's a great point! (and it was a pain point for me when i was working on one ray example for both local and aws mode, so happy it is not supported in 1.1 :)  )
I've updated the doc accordingly.

website/docs/patterns/configuring_plugins.md
```diff
@@ -39,7 +39,7 @@ class Falcon9Conf:
 
 ```python
 @dataclass
-class Simulation:
+class Sim:
```
omry, line 42: We can keep the class name `Simulation`, no one said that the config group name must match the config class name exactly.

website/docs/patterns/configuring_plugins.md
```diff
@@ -0,0 +1,190 @@
...
+import {ExampleGithubLink} from "@site/src/components/GithubLink"
+
+Hydra plugins usually comes with sensible defaults which works with minimal configuration.
+There are two primary ways to customize the configuration of a plugin:
+- Overriding it directly in your primary config
+- Extending the config and using it from your primary config.
+	
+The first method is the simplest, but it makes it harder to switch to a different plugin configuration.
+The second method is a bit more complicated, but makes it easy to switch between different plugin configurations.
+
```
omry, line 15:
```suggestion
The first method is simpler, but it makes it harder to switch to a different plugin configuration.
The second method is a bit more complicated, but makes it easy to switch between different plugin configurations.
```

website/docs/patterns/configuring_plugins.md
```diff
@@ -0,0 +1,190 @@
...
+      height: 768
+```
+</div>
+</div>
+
+
+This approach works well for common configs across different modes. What if we want to override `window_size.width`  for `sim` mode? 
+Overriding the value in primary config will limit launching option to only
+`hydra/launcher=sim` - we can no longer override `hydra/launcher=falcon9` because  `falcon9` does not support configuring
+`window_size.width`. In the next section, we will look at how to extend plugin default for more configuration flexibility.
```
omry, line 115:
```suggestion
This approach makes the assumption that the Launcher used has all the fields we are overriding.
If we wanted to override a field that exists in the Simulation Launcher but not in the Falcon9 Launcher, like window_size.width`, we would no longer be able to use the Falcon9 Launcher!  
The next section solves this problem.
```

website/docs/patterns/configuring_plugins.md
```diff
@@ -0,0 +1,190 @@
...
+- Provides flexibility when a Plugin has different modes
+that requires different schema.
+
+
+Say that we want to override certain values for different Launcher mode:
+
+<div className="row">
+<div className="col col--6">
+
+```yaml title="hydra/launcher/my_falcon9" {4}
```
omry, line 133: Use .yaml extention in the title for config files.
```suggestion
```yaml title="hydra/launcher/my_falcon9.yaml" {4}
```

omry (commented):
Shaping up nicely.
a few more comments.

A followup would be to change the documentation of all official plugins to reference this page.

jieru-hu:
thanks! I've created https://github.com/facebookresearch/hydra/issues/1582 to track the followup.

omry (approved):

jieru-hu:
rebase master... will merge after CI

Issue #1337, opened 2021-01-26
Add unit test for zipfile.Path 


jieru-hu:
see https://github.com/facebookresearch/hydra/pull/1317 for context.

PR #1528, opened 2021-04-01
Add test and check for zipfile.Path

jieru-hu:
This is to address the test failure we got after porting 1.1 to fbcode - T88075022
I did a similar change for `1.0_branch` here https://github.com/facebookresearch/hydra/pull/1317 (but forgot to port the change back to 1.1 🤦🏻‍♀️)
Also added a test, which will close https://github.com/facebookresearch/hydra/issues/1337


hydra/_internal/core_plugins/importlib_resources_config_source.py
```diff
@@ -29,14 +30,16 @@ def __init__(self, provider: str, path: str) -> None:
...
+    def _read_config(self, res: Any) -> ConfigResult:
+        try:
+            if sys.version_info[0:2] >= (3, 8) and isinstance(res, zipfile.Path):
+                f = res.open()
+                header_text = f.read(512)
+                if isinstance(header_text, bytes):
+                    header_text = header_text.decode("utf-8")
+            else:
+                f = res.open(encoding="utf-8")
+                header_text = f.read(512)
```
omry, line 42: Does this path also work for regular files?
If so, we should be able to just keep this and remove the else.

tests/test_config_repository.py
```diff
@@ -194,3 +196,16 @@ def test_restore_singleton_state_hack() -> None:
...
     """
     Singleton.set_state(state)
+
+
+@mark.skipif(sys.version_info[0:2] < (3, 8), reason="only run on python 3.8+")
+def test_importlib_resource_load_zip_path(monkeypatch: Any) -> None:
+    config_source = ImportlibResourcesConfigSource(provider="foo", path="pkg://bar")
+    conf = config_source._read_config(
+        zipfile.Path(  # type: ignore
+            "tests/test_apps/config_source_test/importlib_resource_zip_test/conf.zip",
```
omry, line 206: the location of this file does not make sense to me.
test_apps is reserved to full application, it's not a generic directory.
`hydra/test_utils/configs` is a better place for it.

omry (changes requested):

jieru-hu:
> Does this path also work for regular files?
If so, we should be able to just keep this and remove the else.


Nice!!! removed :) 


> the location of this file does not make sense to me.
test_apps is reserved to full application, it's not a generic directory.
hydra/test_utils/configs is a better place for it.


moved!

omry:
As usual, we have issues on Windows. :/.

jieru-hu:
> As usual, we have issues on Windows. :/.

now I remembered why we had to added these in the first place 😕, added the `else` statements back.

hydra/_internal/core_plugins/importlib_resources_config_source.py
omry: Can you add a comment explaining why we need the else so we both remember?

jieru-hu: good idea! added a comment.

omry (approved):
Awesome.

PR #1531, opened 2021-04-02
add hydra/env config group

### LINKED ISSUE #1529: Add `env` to Hydra's config group
This is a follow up to #1441

the `env` config group will allows users to manually change the env defaults value. (such as provides default callbacks or update run.dir )

jieru-hu:
Closes #1529

this is a follow up to #1441

introduce an `env` config group to HydraConf, this will be used for enviroment specific overrides that will come by default for all Hydra applications in that environment.


jieru-hu:
> I think using None here will eliminate the need for an empty hydra/env/default.yaml.

per our offline chats, keeping the default value to make it easier to override

> env_plugin sounds like a new kind of plugin. but this is not.
I suggest calling env_defaults instead of env_plugin.

sounds good, renamed the files, plugins and the test file.

> A callback (and the corresponding code) is too heavyweight to make the point here.
How about changing something lighter?
An interesting one that is straight forward to test is to inject an environment variable with hydra.job.env_set.
Alternatively, maybe hydra.job.name.

sounds good! I've modified the tests apps to override `hydra.job.env_set` instead.

omry (approved):
noice.

PR #1540, opened 2021-04-06
Raise UserWarning if hydra.searchpath is not available

### LINKED ISSUE #1469: enhancements to hydra.searchpath
follow up from omry's comments in  https://github.com/facebookresearch/hydra/pull/1468
 - [x] add --info searchpath
 - [ ] the searchpath is incomplete and does not contain the effective searchpath from the config file.
 - [ ] output a warning of a config searchpath configured via hydra.searchpath is not pointing to a valid location.

jieru-hu:
Closes #1469

### DIFF
--- a/examples/advanced/config_search_path/conf/cofig_with_no_searchpath.yaml
+++ b/examples/advanced/config_search_path/conf/cofig_with_no_searchpath.yaml
@@ -1,0 +1,2 @@
+defaults:
+  - dataset: cifar10
--- a/examples/advanced/config_search_path/conf/cofig_with_no_searchpath.yaml
+++ b/examples/advanced/config_search_path/conf/cofig_with_no_searchpath.yaml
@@ -1,2 +1,0 @@
-defaults:
-  - dataset: cifar10
--- a/examples/plugins/example_searchpath_plugin/setup.py
+++ b/examples/plugins/example_searchpath_plugin/setup.py
@@ -13,7 +13,9 @@ with open("README.md", "r") as fh:
         long_description=LONG_DESC,
         long_description_content_type="text/markdown",
         url="https://github.com/facebookresearch/hydra/",
-        packages=find_namespace_packages(include=["hydra_plugins.*"]),
+        packages=find_namespace_packages(
+            include=["hydra_plugins.*", "arbitrary_package.*"]
+        ),
         classifiers=[
             # Feel free to use another license.
             "License :: OSI Approved :: MIT License",
--- a/examples/plugins/example_searchpath_plugin/setup.py
+++ b/examples/plugins/example_searchpath_plugin/setup.py
@@ -15,7 +15,7 @@ with open("README.md", "r") as fh:
         long_description_content_type="text/markdown",
         url="https://github.com/facebookresearch/hydra/",
         packages=find_namespace_packages(include=["hydra_plugins.*"])
-        + ["arbitrary_package"],
+        + find_packages(include=["arbitrary_package.*"]),
         classifiers=[
             # Feel free to use another license.
             "License :: OSI Approved :: MIT License",
--- a/examples/plugins/example_searchpath_plugin/setup.py
+++ b/examples/plugins/example_searchpath_plugin/setup.py
@@ -15,7 +15,7 @@ with open("README.md", "r") as fh:
         long_description_content_type="text/markdown",
         url="https://github.com/facebookresearch/hydra/",
         packages=find_namespace_packages(include=["hydra_plugins.*"])
-        + find_packages(include=["arbitrary_package"]),
+        + ["arbitrary_package"],
         classifiers=[
             # Feel free to use another license.
             "License :: OSI Approved :: MIT License",
--- a/examples/plugins/example_searchpath_plugin/setup.py
+++ b/examples/plugins/example_searchpath_plugin/setup.py
@@ -13,9 +13,8 @@ with open("README.md", "r") as fh:
         long_description=LONG_DESC,
         long_description_content_type="text/markdown",
         url="https://github.com/facebookresearch/hydra/",
-        packages=find_namespace_packages(
-            include=["hydra_plugins.*", "arbitrary_package.*"]
-        ),
+        packages=find_namespace_packages(include=["hydra_plugins.*"])
+        + ["arbitrary_package"],
         classifiers=[
             # Feel free to use another license.
             "License :: OSI Approved :: MIT License",
--- a/examples/plugins/example_searchpath_plugin/setup.py
+++ b/examples/plugins/example_searchpath_plugin/setup.py
@@ -1,6 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 # type: ignore
-from setuptools import find_namespace_packages, setup
+from setuptools import find_namespace_packages, setup, find_packages
+
 
 with open("README.md", "r") as fh:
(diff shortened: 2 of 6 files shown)


hydra/_internal/config_loader_impl.py
```diff
@@ -206,6 +206,13 @@ def _err() -> None:
             new_csp.append(provider=provider, path=sp)
         new_csp.append("schema", "structured://")
         repo.initialize_sources(new_csp)
+        for source in repo.get_sources():
+            if source.provider == provider:
+                if not source.available():
+                    warnings.warn(
+                        category=UserWarning,
+                        message=f"hydra.searchpath={searchpath} is not available.",
```
omry, line 214: This seems wrong.
you are checking source and you are printing about searchpath.

omry (changes requested):

hydra/_internal/config_loader_impl.py
jieru-hu: per offline chats, the source here is filtered to be for `searchpath` which is why the message says searchpath.

since we want the warnings for all the sources, I will move this check elsewhere.

examples/plugins/example_launcher_plugin/tests/test_example_launcher_plugin.py
```diff
@@ -32,6 +38,9 @@ class TestExampleLauncher(LauncherTestSuite):
     "task_launcher_cfg, extra_flags",
     [({}, ["-m", "hydra/launcher=example"])],
 )
+@mark.filterwarnings(
+    "ignore:provider=example-searchpath-plugin, path=arbitrary_package.conf is not available:UserWarning"
+)
 class TestExampleLauncherIntegration(IntegrationTestSuite):
```
omry, line 44: Can you explain why all those example plugins has this warning?
example-searchpath-plugin is adding something that - as far as I know - is a valid Python module that should be available once it's installed and should not trigger any warnings.

hydra/_internal/config_loader_impl.py
```diff
@@ -266,6 +266,13 @@ def _load_configuration_impl(
             for key in cfg.hydra.job.env_copy:
                 cfg.hydra.job.env_set[key] = os.environ[key]
 
+        for source in caching_repo.get_sources():
+            if not source.available():
+                warnings.warn(
+                    category=UserWarning,
+                    message=f"provider={source.provider}, path={source.path} is not available.",
+                )
+
```
omry, line 275: This validation is better done right after the sources settles down.
Otherwise you will potentially get errors loading configs that are not found because the searchpath is incorrect - before this warning is issued.

jieru-hu:
submitit launcher test failure is fixed here https://github.com/facebookresearch/hydra/pull/1550

examples/plugins/example_searchpath_plugin/setup.py
```diff
@@ -13,7 +13,9 @@
         long_description=LONG_DESC,
         long_description_content_type="text/markdown",
         url="https://github.com/facebookresearch/hydra/",
-        packages=find_namespace_packages(include=["hydra_plugins.*"]),
+        packages=find_namespace_packages(
+            include=["hydra_plugins.*", "arbitrary_package.*"]
+        ),
```
omry, line 18: I would say that the second package is technically not a namespace package.
Something like this (if works) is the correct form:
```
packages=find_namespace_packages(include=["hydra_plugins.*"]) + find_packages(include=["arbitrary_package.*"]),
```

Now, since this plugin is rather simple, you can also just add the second package directly.

```
packages=find_namespace_packages(include=["hydra_plugins.*"]) + ["arbitrary_package"],
```
(again, need to confirm it).

jieru-hu: thanks, fixed.

tests/test_compose.py
```diff
@@ -533,6 +533,20 @@ def test_searchpath_config_errors(
...
+    ) -> None:
+        config_name = "without_sp"
+        override = "hydra.searchpath=['pkg://fakeconf']"
+        with warns(
+            expected_warning=UserWarning,
+            match=re.escape(
+                "provider=hydra.searchpath in command-line, path=fakeconf is not available."
+            ),
+        ):
+            compose(config_name=config_name, overrides=[override])
```
omry, line 548: no need for a dedicated config here. you can test it with an existing app:
```
 python  examples/tutorials/basic/your_first_hydra_app/1_simple_cli/my_app.py hydra.searchpath=['pkg://fakeconf']
```

examples/advanced/config_search_path/conf/cofig_with_no_searchpath.yaml
```diff
@@ -0,0 +1,2 @@
+defaults:
+  - dataset: cifar10
```
omry, line 2: See comment below, you can delete this one and use an existing app.

hydra/_internal/config_loader_impl.py
```diff
@@ -225,6 +225,13 @@ def _load_configuration_impl(
 
         self._process_config_searchpath(config_name, parsed_overrides, caching_repo)
 
+        for source in caching_repo.get_sources():
+            if not source.available():
+                warnings.warn(
+                    category=UserWarning,
+                    message=f"provider={source.provider}, path={source.path} is not available.",
+                )
+
```
omry, line 234: to reduce the complexity of the _load_configuration_impl, can you move this into _process_config_searchpath?

omry (changes requested):

tests/test_compose.py
jieru-hu: `without_sp` is not a new config, I've removed the extra config I added previously.

hydra/_internal/config_loader_impl.py
jieru-hu: done!

omry (approved):

tests/test_hydra.py
```diff
@@ -649,32 +649,29 @@ def test_help(
 
 
 @mark.parametrize(
-    "script,overrides,expected",
+    "overrides,expected",
     [
         param(
-            "examples/tutorials/basic/your_first_hydra_app/1_simple_cli/my_app.py",
-            ["--info", "searchpath", "hydra.searchpath=['pkg://fakeconf']"],
```
omry, line 656: This is an app without hydra.searchpath in its config.
so it's testing hydra.searchpath in the overrides only.

I think you changed the meaning of the test.

omry (changes requested):

tests/test_hydra.py
jieru-hu: Hmmm in the validation we are  making sure that the overrides comes from the command-line so I don't think the meaning is changed. 

The reason I had to update the app is because we need the package to actually exist next to the app now. 
An alternative is to add an another primary  config next to `examples/advanced/config_search_path/my_app.py` with no hydra.searchpath defined in it,  I didn't do that because: this test is focused on the output of  `--info` and we are already checking to make sure the hydra.searchpath comes from command-line override. 

thoughts?

omry: Why did you use two different apps in the first place when creating it?

omry (approved):
looks okay. you confused me by using two different apps in the first place. I thought there was some significance.

tests/test_hydra.py
jieru-hu: it was not intentional :)

PR #1545, opened 2021-04-08
[Nevergrad-plugin] Add a max failure rate parameter

### LINKED ISSUE #1511: Support multirun partial failure in Nevergrad sweeper
Context here: https://github.com/facebookresearch/hydra/issues/1377

jrapin:
## Motivation

Add a `max_failure_rate` so that failure rates within a batch below the threshold are ignored.

Yes

## Test Plan

Added a test checking that failures are captured or ignored depending on the allowed failure rate

## Related Issues and PRs

Fix #1511


plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -157,11 +159,32 @@ def sweep(self, arguments: List[str]) -> None:
...
+            to_tell: List[Tuple[ng.p.Parameter, float]] = []
             for cand, ret in zip(candidates, returns):
-                loss = direction * ret.return_value
+                if ret.status != utils.JobStatus.COMPLETED:
+                    to_tell.append((cand, float("inf")))
+                    failures += 1
+                    try:
+                        ret.return_value
+                    except HydraJobException as e:
+                        log.info(f"Returning infinity for failed experiment: {e}")
```
jieru-hu, line 175: `log.warn`?

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -157,11 +159,32 @@ def sweep(self, arguments: List[str]) -> None:
...
             # would have been nice to avoid waiting for all jobs to finish
             # aka batch size Vs steady state (launching a new job whenever one is done)
+            self.job_idx += len(returns)
+            # check job status and prepare losses
+            failures = 0
+            to_tell: List[Tuple[ng.p.Parameter, float]] = []
             for cand, ret in zip(candidates, returns):
-                loss = direction * ret.return_value
+                if ret.status != utils.JobStatus.COMPLETED:
+                    to_tell.append((cand, float("inf")))
```
jieru-hu, line 170: noob question, what if the direction is `maximization` in this case?

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
```diff
@@ -1,8 +1,10 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+import subprocess
 from pathlib import Path
 from typing import Any
 
 import nevergrad as ng
+import pytest
```
jieru-hu, line 7: could we do `from pytest import mark` instead to be consistent with the rest of the repo?

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
```diff
@@ -155,3 +156,22 @@ def test_nevergrad_example(with_commandline: bool, tmpdir: Path) -> None:
...
+        "hydra.sweeper.optim.num_workers=2",
+        f"hydra.sweeper.optim.max_failure_rate={max_failure_rate}",
+        "error=true",
+    ]
+    if max_failure_rate >= 1.0:
+        # nevergrad will warn that the values are too high (infinity)
+        run_python_script(cmd, allow_warnings=True)
+    else:
+        with pytest.raises(subprocess.CalledProcessError):
+            run_python_script(cmd)
```
jieru-hu, line 177: Could we assert the exception and the warning?

maybe something similar [here](https://github.com/facebookresearch/hydra/blob/6349706f4413c9431bd1d5bfcabb3ae041a33a4d/tests/test_basic_sweeper.py#L62) where we assert traceback in the test for BasicSweeper.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -157,11 +159,32 @@ def sweep(self, arguments: List[str]) -> None:
...
+                else:
+                    to_tell.append((cand, direction * ret.return_value))
+            # raise if too many failures
+            if failures / len(returns) > self.opt_config.max_failure_rate:
+                log.error(
+                    f"Failed {failures} times out of {len(returns)} "
+                    f"with max_failure_rate={self.opt_config.max_failure_rate}"
+                )
+                for ret in returns:
+                    ret.return_value  # delegate raising to JobReturn, with actual traceback
```
jieru-hu, line 185: What does the traceback look like here?

plugins/hydra_nevergrad_sweeper/example/config.yaml
```diff
@@ -40,3 +40,4 @@ db: cifar
 lr: 0.01
 dropout: 0.6
 batch_size: 8
+error: false
```
jieru-hu, line 43: Could we add a comment here, something like
```
# if True, manually raise exception to mimic a failed experiment run. 
```

jieru-hu (changes requested):
Thanks for the PR Jérémy!

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jrapin: nevergrad only does minimization, so for failures we only get +inf. When there is an actual `return_value` (see below), we multiply by `direction` so as to revert to minimization if need be

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
jrapin: So `from pytest import mark, raises`?
I've reverted to importing `pytest` since I am using those two functions, and I felt it would be more structured/simpler to follow this way, but I'll do what you prefer.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jrapin: Something like this:
Screen Shot 2021-04-13 at 1 30 36 PM

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
jrapin: I've updated the test, is it better ?

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
jieru-hu: Yes, pls import them. This is the [commit](https://github.com/facebookresearch/hydra/commit/9bf7800157692795090f3695efe136bbbd6fef1d) that modifies the whole repo to import the functions.

jieru-hu (commented):
Thank you. lgtm. @omry do you want to take a pass as well?

plugins/hydra_nevergrad_sweeper/tests/test_nevergrad_sweeper_plugin.py
jrapin: done, and i'm not using raises anymore anyway.
out of curiosity, what was the reason for this move? on my side I do prefer keeping the namespace because it feels clearer to me, but that's a personal choice I guess.

omry:
Yup, will take a look in after PTED21.

plugins/hydra_nevergrad_sweeper/example/config.yaml
```diff
@@ -40,3 +40,6 @@ db: cifar
 lr: 0.01
 dropout: 0.6
 batch_size: 8
+
+# if True, manually raise exception to mimic a failed experiment run.
```
omry, line 44:
```suggestion
# if true, simulate a failure by raising an exception
```

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -132,7 +134,7 @@ def sweep(self, arguments: List[str]) -> None:
             ] = create_nevergrad_parameter_from_override(override)
 
         parametrization = ng.p.Dict(**params)
-        parametrization.descriptors.deterministic_function = not self.opt_config.noisy
+        parametrization.function.deterministic = not self.opt_config.noisy
```
omry, line 137: Is this related? I thought @jieru-hu already addressed the deprecation warnings in another PR.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
omry, line 170: nit, this is a bit cleaner:
```suggestion
                    to_tell.append((cand, math.inf))
```

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -157,11 +159,32 @@ def sweep(self, arguments: List[str]) -> None:
...
             self.validate_batch_is_legal(overrides)
             returns = self.launcher.launch(overrides, initial_job_idx=self.job_idx)
-            self.job_idx += len(returns)
             # would have been nice to avoid waiting for all jobs to finish
             # aka batch size Vs steady state (launching a new job whenever one is done)
+            self.job_idx += len(returns)
+            # check job status and prepare losses
+            failures = 0
+            to_tell: List[Tuple[ng.p.Parameter, float]] = []
             for cand, ret in zip(candidates, returns):
```
omry, line 168: What is the motivation for loop twice? (once to check the error and another time to tell the optimizer?)
can both be done in the same iteration?

omry (commented):
A few nits and questions.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jrapin: yes it's been fixed in the meantime

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jrapin: yes just split the logic but probably didnt help, so I updated it

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
```diff
@@ -158,14 +161,32 @@ def sweep(self, arguments: List[str]) -> None:
...
-                    best = (loss, cand)
+                if ret.status == utils.JobStatus.COMPLETED:
+                    rectified_loss = direction * ret.return_value
+                else:
+                    rectified_loss = math.inf
+                    failures += 1
+                    try:
+                        ret.return_value
+                    except HydraJobException as e:
+                        log.warning(f"Returning infinity for failed experiment: {e}")
```
omry, line 178: Is there a scenario where this does not raise?
If yes, what is it?
if no, can you remove the try-except and just print the warning?

omry (commented):
Looking good, question/request.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jrapin: if I am correct this always raises since the job is not COMPLETED. My aim here is to capture the error so that I can log it: I tend to print as much information as I can so that it's easier to debug, being robust to the failures does not mean they are expected.
Does this makes sense to you?

omry: It does. I wonder if we should add another property to access the exception that will raise if there is no exception.

omry (approved):
lgtm.

plugins/hydra_nevergrad_sweeper/hydra_plugins/hydra_nevergrad_sweeper/_impl.py
jrapin: In concurrent.futures, jobs have an `exception()` method that returns the exception if it exists, and `None` otherwise. This way users can handle it the way they want.

omry (changes requested):
@jrapin, 
1. Any doc updates to do here?
2. Can you create a news fragment file in the plugin's news directory?

jrapin:
@omry 
> 1. Any doc updates to do here?

Slight modification, but most of the documentation is the config file itself anyway

> 2. Can you create a news fragment file in the plugin's news directory?

Added

news/1545.plugin
```diff
@@ -0,0 +1 @@
+Add fault tolerance in nevergrad sweeper through `max_failure_rate` parameter
```
omry, line 1: 1. In `plugins/hydra_nevergrad_sweeper/news/`.
2.
```suggestion
Add fault tolerance via `max_failure_rate` parameter
```

omry:
ping:
>     1. Any doc updates to do here?
>     2. Can you create a news fragment file in the plugin's news directory?

Also, there seems to be a conflict.

jrapin:
> @jrapin, ping.
> 
> > ```
> > 1. Any doc updates to do here?
> > 2. Can you create a news fragment file in the plugin's news directory?
> > ```
> 
> Also, there seems to be a conflict.

@omry conflict solved, the other points were already answered though

news/1545.plugin
```diff
@@ -0,0 +1 @@
+Add fault tolerance via `max_failure_rate` parameter
```
omry, line 1: What I mean is to move this file into the news directory inside the plugin itself.

`plugins/hydra_nevergrad_sweeper/news`.

omry:
@jrapin, almost:
the number in the file refers to the GitHub issue for the feature. File name should still be 1545.feature.

jrapin:
> @jrapin, almost:
> the number in the file refers to the GitHub issue for the feature. File name should still be 1545.feature.

@omry, sorry, I should have remembered

omry:
> > @jrapin, almost:
> > the number in the file refers to the GitHub issue for the feature. File name should still be 1545.feature.
> 
> @omry, sorry, I should have remembered

No worries.
Looks like there are some warnings that are failing the test now. I think you addressed them in nevergrad already.

jieru-hu:
ping @jrapin...  :)

jrapin:
> ping @jrapin... :)

@jieru-hu sorry for the delay, I did not want to push a new nevergrad version (solving the warning issue) right before leaving for PTO, then I forgot about this diff when I came back :s 

I think all tests pass, using latest master, and all comments have been addressed.

jieru-hu (approved):

PR #1551, opened 2021-04-13
cannot add new config group from commandline if there's override in config file.

### LINKED ISSUE #1548: [Bug] Cannot use experiment pattern when using overrides in main config file
# 🐛 Bug
When using any override in main config, e.g. for color logging or sweeper:
```yaml
defaults:
    - override hydra/hydra_logging: colorlog
    - override hydra/job_logging: colorlog
```

Then the experiment configs can't be loaded:
```bash
python main.py +experiment=experiment_name
```
Stack trace:
```
In config.yaml: Override 'hydra/job_logging : colorlog' is defined before 'experiment: experiment_name'.
Overrides must be at the end of the defaults list
```

## System information
- **Hydra Version** :  1.1.0.dev5
- **Python version** : 3.8
- **Virtual environment type and version** : conda
- **Operating system** : Ubuntu20.04

## Additional context
I believe this issue didn't exist on 1.1.0.dev4.

jieru-hu:
Closes #1548

### The bug

to repro with the examples from tutorial
```
$ git diff
diff --git a/examples/patterns/configuring_experiments/conf/config.yaml b/examples/patterns/configuring_experiments/conf/config.yaml
index 9554ce582..9077ff646 100644
--- a/examples/patterns/configuring_experiments/conf/config.yaml
+++ b/examples/patterns/configuring_experiments/conf/config.yaml
@@ -1,3 +1,4 @@
 defaults:
   - db: mysql
   - server: apache
+  - override hydra/hydra_logging: colorlog
```

Then run the application. 
```
$ python /Users/jieru/workspace/hydra-fork/hydra/examples/patterns/configuring_experiments/my_app.py +experiment=aglite
In config: Override 'hydra/hydra_logging : colorlog' is defined before 'experiment: aglite'.
Overrides must be at the end of the defaults list

```

The exception comes from [here](https://github.com/facebookresearch/hydra/blob/4d902670cf661ef26cf14adc2da936f238e3ea2c/hydra/_internal/defaults_list.py#L362-L373).  The reason being `+experiment=aglite` is added to the end of  defaults list after the overrides (in this example, after `hydra/hydra_logging: colorlog`) but is not recognized as an override.

The fix is to update the defaults list by going through the defaults list first before calling [`_update_overrides `](https://github.com/facebookresearch/hydra/blob/master/hydra/_internal/defaults_list.py#L466-L469), if `overrides.append_group_defaults` is not empty, it means there is appending overrides from the commandline, in this case, we go through the defaults list to see if override exists, if override defaults exists, we append override right before the overrides default (so it can be overriden later), we also append the append commandline override to after the defaults override and make `override=True` which won't trigger the above exception.

### DIFF
--- a/examples/patterns/configuring_experiments/conf/config_with_override.yaml
+++ b/examples/patterns/configuring_experiments/conf/config_with_override.yaml
@@ -1,0 +1,4 @@
+defaults:
+  - db: mysql
+  - server: apache
+  - override hydra/env: custom
--- a/examples/patterns/configuring_experiments/conf/config_with_override.yaml
+++ b/examples/patterns/configuring_experiments/conf/config_with_override.yaml
@@ -1,4 +1,0 @@
-defaults:
-  - db: mysql
-  - server: apache
-  - override hydra/env: custom
--- a/tests/test_examples/test_patterns.py
+++ b/tests/test_examples/test_patterns.py
@@ -99,23 +99,13 @@ def test_extending_configs(
             {"db": {"name": "sqlite"}, "server": {"name": "apache", "port": 8080}},
             id="exp1+override",
         ),
-        param(
-            [
-                "--config-name=config_with_override",
-                "+experiment=nglite",
-                "server=apache",
-            ],
-            {"db": {"name": "sqlite"}, "server": {"name": "apache", "port": 8080}},
-            id="exp1+defaults_override",
-        ),
     ],
 )
 def test_configuring_experiments(
     monkeypatch: Any, tmpdir: Path, overrides: List[str], expected: Any
 ) -> None:
     monkeypatch.chdir("examples/patterns/configuring_experiments")
-    cmd = ["my_app.py"] + overrides
-    cmd.extend([f"hydra.run.dir={tmpdir}"])
+    cmd = ["my_app.py", "hydra.run.dir=" + str(tmpdir)] + overrides
     result, _err = run_python_script(cmd)
     assert OmegaConf.create(result) == expected
 
--- a/tests/test_examples/test_patterns.py
+++ b/tests/test_examples/test_patterns.py
@@ -115,7 +115,7 @@ def test_configuring_experiments(
 ) -> None:
     monkeypatch.chdir("examples/patterns/configuring_experiments")
     cmd = ["my_app.py"] + overrides
-    cmd.extend(["hydra.run.dir=" + str(tmpdir)])
+    cmd.extend([f"hydra.run.dir={tmpdir}"])
     result, _err = run_python_script(cmd)
     assert OmegaConf.create(result) == expected
 
--- a/tests/test_examples/test_patterns.py
+++ b/tests/test_examples/test_patterns.py
@@ -99,13 +99,23 @@ def test_extending_configs(
             {"db": {"name": "sqlite"}, "server": {"name": "apache", "port": 8080}},
             id="exp1+override",
         ),
+        param(
+            [
+                "--config-name=config_with_override",
+                "+experiment=nglite",
+                "server=apache",
+            ],
+            {"db": {"name": "sqlite"}, "server": {"name": "apache", "port": 8080}},
+            id="exp1+defaults_override",
+        ),
     ],
 )
 def test_configuring_experiments(
     monkeypatch: Any, tmpdir: Path, overrides: List[str], expected: Any
 ) -> None:
     monkeypatch.chdir("examples/patterns/configuring_experiments")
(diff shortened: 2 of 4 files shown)


omry:
1. tests should come first.
2. The implementation looks a bit cumbersome, I did not take a close look at the issue yet so it's hard for me to comment further.

Once there are tests, it will be easier to iterate on it.

jieru-hu:
tests added :)

tests/test_examples/test_patterns.py
```diff
@@ -99,13 +99,23 @@ def test_extending_configs(
...
+        ),
     ],
 )
 def test_configuring_experiments(
     monkeypatch: Any, tmpdir: Path, overrides: List[str], expected: Any
 ) -> None:
     monkeypatch.chdir("examples/patterns/configuring_experiments")
-    cmd = ["my_app.py", "hydra.run.dir=" + str(tmpdir)] + overrides
+    cmd = ["my_app.py"] + overrides
+    cmd.extend(["hydra.run.dir=" + str(tmpdir)])
```
omry, line 118:
```suggestion
    cmd.extend([f"hydra.run.dir={tmpdir}")])
```

omry:
Can you explain the problem and the fix in words?

omry (changes requested):
see comments.

omry:
> Can you explain the problem and the fix in words?

Still waiting for this.

jieru-hu:
> > Can you explain the problem and the fix in words?
> 
> Still waiting for this.

Yes, I've updated the PR description. Should have respond to this message to call your attention to that :)

omry:
Try this as an alternative solution (instead of your first commit).
The idea: when appending the command line overrides to the primary defaults list - insert them before the first override entry.
This seems simpler than you approach and also solves the problem.

```diff
diff --git a/hydra/_internal/defaults_list.py b/hydra/_internal/defaults_list.py
index 7ee9cb389..f553d8f66 100644
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -464,7 +464,13 @@ def _create_defaults_tree_impl(
         self_added = _validate_self(containing_node=parent, defaults=defaults_list)
 
     if is_root_config:
-        defaults_list.extend(overrides.append_group_defaults)
+        # To ensure config overrides are last, insert the external overrides before the first config override.
+        insert_idx = len(defaults_list)
+        for idx, default in enumerate(defaults_list):
+            if default.is_override():
+                insert_idx = idx
+                break
+        defaults_list[insert_idx:insert_idx] = overrides.append_group_defaults
 
     _update_overrides(defaults_list, overrides, parent, interpolated_subtree)
 

```

omry (changes requested):
see comments.

jieru-hu:
> Try this as an alternative solution (instead of your first commit).
> The idea: when appending the command line overrides to the primary defaults list - insert them before the first override entry.
> This seems simpler than you approach and also solves the problem.
> 
> ```diff
> diff --git a/hydra/_internal/defaults_list.py b/hydra/_internal/defaults_list.py
> index 7ee9cb389..f553d8f66 100644
> --- a/hydra/_internal/defaults_list.py
> +++ b/hydra/_internal/defaults_list.py
> @@ -464,7 +464,13 @@ def _create_defaults_tree_impl(
>          self_added = _validate_self(containing_node=parent, defaults=defaults_list)
>  
>      if is_root_config:
> -        defaults_list.extend(overrides.append_group_defaults)
> +        # To ensure config overrides are last, insert the external overrides before the first config override.
> +        insert_idx = len(defaults_list)
> +        for idx, default in enumerate(defaults_list):
> +            if default.is_override():
> +                insert_idx = idx
> +                break
> +        defaults_list[insert_idx:insert_idx] = overrides.append_group_defaults
>  
>      _update_overrides(defaults_list, overrides, parent, interpolated_subtree)
>  
> ```

Very nice!

examples/patterns/configuring_experiments/conf/config_with_override.yaml
```diff
@@ -0,0 +1,4 @@
+defaults:
+  - db: mysql
+  - server: apache
+  - override hydra/env: custom
```
omry, line 4: The default list tests are sufficient.
We should avoid complicating examples for the purpose of testing.
Can you undo the patterns changes?

omry (changes requested):

examples/patterns/configuring_experiments/conf/config_with_override.yaml
jieru-hu: sounds good, removed.

omry (approved):
shipit

PR #1570, opened 2021-04-23
Updated release process to use python -m build

omry:
Switching to build with `python -m build`.

1. clean no longer cleaning the dist directory. this is conflicting with python -m build.
2. include the `build` package in the development deps.
3. updated developer guild to reflect change + some other tweaks


omry:
@jieru-hu, @shagunsodhani  - fyi.

PR #1580, opened 2021-04-28
Move Callbacks to hydra._internal.utils

jieru-hu:
initially `Callbacks` was added to `hydra/core/callbacks.py`. However this creates an unnecessary circular dependency between 
`hydra/core/callbacks.py` and `hydra/core/utils.py`. 

This PR is to move Callbacks to `hydra/core/utils.py`, which seems like a more suitable place (it is a util class after all :) )


omry (commented):
Is user/plugin code ever interacting with Callbacks directly?
if not, can it be in hydra._internal.utils?

jieru-hu:
> Is user/plugin code ever interacting with Callbacks directly?
> if not, can it be in hydra._internal.utils?

`Callbacks` also [references](https://github.com/facebookresearch/hydra/blob/d01387e4e4e1340f2fec2d9b8f964ab5d6fed545/hydra/core/callbacks.py#L44-L46) [`JobReturn`](https://github.com/facebookresearch/hydra/blob/d01387e4e4e1340f2fec2d9b8f964ab5d6fed545/hydra/core/utils.py#L192)
which is also in hydra.core.utils. I think `hydra.core.utils` is probably a better place.

omry:
Is this unavoidable? I would like to avoid putting internal classes in user facing modules.

omry:
Callbacks only need JobReturn for type safety reasons.
Maybe this can help:
https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

jieru-hu:
> Callbacks only need JobReturn for type safety reasons.
> Maybe this can help:
> https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

TIL :) Thanks!

omry:
I still think it's a good idea to move the Callbacks class to _internal for the reasons I stated.

jieru-hu:
> I still think it's a good idea to move the Callbacks class to _internal for the reasons I stated.

Sounds good. I've updated the PR accordingly.

omry (commented):
What I meant was to move the callback.py file it to hydra._internal.
Any reason to to move the content to internal utils?

jieru-hu:
> What I meant was to move the callback.py file it to hydra._internal.
> Any reason to to move the content to internal utils?

oh ok. In your previous comment you suggested `hydra._internal.utils` :) 

>Is user/plugin code ever interacting with Callbacks directly?
if not, can it be in hydra._internal.utils?

omry:
> > What I meant was to move the callback.py file it to hydra._internal.
> > Any reason to to move the content to internal utils?
> 
> oh ok. In your previous comment you suggested `hydra._internal.utils` :)
> I can move the `callback.py` then.

I think this was before you found out that you can use TYPE_CHECKING to eliminate the circular dependency.

omry (approved):
looking good.

PR #1595, opened 2021-04-30
docs update

### LINKED ISSUE #1579: [Doc] Improve documentation for the deprecation of matching config name
The doc page https://hydra.cc/docs/next/upgrades/1.0_to_1.1/automatic_schema_matching/ explains the deprecation that is happening in Hydra 1.1, where a warning is issued (redirecting to that doc page) when a config file is matched based on its name.

There are two issues in this doc page:
1. A minor grammatical issue with the sentence "Store use a different name when storing the schema into the config store"
2. It only describes clearly the fix that consists in keeping the same filename and renaming the entry in the `cs.store()` statement. However, there is a secondary approach that people can use, and that is actually mandatory in cases where the `cs.store()` statement can't be modified (e.g., with plugins): renaming the file instead. This secondary approach should be documented with an example, also mentioning that if we rename the file then we also need to update other references to it (two typical examples worth mentioning are (i) on the command line, and (ii) in the defaults list)

omry:
1. Closes #1579 
2. Some other minor doc changes

Preview (see netlify link in checks):

https://deploy-preview-1595--hydra-preview.netlify.app/docs/next/upgrades/intro

(Multiple pages are changed by this PR).

### DIFF
--- a/website/docs/upgrades/1.0_to_1.1/automatic_schema_matching.md
+++ b/website/docs/upgrades/1.0_to_1.1/automatic_schema_matching.md
@@ -107,7 +107,7 @@ Steps:
 3. Update references to the config name accordingly, e.g. on the command-line `db=mysql` would become `db=custom_mysql`, and in a defaults list `db: mysql` would become `db: custom_mysql`.
 
 
-<details><summary>Click to expand</summary>
+<details><summary>Click to show an example</summary>
 
 #### Hydra 1.0
 <div className="row">
--- a/website/docs/upgrades/1.0_to_1.1/automatic_schema_matching.md
+++ b/website/docs/upgrades/1.0_to_1.1/automatic_schema_matching.md
@@ -4,29 +4,35 @@ title: Automatic schema-matching
 hide_title: true
 ---
 
-In Hydra 1.0, when config file is loaded, if a config with a matching name and group is present in the `ConfigStore`,
+In Hydra 1.0, when a config file is loaded, if a config with a matching name and group is present in the `ConfigStore`,
 it is used as the schema for the newly loaded config.
 
 There are several problems with this approach:
 
 - **Inflexible**: This approach can only be used when a schema should validate a single config file.
 It does not work if you want to use the same schema to validate multiple config files.
-- **Surprising**: This hidden behavior can be surprising. There is no way to tell this is going to happen when looking at a given
+- **Unexpected**: This behavior can be unexpected. There is no way to tell this is going to happen when looking at a given
 config file.
 
 Hydra 1.1 deprecates this behavior in favor of an explicit config extension via the Defaults List.  
-This upgrade page aims to provide the TLDR. It is highly recommended that you read the following pages:
+This upgrade page aims to provide summary of the required changes. It is highly recommended that you read the following pages:
 - [Background: The Defaults List](../../advanced/defaults_list.md)
 - [Background: Extending configs](../../patterns/extending_configs.md)
 - [Tutorial: Structured config schema](../../tutorials/structured_config/5_schema.md)
 
-### Migration
-The migration involves two steps for each config file.
-1. Store use a different name when storing the schema into the config store. Common choices:
+Before the upgrade, you have two different configs with the same name (a config file, and a Structured Config in the `ConfigStore`).
+You need to rename one of them. Depending on the circumstances and your preference you may rename one or the other.
+- If you control both configs, you can rename either of them. 
+- If you only control the config file, rename it.
+
+### Rename the Structured Config
+Steps:
+1. Use a different name when storing the schema into the Config Store. Common choices:
    - `base_` prefix, e.g. `base_mysql`.
    - `_schema` suffix, e.g. `mysql_schema`.
-2. Add the schema to the Defaults List in the config file.
-    
+2. Add the schema to the Defaults List of the extending config file.
+
+<details><summary>Click to expand</summary>
 
 #### Hydra 1.0
 <div className="row">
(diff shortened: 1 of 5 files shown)


website/docs/upgrades/1.0_to_1.1/automatic_schema_matching.md
```diff
@@ -92,3 +98,78 @@ cs.store(group="db",
...
+         node=MySQLConfig)
+```
+</div>
+</div>
+
+#### Hydra 1.1
+<div className="row">
+<div className="col col--6">
+
+```yaml title="db/custom_mysql.yaml" {1,2}
```
odelalleau, line 147: It would be good to highlight the new name. I'm not sure it can be done through Docusaurus, but something like this could work:
```suggestion
```yaml title="db/custom_mysql.yaml <= NEW NAME!" {1,2}
```

website/docs/upgrades/1.0_to_1.1/automatic_schema_matching.md
```diff
@@ -92,3 +98,78 @@ cs.store(group="db",
 </div>
 </div>
 
+</details>
+
+### Rename the config file
+Steps:
+1. Rename the config file. Common choices are `custom_` or `my_` prefix, e.g. `custom_mysql.yaml`.
+2. Add the schema to the Defaults List of the extending config file.
```
odelalleau, line 106: This makes it look like things will work as before after performing these 2 steps. But that may not be the case => I think it would be important to add a "step 3" along the lines of:
```
3. Update references to the old file name accordingly, e.g. on the command-line `db=mysql`
   would become `db=custom_mysql`, and in a defaults list `db: mysql` would become
   `db: custom_mysql` (see below for an example)
```
Then in the example below, replace the config store code block (which doesn't change anyway) with a config file that has `db: mysql` in its defaults list.

website/docs/upgrades/1.0_to_1.1/hydra_main_config_path.md
```diff
@@ -5,25 +5,24 @@ title: Changes to @hydra.main() and hydra.initialize()
...
 ```python
 @hydra.main(config_path="conf")
 # or:
 hydra.initialize(config_path="conf")
 ```
 
 ### No config directory
-Do not add any directory to the config searchpath.
-Recommended if you do not want to use configs files defined next to your Python script (Typically applications using only Structured Configs).
+For applications that do not define config files next to the Python script (Typically applications using only Structured Configs), it is recommended that
```
odelalleau, line 24: T -> t:
```suggestion
For applications that do not define config files next to the Python script (typically applications using only Structured Configs), it is recommended that
```

website/docs/upgrades/1.0_to_1.1/hydra_main_config_path.md
```diff
@@ -34,7 +33,7 @@ hydra.initialize(config_path=None)
 ### Using the application directory
 Use the directory/module of the Python script.
 This was the default behavior up Hydra 1.0.
```
odelalleau, line 35:
```suggestion
This was the default behavior up to Hydra 1.0.  
```

website/docs/upgrades/intro.md
```diff
@@ -0,0 +1,29 @@
+---
+id: intro
+title: Introduction
+sidebar_label: Introduction
+---
+
+Upgrading to a new Hydra version is usually an easy process.
+
+## Major version upgrades
+Hydra will typically provide a helpful warnings about required changes, sometimes pointing to an upgrade page that provides more details about the required changes.
```
odelalleau, line 10:
```suggestion
Hydra will typically provide helpful warnings about required changes, sometimes pointing to an upgrade page that provides more details about the required changes.
```

website/docs/upgrades/intro.md
```diff
@@ -0,0 +1,29 @@
+---
+id: intro
+title: Introduction
+sidebar_label: Introduction
+---
+
+Upgrading to a new Hydra version is usually an easy process.
+
+## Major version upgrades
```
odelalleau, line 9: Hydra uses an unconventional definition of what is "major" vs "minor" versions => it may be worth clarifying what is meant by "major" here.

odelalleau (changes requested):

website/docs/upgrades/1.0_to_1.1/automatic_schema_matching.md
```diff
@@ -4,29 +4,35 @@ title: Automatic schema-matching
...
+### Rename the Structured Config
+Steps:
+1. Use a different name when storing the schema into the Config Store. Common choices:
    - `base_` prefix, e.g. `base_mysql`.
    - `_schema` suffix, e.g. `mysql_schema`.
-2. Add the schema to the Defaults List in the config file.
-    
+2. Add the schema to the Defaults List of the extending config file.
+
+<details><summary>Click to expand</summary>
```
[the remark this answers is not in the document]
omry: It's too easy to lose sight of the high level structure of the examples are inlined.

website/docs/upgrades/1.0_to_1.1/automatic_schema_matching.md
omry: Good catch.

website/docs/upgrades/intro.md
omry: Yes. I am wondering if I want to switch to a more traditional semver, but in truth I am only using two version types now, major and patch.

odelalleau (changes requested):
Just a minor change for consistency, then good to go!
(I had also suggested to modify the "Option 2" example to show the renaming of a reference in a defaults list, but I don't think it's absolutely necessary)

odelalleau (approved):
Seems clear enough to me now! :)

omry:
> Just a minor change for consistency, then good to go!
> (I had also suggested to modify the "Option 2" example to show the renaming of a reference in a defaults list, but I don't think it's absolutely necessary)

I think it will be a bit too dense (we can also add a command line example)
We are calling out that you need to do it in the third step already so I think it's enough.

omry:
@odelalleau, gave it a shot (along with some other updates).
let me know what you think.

omry:
image

odelalleau (approved):
Awesome!

omry:
Thanks for the feedback!

PR #1598, opened 2021-05-03
Override grammar: allow question marks in unquoted strings

### LINKED ISSUE #1597: Override grammar: allow question marks in unquoted strings
# 🚀 Feature Request

Question marks should be allowed in unquoted strings so as to match OmegaConf's grammar (see the corresponding PR in OmegaConf: https://github.com/omry/omegaconf/pull/638)

Edit: I am working on it

odelalleau:
Fixes #1597

### DIFF
--- a/news/1597.feature
+++ b/news/1597.feature
@@ -1,0 +1,1 @@
+Override grammar: question marks are now allowed in unquoted strings (e.g. "key=abc?def")
--- a/tests/test_overrides_parser.py
+++ b/tests/test_overrides_parser.py
@@ -31,6 +31,8 @@ from hydra.core.override_parser.types import (
 )
 from hydra.errors import HydraException
 
+UNQUOTED_SPECIAL = r"/-\+.$%*@?"  # special characters allowed in unquoted strings
+
 parser = OverridesParser(create_functions())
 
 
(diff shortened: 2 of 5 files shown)


tests/test_overrides_parser.py
```diff
@@ -993,7 +1000,11 @@ def test_override_get_value_element_method(
         param("key={123id: 0}", {"123id": 0}, id="dict_key_int_plus_id"),
         param("key={' abc ': 0}", {" abc ": 0}, id="dict_key_quoted_single"),
         param('key={" abc ": 0}', {" abc ": 0}, id="dict_key_quoted_double"),
-        param("key={a/-\\+.$%*@: 0}", {"a/-\\+.$%*@": 0}, id="dict_key_noquote"),
+        param(
+            "key={a%s: 0}" % UNQUOTED_SPECIAL,
+            {f"a{UNQUOTED_SPECIAL}": 0},
+            id="dict_key_noquote",
+        ),
```
omry, line 1007: I think the leading a is accidental here.
not sure what value it's adding here. I vote to remove it (and in that case, is this test case already covered?)

news/1597.feature
```diff
@@ -0,0 +1 @@
+Override grammar: question marks are now allowed in unquoted strings (e.g. "key=abc?def")
```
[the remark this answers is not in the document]
odelalleau: Yeah, I figured it should be fine since my example has quotes -- do you prefer me to change it or remove it?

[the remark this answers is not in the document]
omry: no, I think it's fine.

tests/test_overrides_parser.py
odelalleau: I removed it in 4f89e3889 (also rebased on master to fix a conflict)

There is a test that is very similar in `test_dict_container()` but it doesn't test the same parser rule and doesn't use the `.value()` method (which seems to be the purpose here).

tests/test_overrides_parser.py
```diff
@@ -955,8 +961,9 @@ def test_get_key_element(override: str, expected: str) -> None:
         param("key={a:10,b:20}", "{a: 10, b: 20}", True, id="dict"),
         param("key={a:10,b:[1,2,3]}", "{a: 10, b: [1, 2, 3]}", True, id="dict"),
         param(
-            "key={/-\\+.$%*@: 1}",
-            "{/-\\\\+.$%*@: 1}",  # note that \ gets escaped
+            "key={%s: 1}" % UNQUOTED_SPECIAL,
+            # Note that \ gets escaped.
+            "{%s: 1}" % UNQUOTED_SPECIAL.replace("\\", "\\\\"),
```
omry, line 966: Can you explain what's going on here?

odelalleau: > Can you explain what's going on here?

The test is still the same as before, I just wrote it based on`UNQUOTED_SPECIAL` (the motivation for introducing `UNQUOTED_SPECIAL` was to avoid having to modify multiple tests every time we'd add support for a new character in unquoted strings).

If you're asking why `\` gets escaped, this is because `_get_value_element_as_str()` calls `escape_special_characters()`.

omry (approved):
shipit

PR #1599, opened 2021-05-03
Quoted values are not valid dictionary keys anymore

odelalleau:
This is to match OmegaConf's behavior as of https://github.com/omry/omegaconf/pull/695

There is no corresponding issue / news item for Hydra since quoted dictionary keys were not allowed in Hydra 1.0.


omry (approved):
looking good.

PR #1601, opened 2021-05-03
Fix escaping in quoted values

### LINKED ISSUE #1600: [Bug] Override grammar: quoted strings ending with a trailing backslash may trigger parsing errors
# 🐛 Bug
## Description

The following override examples cause exceptions (note: there is escaping from the shell, so when using `\\` below, Hydra only sees `\`, and escaped quotes are also replaced, e.g. `\"` -> `"`):
```bash
python my_app.py "+foo=['#\\', 'x']"  # case 1a
python my_app.py "+foo={a: '#\\', b: 'x'}"  # case 2a
python my_app.py "+foo=['#\\\\', 'x']"  # case 3a
python my_app.py "+foo={a: '#\\\\', b: 'x'}"  # case 4a
```

Note that if the second string is using double quotes then there is no exception:
```bash
python my_app.py "+foo=['#\\', \"x\"]"  # case 1b
python my_app.py "+foo={a: '#\\', b: \"x\"}"  # case 2b
python my_app.py "+foo=['#\\\\', \"x\"]"  # case 3b
python my_app.py "+foo={a: '#\\\\', b: \"x\"}"  # case 4b
```

## Expected behavior

The fix to this issue should make it so that:
* Using single or double quotes in the second string ('x' vs "x") yields the same result (i.e., cases Xa and Xb should be similar)
* Cases 1 and 2 trigger an exception
* Cases 3 and 4 run without exception, with the first string containing characters `#\` (i.e., with a single backslash), compared to the current behavior where 3b and 4b yield `#\\`

## Migration instructions

If a working string in Hydra 1.0 triggers an exception or behaves unexpectedly in Hydra 1.1, this means some `\` need to be properly escaped. Here are two examples (where again, `\` are doubled because of shell escaping):
```
(Hydra 1.0)                           -> (Hydra 1.1)
python my_app.py "+foo='abc\\'"       -> python my_app.py "+foo='abc\\\\'"        # abc\
python my_app.py "+foo='abc\\\\'def'" -> python my_app.py "+foo='abc\\\\\\'def'"  # abc\'def
```

Yes I will.

## Additional context

This is caused by the definition of `QUOTED_VALUE` in the grammar, similar to OmegaConf's issue https://github.com/omry/omegaconf/issues/617

odelalleau:
Fixes #1600

@omry I realize you would prefer something simpler, but since I had it ready I thought it may still be useful to put it out so as to help ground discussions (we can keep the discussion in #1594 or move it here if you prefer)

Technically I should be able to replace the lexer mode approach from this PR with the simpler regex you mentioned in https://github.com/facebookresearch/hydra/issues/1600#issuecomment-831456040, but then I'll hit a problem later with something like `"key='abc ${foo:'def'}'"` when trying to support more interpolation types from OmegaConf (including quoted strings, dictionaries, nested interpolations...).

### DIFF
--- a/hydra/core/override_parser/overrides_visitor.py
+++ b/hydra/core/override_parser/overrides_visitor.py
@@ -309,13 +309,16 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
         return ret
 
     def _unescape_quoted_string(self, text: str) -> str:
-        """
-        Unescape a quoted string, by looking at backslashes that either:
-            * precede a quote, or
-            * end the string content
+        r"""
+        Unescape a quoted string, by looking at \ that precede a quote.
 
         The input string should contain enclosing quotes, which are stripped away
         by this function.
+
+        Due to the grammar definition of quoted strings, it is assumed that:
+            * if there are \ preceding the closing quote, their number must be even
+            * if there are \ preceding a quote in the middle of the string, their
+              number must be odd
 
         Examples (with double quotes, but the same logic applies to single quotes):
             * "abc\"def"    -> abc"def
@@ -337,7 +340,7 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
             # Add characters before the escaped sequence.
             tokens.append(text[0:start])
             # Un-escaping. Note that this works both for escaped quotes in the middle of
-            # a string, as well as trailing backslashes:
+            # a string, as well as trailing backslashes where the end quote is stripped:
             #   \"    -> "  (escaped quote in the middle)
             #   \\"   -> \  (escaped trailing backslash)
             #   \\\"  -> \" (escaped backslash followed by escaped quote in the middle)
@@ -348,9 +351,9 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
             text = text[stop:]
             match = pattern.search(text)
 
-        if text:
+        if len(text) > 1:
             # Add characters after the last match, removing the end quote.
-            tokens.append(text[:-1])
+            tokens.append(text[0:-1])
 
         return "".join(tokens)
 
--- a/hydra/core/override_parser/overrides_visitor.py
+++ b/hydra/core/override_parser/overrides_visitor.py
@@ -8,6 +8,7 @@ from antlr4.error.ErrorListener import ErrorListener
 from antlr4.tree.Tree import TerminalNodeImpl
 
 from hydra._internal.grammar.functions import FunctionCall, Functions
+from hydra._internal.grammar.utils import _ESC_QUOTED_STR
 from hydra.core.override_parser.types import (
     ChoiceSweep,
     Glob,
@@ -277,16 +278,13 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
             if node.symbol.type == OverrideLexer.QUOTED_VALUE:
                 text = node.getText()
                 qc = text[0]
-                text = text[1:-1]
                 if qc == "'":
                     quote = Quote.single
-                    text = text.replace("\\'", "'")
                 elif qc == '"':
                     quote = Quote.double
-                    text = text.replace('\\"', '"')
                 else:
                     assert False
-                text = text.replace("\\\\", "\\")
+                text = self._unescape_quoted_string(text)
                 return QuotedString(text=text, quote=quote)
             elif node.symbol.type in (OverrideLexer.ID, OverrideLexer.INTERPOLATION):
                 ret = node.symbol.text
@@ -309,6 +307,52 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
             else:
                 return node.getText()  # type: ignore
         return ret
+
+    def _unescape_quoted_string(self, text: str) -> str:
+        """
+        Unescape a quoted string, by looking at backslashes that either:
+            * precede a quote, or
+            * end the string content
+
+        The input string should contain enclosing quotes, which are stripped away
+        by this function.
+
+        Examples (with double quotes, but the same logic applies to single quotes):
+            * "abc\"def"    -> abc"def
+            * "abc\\\"def"  -> abc\"def
+            * "abc\\"       -> abc\
+            * "abc\\\\"     -> abc\\"
+        """
+        qc = text[0]  # quote character
+        text = text[1:]  # remove first quote *but* keep the last one
+        pattern = _ESC_QUOTED_STR[qc]
+        match = pattern.search(text)
+
+        if match is None:
+            return text[0:-1]  # remove last quote
+
+        tokens = []
+        while match is not None:
+            start, stop = match.span()
+            # Add characters before the escaped sequence.
+            tokens.append(text[0:start])
+            # Un-escaping. Note that this works both for escaped quotes in the middle of
+            # a string, as well as trailing backslashes:
+            #   \"    -> "  (escaped quote in the middle)
+            #   \\"   -> \  (escaped trailing backslash)
+            #   \\\"  -> \" (escaped backslash followed by escaped quote in the middle)
+            #   \\\\" -> \\ (two escaped trailing backslashes)
+            #   ...
+            tokens.append(text[start + 1 : stop : 2])
+            # Move on to next match.
+            text = text[stop:]
+            match = pattern.search(text)
+
+        if text:
+            # Add characters after the last match, removing the end quote.
+            tokens.append(text[:-1])
+
+        return "".join(tokens)
 
 
 class HydraErrorListener(ErrorListener):  # type: ignore
--- a/hydra/core/override_parser/overrides_visitor.py
+++ b/hydra/core/override_parser/overrides_visitor.py
@@ -240,42 +240,6 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
                 f"{type(e).__name__} while evaluating '{ctx.getText()}': {e}"
             ) from e
 
-    def visitQuotedValue(self, ctx: OverrideParser.QuotedValueContext) -> QuotedString:
-        children = list(ctx.getChildren())
-        assert len(children) >= 2
-
-        # Identity quote type.
-        first_quote = children[0].getText()
-        if first_quote == "'":
-            quote = Quote.single
-        else:
-            assert first_quote == '"'
-            quote = Quote.double
-
-        # Inspect string content.
-        tokens = []
-        is_interpolation = False
-        for child in children[1:-1]:
-            assert isinstance(child, TerminalNode)
-            symbol = child.symbol
-            text = symbol.text
-            if symbol.type == OverrideLexer.ESC_QUOTE:
-                # Always un-escape quotes.
-                text = text[1]
-            elif symbol.type == OverrideLexer.INTERPOLATION:
-                is_interpolation = True
-            tokens.append(text)
-
-        # Contactenate string fragments.
-        ret = "".join(tokens)
-
-        # If it is an interpolation, then OmegaConf will take care of un-escaping
-        # the `\\`. But if it is not, then we need to do it here.
-        if not is_interpolation:
-            ret = ret.replace("\\\\", "\\")
-
-        return QuotedString(text=ret, quote=quote, esc_backslash=not is_interpolation)
-
     def _createPrimitive(
         self, ctx: ParserRuleContext
     ) -> Optional[Union[QuotedString, int, bool, float, str]]:
@@ -310,8 +274,20 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
             ret = "".join(tokens)
         else:
             node = ctx.getChild(first_idx)
-            if isinstance(node, OverrideParser.QuotedValueContext):
-                return self.visitQuotedValue(node)
+            if node.symbol.type == OverrideLexer.QUOTED_VALUE:
+                text = node.getText()
+                qc = text[0]
+                text = text[1:-1]
+                if qc == "'":
+                    quote = Quote.single
+                    text = text.replace("\\'", "'")
+                elif qc == '"':
+                    quote = Quote.double
+                    text = text.replace('\\"', '"')
+                else:
+                    assert False
+                text = text.replace("\\\\", "\\")
+                return QuotedString(text=text, quote=quote)
             elif node.symbol.type in (OverrideLexer.ID, OverrideLexer.INTERPOLATION):
                 ret = node.symbol.text
             elif node.symbol.type == OverrideLexer.INT:
--- a/hydra/core/override_parser/overrides_visitor.py
+++ b/hydra/core/override_parser/overrides_visitor.py
@@ -240,6 +240,42 @@ class HydraOverrideVisitor(OverrideParserVisitor):  # type: ignore
                 f"{type(e).__name__} while evaluating '{ctx.getText()}': {e}"
             ) from e
 
+    def visitQuotedValue(self, ctx: OverrideParser.QuotedValueContext) -> QuotedString:
+        children = list(ctx.getChildren())
+        assert len(children) >= 2
+
+        # Identity quote type.
+        first_quote = children[0].getText()
+        if first_quote == "'":
+            quote = Quote.single
+        else:
+            assert first_quote == '"'
+            quote = Quote.double
+
+        # Inspect string content.
+        tokens = []
+        is_interpolation = False
+        for child in children[1:-1]:
+            assert isinstance(child, TerminalNode)
+            symbol = child.symbol
+            text = symbol.text
+            if symbol.type == OverrideLexer.ESC_QUOTE:
+                # Always un-escape quotes.
+                text = text[1]
+            elif symbol.type == OverrideLexer.INTERPOLATION:
+                is_interpolation = True
+            tokens.append(text)
+
+        # Contactenate string fragments.
+        ret = "".join(tokens)
+
+        # If it is an interpolation, then OmegaConf will take care of un-escaping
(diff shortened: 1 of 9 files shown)


omry:
Thanks for putting it up.
I am not strictly against something more complicated that can support more things, but I want to make sure that we know what we are walking into.

I just [commented](https://github.com/facebookresearch/hydra/pull/1594#issuecomment-831546615) on 1594.
I don't mind moving the discussion here.

omry:
A few additional points to consider:
1. If we eventually support function syntax in custom resolvers : `concat(a,b)` as opposed to `concat:a,b`, this might interact with the function support in Hydra overrides (`glob(*, exclude=abc)`). keeping Hydra unaware of the syntax of OmegaConf by making it only supported in quoted strings side steps this potential issue).
2. we need to ensure that things are working correctly when sweeping over quotes strings (in their different forms).

odelalleau:
> Thanks for putting it up.
> I am not strictly against something more complicated that can support more things, but I want to make sure that we know what we are walking into.
> 
> I just [commented](https://github.com/facebookresearch/hydra/pull/1594#issuecomment-831546615) on 1594.

I'll update this PR with the suggestion of relying on "quoted interpolations" instead, so we can see the diff.

> A few additional points to consider:
> 
> 1. If we eventually support function syntax in custom resolvers : `concat(a,b)` as opposed to `concat:a,b`, this might interact with the function support in Hydra overrides (`glob(*, exclude=abc)`). keeping Hydra unaware of the syntax of OmegaConf by making it only supported in quoted strings side steps this potential issue).
> 2. we need to ensure that things are working correctly when sweeping over quotes strings (in their different forms).

1. The follow-up diff I had to wrap up the support of more complex interpolations was actually very "naive" regarding what is inside interpolations, essentially ignoring everything except braces `{}` and quotes `'"`, so this wouldn't have been an issue.
2. Is there any test I should add? I'm not sure I see what kind of situation could cause problems, as long as we know when an interpolation starts and ends (or, in the upcoming new version, when a quoted string starts and ends)

odelalleau:
I pushed 37d096ea8 which uses the regex-based version of `QUOTED_VALUE`. We lost the simpler escaping when there are both `\` and interpolations in a string, but as discussed, it may not be worth the effort.

You can check this commit's diff to see what we gain/lose vs the previous lexer-based version, and the overall PR's diff for comparison to master. I think it's essentially ready to be merged (pending review), except for the missing doc & news updates (will look into it now).

odelalleau:
> except for the missing doc & news updates

Added in ffeadb183, should be good to go for review!

tests/test_overrides_parser.py
```diff
@@ -561,15 +593,41 @@ def test_key(value: str, expected: Any) -> None:
...
-            '"foo \\"bar"',
+            r'"foo \"bar"',
             QuotedString(text='foo "bar', quote=Quote.double),
             id="value:escape_double_quote",
         ),
+        param(
+            r"'foo \\\'bar'",
+            QuotedString(text=r"foo \'bar", quote=Quote.single),
+            id="value:escape_single_quote_x3",
+        ),
```
omry, line 609: I understand that this is why you added unescaping of double backslash in _createPrimitive().
Unescaping seems like a big hammer.
Can you remind me why we can't represent `QuotedString(text=r"foo \"bar", quote=Quote.double),` with `r"'foo \\'bar'"`?

no unescaping of `\\`, only of `\'` -> `'`.

omry: Reading your doc change:

> which can be required e.g. for strings ending with `\`

If I understand things correctly, this looks like:
We unescaping all the double backslashes (essentially requiring double number of backslashes if downstream (OmegaConf also needs more than one backslash), and in exchange we get support for the edge case of trailing backslash (which is ambiguous with an escaped quote).

Unescaping double backslashes everywhere in a quoted string seems like a pretty big hammer.
Can we limit the effect somehow to when we really need it?

What if we had special handling for the case where a string ends with a double backslash and only unescape it in that case?
It's a bit tricky, and we will also need to apply the reverse logic when quoting a string (`with_quotes`) but I think it might be possible.

omry: @odelalleau  I think it should be possible:

```python
from pytest import mark


def forward(s: str) -> str:
    if s.endswith("\\\\"):
        s1 = s[0:-2].replace(r"\'", "'") + "\\"
    else:
        s1 = s.replace(r"\'", "'")
    return s1


def backward(s):
    if s.endswith("\\"):
        s1 = s.replace("'", r"\'") + "\\"
    else:
        s1 = s.replace("'", r"\'")
    return s1


@mark.parametrize(
    ("input_", "expected"),
    [
        (r"ab", r"ab"),
        (r"a\'b", r"a'b"),
        (r"ab\'", "ab'"),
        (r"\ab", r"\ab"),
        (r"ab\\ ".strip(), r"ab\ ".strip()),
        (r"ab\\'", r"ab\'"),
    ],
)
class TestEscape:
    def test_forward(self, input_, expected):
        res = forward(input_)
        assert res == expected

    def test_backward(self, input_, expected):
        res = forward(input_)
        assert res == expected
        back = backward(res)
        assert back == input_
```

odelalleau: > Can you remind me why we can't represent `QuotedString(text=r"foo \"bar", quote=Quote.double),` with `r"'foo \\'bar'"`?

That's because the lexer will see a QUOTED_VALUE matching `'foo \\'` (meant to be resolved to `foo \`) and then an exception will be raised because it won't know what to do with the remaining characters `bar'` (same reason why `x = 'foo \\'bar'` in Python raises a SyntaxError).

> We unescaping all the double backslashes (essentially requiring double number of backslashes if downstream (OmegaConf also needs more than one backslash), and in exchange we get support for the edge case of trailing backslash (which is ambiguous with an escaped quote).

Essentially, yes.

> Unescaping double backslashes everywhere in a quoted string seems like a pretty big hammer.
> Can we limit the effect somehow to when we really need it?

Yes, we can limit ourselves to the `\\...\'` case (i.e., an arbitrary number of backslashes followed by a quote). This is similar to what I had in https://github.com/omry/omegaconf/pull/621 (see examples in this PR's description) -- I think if we do it in Hydra it would also make sense to do the same in OmegaConf for consistency.

> What if we had special handling for the case where a string ends with a double backslash and only unescape it in that case?

It would need to be generalized to the more general case I just mentioned but yes I think it should be doable.

Sounds good? (thanks for the sample code)

omry: > > Can you remind me why we can't represent `QuotedString(text=r"foo \"bar", quote=Quote.double),` with `r"'foo \\'bar'"`?
> 
> That's because the lexer will see a QUOTED_VALUE matching `'foo \\'` (meant to be resolved to `foo \`) and then an exception will be raised because it won't know what to do with the remaining characters `bar'` (same reason why `x = 'foo \\'bar'` in Python raises a SyntaxError).

So this is not because we are unescaping \\ on the resulting string (after the lexer job is done)?

> > Unescaping double backslashes everywhere in a quoted string seems like a pretty big hammer.
> > Can we limit the effect somehow to when we really need it?
> 
> Yes, we can limit ourselves to the `\\...\'` case (i.e., an arbitrary number of backslashes followed by a quote). This is similar to what I had in [omry/omegaconf#621](https://github.com/omry/omegaconf/pull/621) (see examples in this PR's description) -- I think if we do it in Hydra it would also make sense to do the same in OmegaConf for consistency.

This is not related to us using lexer modes there, right? it will just eliminate the need to use double backslash to get a backslash in quotes strings? Put up a PR and we'll see.

> > What if we had special handling for the case where a string ends with a double backslash and only unescape it in that case?
> 
> It would need to be generalized to the more general case I just mentioned but yes I think it should be doable.
> 
> Sounds good? (thanks for the sample code)

Why do you care about arbitrary number of trailing backslashes? the last one can be special (ensure we do not confuse with an escaped trailing quote), the rest can be regular and not escaped.

odelalleau: > > > Can you remind me why we can't represent `QuotedString(text=r"foo \"bar", quote=Quote.double),` with `r"'foo \\'bar'"`?
> > 
> > 
> > That's because the lexer will see a QUOTED_VALUE matching `'foo \\'` (meant to be resolved to `foo \`) and then an exception will be raised because it won't know what to do with the remaining characters `bar'` (same reason why `x = 'foo \\'bar'` in Python raises a SyntaxError).
> 
> So this is not because we are unescaping \ on the resulting string (after the lexer job is done)?

No, un-escaping `foo \\'bar` would actually yield the desired `foo \'bar`.

> This is not related to us using lexer modes there, right? it will just eliminate the need to use double backslash to get a backslash in quotes strings? Put up a PR and we'll see.

Correct. I’ll give it a shot.

Just throwing another idea out there: use duplicated quotes `''` or `""` instead of escaping. This would avoid making ˋ\ˋ special.

> > It would need to be generalized to the more general case I just mentioned but yes I think it should be doable.
> > Sounds good? (thanks for the sample code)
> 
> Why do you care about arbitrary number of trailing backslashes? the last one can be special (ensure we do not confuse with an escaped trailing quote), the rest can be regular and not escaped.

This is related to the first point above: it is for proper identification of the end of a quoted string with a regex. If `'abc\\\'`resolves to `abc\\` (I think this is what you are suggesting) and `'abc\\'` resolves to `abc\'` then how do you get a quoted string with one backslash followed by a quote in the middle of the string? (desired output: `abc\'def`)

odelalleau: > > This is not related to us using lexer modes there, right? it will just eliminate the need to use double backslash to get a backslash in quotes strings? Put up a PR and we'll see.
> 
> Correct. I’ll give it a shot.

Done in 2369a0638

It does make the un-escaping code (and its `with_quotes()` counterpart) more complex, but personally I think the benefits in terms of user-friendliness are worth it.

odelalleau: I also added a few more tests in 3bf6156b6 to better test what happens with more complex interpolations inside quoted strings

odelalleau: Aaaand a doc improvement to better explain how to pass interpolations through the command line (in case someone would actually want to do it): 00cd37f9a

omry: > This is related to the first point above: it is for proper identification of the end of a quoted string with a regex. If `'abc\\\'`resolves to `abc\\` (I think this is what you are suggesting) and `'abc\\'` resolves to `abc\'` then how do you get a quoted string with one backslash followed by a quote in the middle of the string? (desired output: `abc\'def`)

Regex is an implementation detail here.
With my implementation, this passes:
```
(r"abc\\'def", r"abc\'def"),
```
I will hold reviewing the changes until we clear this point up.

odelalleau: > > This is related to the first point above: it is for proper identification of the end of a quoted string with a regex. If `'abc\\\'`resolves to `abc\\` (I think this is what you are suggesting) and `'abc\\'` resolves to `abc\'` then how do you get a quoted string with one backslash followed by a quote in the middle of the string? (desired output: `abc\'def`)
> 
> Regex is an implementation detail here.
> With my implementation, this passes:
> 
> ```
> (r"abc\\'def", r"abc\'def"),
> ```

Your implementation is operating at a lower level, *after* the content of the string has been extracted (in your example, the characters `abc\\'def`, before un-escaping).

Stepping back to the higher-level grammar parsing, this content must from a quoted string, that one might want to look like this (again, these are the characters inside the string):
```
'abc\\'def'
```
I consider that such a quoted string would actually be invalid, because `\\'` should mark the end of the string, since we also want the following quoted string to be valid:
```
'abc\\'  <== should resolve to: abc\
```

It *might* be possible to somehow detect the end of the quoted string so that both examples above work, but this seems tricky (maybe I'm not seeing it?). It kinda worked in the previous implementation but at the expense of bugs like #1600.

omry: > 
```
'abc\\'def'
```
> I consider that such a quoted string would actually be invalid, because \\' should mark the end of the string, since we also want the following quoted string to be valid:

If we treat `\'` as an escaped quote that cannot close a string,  and the preceding `\` as a regular backslash, this string is not invalid.

If the low level lexer can correctly tokenize all the inputs we are talking about, I am not sure what is the problem.
The logic you are applying is operating on the characters in the quoted string token, right?
Here is an example that is closer to that for single quoted string, including `'abc\\'  <== should resolve to: abc\` case

The only real change is that now forward is also stripping the outer quotes.
I renamed forward to unquote and backward to quote.
This logic does not handle negative cases, it assumes the lexer would fail to produce a token if the string is illegal.
Example of an illegal string I am expecting the lexer to fail on: `'a\'` (It's illegal because it's not closed).

```python
from pytest import mark


def unquote(s: str) -> str:
    assert s[0] == s[-1] == "'"
    s = s[1:-1]
    if s.endswith("\\\\"):
        s1 = s[0:-2].replace(r"\'", "'") + "\\"
    else:
        s1 = s.replace(r"\'", "'")
    return s1


def quote(s):
    if s.endswith("\\"):
        s1 = s.replace("'", r"\'") + "\\"
    else:
        s1 = s.replace("'", r"\'")
    return f"'{s1}'"


@mark.parametrize(
    ("input_", "expected"),
    [
        (r"'ab'", r"ab"),
        (r"'a\'b'", r"a'b"),
        (r"'ab\''", "ab'"),
        (r"'\ab'", r"\ab"),
        (r"'ab\\'", r"ab\ ".strip()),
        (r"'ab\\''", r"ab\'"),
        (r"'abc\\'def'", r"abc\'def"),
        (r"'abc\\'", r"abc\ ".strip()),
    ],
)
class TestEscape:
    def test_unquote(self, input_, expected):
        res = unquote(input_)
        assert res == expected

    def test_round_trip(self, input_, expected):
        res = unquote(input_)
        assert res == expected
        back = quote(res)
        assert back == input_
```

odelalleau: > > 
> 
> ```
> 'abc\\'def'
> ```
> 
> > I consider that such a quoted string would actually be invalid, because \' should mark the end of the string, since we also want the following quoted string to be valid:
> 
> If we treat `\'` as an escaped quote that cannot close a string, and the preceding `\` as a regular backslash, this string is not invalid.
> 
> If the low level lexer can correctly tokenize all the inputs we are talking about, I am not sure what is the problem.

The problem *is* to get the lexer to correctly tokenize the inputs. What if you have the following override string (these are the characters in the string):
```
["abc\\", "def"]
```
=> how does it know that `"abc\\"` should be a single quoted value if we also allow `"abc\\", "` to be valid? I'm not saying it's impossible to achieve but I feel like it would make the lexer and/or parser significantly more complex, and may be heavier to maintain over time. I'd rather stay away from this unless you can suggest a satisfactory solution to this problem.

odelalleau: I rebased on top of master and added 499802019 which contains the comments explaining the lexer regex for QUOTED_VALUE

odelalleau:
I just pushed b5f505e95 which doesn't change the behavior but makes a few minor improvements (mostly comments / doc)

hydra/core/override_parser/overrides_visitor.py
```diff
@@ -309,6 +308,55 @@ def _createPrimitive(
                 return node.getText()  # type: ignore
         return ret
 
+    def _unescape_quoted_string(self, text: str) -> str:
```
omry, line 311: Please add roundtrip testing for _unescape_quoted_string and with_quotes similar to what is in my code snippet.

website/docs/advanced/override_grammar/basic.md
```diff
@@ -178,12 +182,19 @@ $ python my_app.py 'dir=job\{a\=1\,b\=2\,c\=3\}'
...
 - `bool`: true, false
 - `dot_path`: foo.bar
-- `interpolation`: ${foo.bar}, ${co.env:USER,me}
+- `interpolation`: ${foo.bar}, ${oc.env:USER,me}
 
 Constants (null, true, false, inf, nan) are case-insensitive.
 
 :::important
-Always single-quote interpolations in the shell.
+Always single-quote interpolations in the shell, to prevent the shell from replacing them with environment variables:
```
omry, line 190:
```suggestion
Always single-quote interpolations in the shell, to prevent the shell from replacing them with shell variables:
```

website/docs/advanced/override_grammar/basic.md
```diff
@@ -178,12 +182,19 @@ $ python my_app.py 'dir=job\{a\=1\,b\=2\,c\=3\}'
...
+```shell
+$ python my_app.py 'dir=/root/${name}'
+```
+In addition, more complex interpolations containing special characters may require being passed within a quoted value:
+```shell
+$ python my_app.py 'dir="${get_dir: {root: /root, name: ${name}}}"'
+```
```
omry, line 197: This does not seem to be adding much to "Always single-quote interpolations in the shell".

hydra/core/override_parser/overrides_visitor.py
odelalleau: As just discussed, this is done in the current test => resolving

website/docs/advanced/override_grammar/basic.md
odelalleau: That was a lot of repetitions of "shell" so I made it a bit shorter in 26603c5bb

website/docs/advanced/override_grammar/basic.md
odelalleau: The important part here is that double quotes `"` around the interpolation => I emphasized more in bc328dc69

omry (approved):
Looks good.

@jieru-hu, `AssertionError: ray version mismatch` in https://app.circleci.com/pipelines/github/facebookresearch/hydra/9707/workflows/bb075f1d-7a96-420c-8c6b-c59c4789d29a/jobs/86093. Can you take a look?

Merge once CI is green.

PR #1605, opened 2021-05-04
Override grammar: allow the dollar sign in key names

### LINKED ISSUE #1437: [Feature Request] Allow keys with `$` in them
# 🚀 Feature Request

Following the discussion at https://hydra-framework.zulipchat.com/#narrow/stream/213879-Hydra/topic/Override.20keys.20with.20'.24'.20in.20them , it would be great if the grammar could be tweaked so that `$` is an allowed character in YAML keys when it's not part of an interpolation like `${foo}`, since `$` is a valid non-special character in YAML.

## Motivation

Our application uses special keys like `$args` and `$class` with a prefix to distinguish them from the other data in the configuration.  However, since these aren't allowed as keys in some contexts by Hydra, attempting to override them (e.g. at the command line) produces an error:

```
% python app.py 'agents.$args.c=11'
LexerNoViableAltException: agents.\$args.c=11
                                   ^
```

## Pitch

In general, it seems like any valid YAML should be accepted by Hydra in its config files and override mechanisms, unless it's specifically in conflict with Hydra additions.

We could use a different prefix like `_` or `APP_` or the like, but we'd love to avoid that, since we already have substantial infrastructure in place with this syntax and it seems at first look to not actually conflict with the Hydra `${foo}` syntax.

Another alternative would be to use Hydra's `_target_` mechanism for instantiation; so far it looks like we couldn't do this because our in-house mechanism is a bit more full-featured for our specific use cases.

I probably don't have enough expertise in the existing codebase to do so.

odelalleau:
Closes #1437

Note: I am not entirely sure we want to support `group@$pkg` but forbidding it would make the grammar more complex.

### DIFF
--- a/hydra/grammar/OverrideLexer.g4
+++ b/hydra/grammar/OverrideLexer.g4
@@ -25,7 +25,8 @@ COLON: ':';
 SLASH: '/';
 
 KEY_ID: ID -> type(ID);
-DOT_PATH: (ID | INT_UNSIGNED) ('.' (ID | INT_UNSIGNED))+;
+KEY_SPECIAL: (CHAR|'_'|'$') (CHAR|DIGIT|'_'|'$')*;  // same as ID but allowing $
+DOT_PATH: (KEY_SPECIAL | INT_UNSIGNED) ('.' (KEY_SPECIAL | INT_UNSIGNED))+;
 
 ////////////////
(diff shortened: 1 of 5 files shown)


omry (approved):
Overall looking good.

PR #1573, opened 2021-04-23
[WIP] Updates ray launcher with autoscaler sdk API

syed-ahmed:
Signed-off-by: Syed Tousif Ahmed <stahmed@seas.upenn.edu>

## Motivation

This PR updates the Ray Launcher with the autoscaler sdk API. The existing implementation uses python subprocesses to launch ray cluster using the ray CLI interface. This approach loses the live outputs emitted by ray, which are useful in seeing if a cluster was successfully launched or if we are stuck because of some permissions issue. The autoscaler sdk API emits the logs from ray and solves this issue.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebookresearch/hydra/blob/master/CONTRIBUTING.md)?

Yes

## Test Plan

Will possibly need a hydra project managed EC2 access for CI tests. Haven't developed the test plan yet.

## Related Issues and PRs

closes https://github.com/ray-project/ray/issues/8036

PR #1611, opened 2021-05-07
Updates ray launcher with autoscaler sdk API

syed-ahmed:
## Motivation

Follow up PR from https://github.com/facebookresearch/hydra/pull/1573.

This PR updates the Ray Launcher with the autoscaler sdk API. The existing implementation uses python subprocesses to launch ray cluster using the ray CLI interface. This approach loses the live outputs emitted by ray, which are useful in seeing if a cluster was successfully launched or if we are stuck because of some permissions issue. The autoscaler sdk API emits the logs from ray and solves this issue.


syed-ahmed:
@jieru-hu here's the first part with only sdk api changes.

jieru-hu:
thank you @syed-ahmed! we will take a look :)

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
```diff
@@ -14,18 +14,21 @@
...
-    ray_rsync_up,
     ray_tmp_dir,
-    ray_up,
-    rsync,
 )
 from hydra_plugins.hydra_ray_launcher.ray_aws_launcher import (  # type: ignore
     RayAWSLauncher,
 )
 
+# mypy complains about "unused type: ignore comment" on macos
```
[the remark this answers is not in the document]
syed-ahmed: When I just do a normal import with `from ray.autoscaler import sdk`, mypy says the following on linux:
```
error: Skipping analyzing 'ray.autoscaler': found module but no type hints or library stubs
```
On macos, the import doesn't complain, but it gives type errors on the sdk apis like:
```
error: Argument 1 to "create_or_update_cluster" has incompatible type "Union[Dict[Union[str, int, float, bool], Any], List[Any], None, str]"; expected "Union[Dict[Any, Any], str]"
```
So from linux, it looks like a mypy issue and on mac it looks like a ray issue?

plugins/hydra_ray_launcher/examples/upload_download/conf/hydra/launcher/custom_ray_aws.yaml
```diff
@@ -9,12 +9,11 @@ sync_up:
   # we leave target_dir to be null
   # as a result the files will be synced to a temp dir on remote cluster.
   # the temp dir will be cleaned up after the jobs are done.
-  # recommend to leave target_dir to be null if you are syncing code/artifacts to remote cluster so you don't need
+  # recommend to leave target_dir to be null if you are syncing code/artifacts
+  # to remote cluster so you don't need
   # configure $PYTHONPATH on remote cluster
-  include: ['model', '*.py']
```
jieru-hu, line 14: could you provide more info on why `include`  is removed?

plugins/hydra_ray_launcher/examples/upload_download/conf/hydra/launcher/custom_ray_aws.yaml
```diff
@@ -9,12 +9,11 @@ sync_up:
...
   # No need to sync up config files.
-  exclude: ['*']
+  exclude: ['conf', 'multirun', "**/__pycache__"]
```
jieru-hu, line 16: I'd prefer not to hardcode `multirun` here, since people can override output dir, and `multirun` won't necesserily be their output dir always.

plugins/hydra_ray_launcher/examples/upload_download/conf/hydra/launcher/custom_ray_aws.yaml
```diff
@@ -9,12 +9,11 @@ sync_up:
...
 sync_down:
-  include: ['*.pt', '*/']
   # No need to sync down config files.
-  exclude: ['*']
+  exclude: ['conf']
```
jieru-hu, line 19: same here, conf won't always be the name of the conf dir. 
I guess for the example purpose it's OK, but i still think it's better to specify what you want download instead.

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
```diff
@@ -71,7 +74,14 @@ def launch(
         launcher.ray_cfg.cluster.setup_commands = setup_commands
 
     configure_log(launcher.config.hydra.hydra_logging, launcher.config.hydra.verbose)
+    sdk.configure_logging(  # type: ignore[attr-defined]
+        log_style=launcher.log_style,
+        color_mode=launcher.color_mode,
+        verbosity=launcher.verbosity,
+    )
```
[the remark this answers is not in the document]
jieru-hu: +1

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
```diff
@@ -96,39 +106,41 @@ def launch(
...
 ) -> Sequence[JobReturn]:
-    ray_up(launcher.ray_yaml_path)
+    config = OmegaConf.to_container(
+        launcher.ray_cfg.cluster, resolve=True, enum_to_str=True
+    )
+    sdk.create_or_update_cluster(  # type: ignore[attr-defined]
+        config,
+        no_restart=launcher.update_cluster_no_restart,
+        restart_only=launcher.update_cluster_restart_only,
+        no_config_cache=launcher.update_cluster_no_config_cache,
```
jieru-hu, line 124: Is it possible to fold these 3 fields into the cluster config, just like how we are planning on grouping the logging config?

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
```diff
@@ -193,7 +206,11 @@ def launch_jobs(
                 log.info("NOT deleting the cluster (provider.cache_stopped_nodes=true)")
             else:
                 log.info("Deleted the cluster (provider.cache_stopped_nodes=false)")
-            ray_down(launcher.ray_yaml_path)
+            sdk.teardown_cluster(  # type: ignore[attr-defined]
+                config,
```
jieru-hu, line 210: these all look like cluster config, same as before, can we include workersonly and keep min workers in the config?

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_launcher_util.py
```diff
@@ -1,16 +1,23 @@
...
+from omegaconf import DictConfig
+
+# mypy complains about "unused type: ignore comment" on macos
+# workaround adapted from: https://github.com/twisted/twisted/pull/1416
+try:
+    import importlib
+
+    sdk = importlib.import_module("ray.autoscaler.sdk")
+except ModuleNotFoundError as e:
+    raise ImportError(e)
```
jieru-hu, line 20: I remember running into something similar ( mypy complains about different things on Mac vs Linux with ray ) 😢  - I can run this locally on my Mac to see if there's better solution after other comments are more or less addressed.

jieru-hu (changes requested):
Thanks for the PR!

Could you paste the output of running the example app with this new sdk api?

plugins/hydra_ray_launcher/examples/upload_download/conf/hydra/launcher/custom_ray_aws.yaml
syed-ahmed: It's because ray doesn't define it. They only have `rsync_exclude` and `rsync_filter` in the [schema](https://github.com/ray-project/ray/blob/35ec91c4e04c67adc7123aa8461cf50923a316b4/python/ray/autoscaler/ray-schema.json#L286-L293), which is a bummer :(. So I was thinking that either we keep the current rsync_up/down in `_launcher_util` and file a PR in ray for `--include` support or we only use `exclude` and assume that anything that isn't excluded or filtered using `rsync_filter` is going to be downloaded (which is what I'm assuming in the PR).

jieru-hu: I see! Thank you!
I have created an issue with ray here https://github.com/ray-project/ray/issues/15859, let's see if they will consider adding this option.
For now, I suggest we keep the rsync down/up configuration.

syed-ahmed:
@jieru-hu @omry I believe I have addressed the review. Please let me know if the new changes look ok.

Here are the outputs of the examples: [simple](https://gist.github.com/syed-ahmed/630e77a9c7a6808fe32cc17435cba357), [upload_download](https://gist.github.com/syed-ahmed/c3b4fda9054155092b922bf2e8d32d34)

Note that to run the examples, you have to override [custom_ray_aws.yaml](https://gist.github.com/syed-ahmed/1782967b22fad3fa940c829a16405e49) since this PR is not on master yet.

plugins/hydra_ray_launcher/tests/test_ray_aws_launcher.py
```diff
@@ -19,20 +19,22 @@
...
 from hydra_plugins.hydra_ray_launcher.ray_aws_launcher import (  # type: ignore
     RayAWSLauncher,
 )
 
+# mypy complains about "unused type: ignore comment" on macos
+# workaround adapted from: https://github.com/twisted/twisted/pull/1416
+try:
+    import importlib
+
+    sdk = importlib.import_module("ray.autoscaler.sdk")
```
omry, line 33: If you change this to:
```python
    sdk : Any = importlib.import_module("ray.autoscaler.sdk")
```

or:
```python
    sdk = importlib.import_module("ray.autoscaler.sdk") # type: ignore
```

Does it eliminate the need for all the specific mypy type comments all over?

syed-ahmed: `sdk : Any = importlib.import_module("ray.autoscaler.sdk")` works on linux and gets rid of the type comments, checking macos now with the CI run. The second one doesn't get rid of the type comments all over and in addition gives a `error: unused 'type: ignore' comment` on the import line.

syed-ahmed: Looks like `sdk : Any = importlib.import_module("ray.autoscaler.sdk")` gets rid of the type comments for both linux and macos.

omry: Awesome.

Side note:
I am really curious about the root cause of the difference between Mac and Linux here.

syed-ahmed: There might be some platform specific issues with mypy but may be mypy is not the issue here and the search results sound like what I want to hear. It seems like the main problem is with ray wheels: https://github.com/ray-project/ray/issues/14431. https://github.com/ray-project/ray/pull/14492 tried to fix it but I still could not find the `py.typed` file in the 1.3.0 release. 

The macos wheel does have the py.typed file, and so the normal import succeeds. The actual type errors on macos kinda make sense and is probably due to some type usage (also I'm a total beginner on the whole type system thing in python, so don't really have an in depth understanding unfortunately).

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_config.py
```diff
@@ -39,6 +39,86 @@ class RayAutoScalingMode(Enum):
...
+
+@dataclass
+class RayLoggingConf:
+    """
+    Ray sdk.configure_logging parameters: log_style, color_mode, verbosity
+    """
+
+    log_style: RayLoggingStyle = RayLoggingStyle.auto
+    color_mode: RayLoggingColorMode = RayLoggingColorMode.auto
+    verbosity: int = RayLoggingVerbosity.verbose.value
```
syed-ahmed, line 84: FYI, I originally wanted this to be `verbosity: RayLoggingVerbosity = RayLoggingVerbosity.verbose`. `OmegaConf.to_container` with `enum_to_str` gives `verbosity: "verbose"` instead of `verbosity: 1` and hence had to adopt the current solution. If you have a better solution, I'm happy to change this.

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
```diff
@@ -170,13 +179,13 @@ def launch_jobs(
                     sync_down_cfg.source_dir if sync_down_cfg.source_dir else sweep_dir
                 )
                 target_dir = (
-                    sync_down_cfg.source_dir if sync_down_cfg.source_dir else sweep_dir
+                    sync_down_cfg.target_dir if sync_down_cfg.target_dir else sweep_dir
```
jieru-hu, line 182: nice catch!

plugins/hydra_ray_launcher/setup.py
```diff
@@ -26,7 +26,7 @@
     install_requires=[
         "boto3==1.17.17",
         "hydra-core>=1.1.0.dev7",
-        "ray==1.2.0",
+        "ray==1.3.0",
```
jieru-hu, line 29: since we are bumping dependency, could you also bump the plugin version here
```
plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/__init__.py
```

jieru-hu (changes requested):
Looks great! Some comments inline.

Could you also update the plugin doc with the extra configs added, and also include a news segments


https://hydra.cc/docs/next/development/documentation

plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_config.py
syed-ahmed: It looks like there isn't much difference between [minimal](https://gist.github.com/syed-ahmed/27c64c0ff0fdc2eec6d4a45406a3119a) and [verbose](https://gist.github.com/syed-ahmed/c3b4fda9054155092b922bf2e8d32d34), and ray still outputs a lot at `minimal` logging. I additionally print out the cluster config. So may be I'll turn the cluster config logging to debug and the verbosity to minimal?

jieru-hu: Yea, i think we can skip printing out the cluster config (users can get the config from the app via Hydra cli `--cfg hydra --resolve`)
Let's keep the default logging minimal then :)

syed-ahmed:
@jieru-hu Addressed the latest comments. Please feel free to suggest any changes on the docs added.

website/docs/plugins/ray_launcher.md
```diff
@@ -185,31 +198,99 @@ Loaded cached provider configuration
...
-You can manage the Ray EC2 cluster lifecycle by configuring the two flags provided by the plugin:
+You can manage the Ray EC2 cluster lifecycle by configuring the flags provided by the plugin:
 
 - Default setting (no need to specify on commandline): Delete cluster after job finishes remotely:
 ```commandline
 hydra.launcher.stop_cluster=true
-hydra.launcher.ray.cluster.provider.cache_stopped_nodes=False
+hydra.launcher.ray.cluster.provider.cache_stopped_nodes=false
+hydra.launcher.teardown_cluster.workers_only=false
+hydra.launcher.teardown_cluster.keep_min_workers=false
```
jieru-hu, line 218: This is confusing - it's not super clear to me what is the difference between
```
hydra.launcher.ray.cluster.provider.cache_stopped_nodes
```
and 
```
hydra.launcher.teardown_cluster.workers_only
```

Could you add more details here?

website/docs/plugins/ray_launcher.md
```diff
@@ -185,31 +198,99 @@ Loaded cached provider configuration
...
+- Fully resolve all environment settings from the cloud provider again
+```commandline
+hydra.launcher.create_update_cluster.no_config_cache=true
+```
+
+
+##### Configure Ray Logging
+You can manage Ray specific logging by configuring the flags provided by the plugin:
+
+- Default setting (no need to specify on commandline): Use minimal verbosity and automatically
```
jieru-hu, line 272:
```suggestion
- Default config: Use minimal verbosity and automatically
```

website/docs/plugins/ray_launcher.md
```diff
@@ -185,31 +198,99 @@ Loaded cached provider configuration
...
+
+- Keep `hydra.launcher.ray.cluster.min_workers` worker nodes
+and delete the rest of the worker nodes
+```commandline
+hydra.launcher.teardown_cluster.keep_min_workers=true
+```
+
+Additionally, you can configure how to create or update the cluster:
+
+- Default setting (no need to specify on commandline): Run setup commands, restart Ray and use
```
jieru-hu, line 244:
```suggestion
- Default config: run setup commands, restart Ray and use
```

plugins/hydra_ray_launcher/news/1611.bugfix
```diff
@@ -0,0 +1 @@
+Use `target_dir` when using `hydra.launcher.sync_down`
```
jieru-hu, line 1: We can probably just skip this one

plugins/hydra_ray_launcher/news/1611.config
```diff
@@ -0,0 +1 @@
+Add Ray SDK API configs - logging, create_update_cluster, teardown_cluster
```
jieru-hu, line 1: pls add new line after text for each of the news segments

plugins/hydra_ray_launcher/news/1611.maintenance
```diff
@@ -0,0 +1 @@
+Remove code utilizing the Ray CLI
```
jieru-hu, line 1:
```suggestion
Use autoscaler sdk instead of the CLI
```

plugins/hydra_ray_launcher/news/1611.docs
```diff
@@ -0,0 +1 @@
+Add Ray SDK API config usage
```
jieru-hu, line 1: We can skip this :)

plugins/hydra_ray_launcher/news/1611.feature
```diff
@@ -0,0 +1 @@
+Add live Ray logging using the Ray Autoscaler SDK API
```
jieru-hu, line 1: I think this included in 1611.config?

jieru-hu (changes requested):
Thanks for the doc updates!

Could you also update this section as well (github won't let me comment on them directly)

image

in particular, `ray's cluster launcher cli`, `ray cluster launcher expects a yaml file ` part

website/docs/plugins/ray_launcher.md
syed-ahmed: I see, it is confusing indeed. Here's what I'm seeing with those options:

| cached_stopped_nodes | workers_only | behavior |
|---------------------------------|--------------------|-------------|
| false | false | All nodes are terminated |
| false | true | Keeps head node running and terminates only worker node |
| true | false | Keeps both head node and worker node and stops both of them |
| true | true | Keeps both head node and worker node and stops only worker node |

jieru-hu: Thanks for trying every combo out! Let's add this to the doc for future reference.

syed-ahmed:
@jieru-hu Made the doc changes :)

jieru-hu:
@syed-ahmed  awesome work!!!

PR #1614, opened 2021-05-10
removed HydraJobException

### LINKED ISSUE #1613: [Bug] pdb support is broken
# 🐛 Bug
## Description

In patch #1239, debugger can work with hydra by setting `HYDRA_FULL_ERROR=1` , but pr #1465 breaks it. In #1465, `HydraJobException` is always raised if the job raise an exception, such that debugger will catch `HydraJobException` instead of the original exceptions.

## Checklist

## To reproduce

```python
# mini.py
import hydra

@hydra.main()
def foo(cfg):
    raise Exception

foo()
```
then run `HYDRA_FULL_ERROR=1 python -m pdb -c continue mini.py` in shell.

pdb catch `HydraJobException` in `hydra/core/utils.py`

## Expected Behavior
pdb catch `Exception` in `foo`.

## System information
- **Hydra Version** : 1.1.0dev6
- **Python version** : 3.8.8
- **Operating system** : linux 5.9.10

## Additional context

I think add `if _is_env_set("HYDRA_FULL_ERROR") or is_under_debugger():` in `JobReturn.return_value` can temporarily fix this.
```python
@dataclass
class JobReturn:
    ....

    @property
    def return_value(self) -> Any:
        assert self.status != JobStatus.UNKNOWN, "return_value not yet available"
        if self.status == JobStatus.COMPLETED:
            return self._return_value
        else:
            if _is_env_set("HYDRA_FULL_ERROR") or is_under_debugger():  # add this
                raise self._return_value
            raise HydraJobException(
                f"Error executing job with overrides: {self.overrides}"
            ) from self._return_value
```

omry:
Closes #1613

### DIFF
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -17,7 +17,6 @@ from hydra.core.config_search_path import ConfigSearchPath, SearchPathQuery
 from hydra.core.utils import get_valid_filename, validate_config_path
 from hydra.errors import (
     CompactHydraException,
-    HydraException,
     InstantiationException,
     SearchPathException,
 )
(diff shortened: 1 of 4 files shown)


jieru-hu (approved):
lgtm

this was also added to support the HydraJobException
https://github.com/facebookresearch/hydra/blob/c1ffab5357aeedfca3ef486dfae9f7e054e69588/hydra/_internal/utils.py#L227-L231

but probably OK to leave it as it is.

omry:
> lgtm
> 
> this was also added to support the HydraJobException
> https://github.com/facebookresearch/hydra/blob/c1ffab5357aeedfca3ef486dfae9f7e054e69588/hydra/_internal/utils.py#L227-L231
> 
> but probably OK to leave it as it is.

Good call. Will check if this can be removed now.

Issue #1602, opened 2021-05-03
[Bug] sporadic transient errors from Ray plugin CI


omry:
The Ray plugin is sometimes failing CI with what appears to be [transient errors](https://app.circleci.com/pipelines/github/facebookresearch/hydra/9528/workflows/38b7f487-d8ef-4957-995f-24e0fd04a8ef/jobs/84116).
Retrying usually makes this go away.
I suspect it has something to do with multiple sessions testing agaisnt the same AWS instance, but I am not sure.

jieru-hu:
Thanks Omry!

this particular failed run
[test_plugin_macos-3.8-hydra_ray_launcher](https://app.circleci.com/pipelines/github/facebookresearch/hydra/9528/workflows/38b7f487-d8ef-4957-995f-24e0fd04a8ef/jobs/84116) - _ TestRayAWSLauncherIntegration.test_custom_task_name[different_filename_and_config_file_name_override_and_command_line_override-tmpdir0-task_launcher_cfg0-extra_flags0]

jieru-hu:
another failure:

test_plugin_macos-3.7-hydra_ray_launcher
https://app.circleci.com/pipelines/github/facebookresearch/hydra/9690/workflows/dc6b027a-6a02-43e1-ab48-409d0d582a32/jobs/85904

```
>           assert local_version in out, f"{lib} version mismatch"
E           AssertionError: ray version mismatch
E           assert '1.2.0' in '2021-05-05 19:53:09,880\tVWARN commands.py:257 -- Loaded cached provider configuration from /var/folders/6y/gy9gggt14...ommands.py:263 -- If you experience issues with the cloud provider, try re-running the command with --no-config-cache.'

/Users/distiller/project/plugins/hydra_ray_launcher/tests/test_ray_aws_launcher.py:182: AssertionError

```

jieru-hu:
also things to fix for the tests:
1.  this [logic](https://github.com/facebookresearch/hydra/blob/207f6797b7c28f3784b8386fe44bc46be37076f6/plugins/hydra_ray_launcher/tests/test_ray_aws_launcher.py#L116-L117) is vulnerble, for example if the command issues a warning
2. The logging are way too verbose to help debugging

jieru-hu:
another failure
https://app.circleci.com/pipelines/github/facebookresearch/hydra/9912/workflows/82daa71e-fd2a-4556-b626-c10968e24719/jobs/88351
the test that failed
[...]

PR #1631, opened 2021-05-20
Update Ray AMI

jieru-hu:
python 3.8.10 is out, need to update the test AMI.

### DIFF
--- a/plugins/hydra_ray_launcher/tests/test_ray_aws_launcher.py
+++ b/plugins/hydra_ray_launcher/tests/test_ray_aws_launcher.py
@@ -54,7 +54,7 @@ except (NoCredentialsError, NoRegionError):
     aws_not_configured = True
 
 
-ami = os.environ.get("AWS_RAY_AMI", "ami-0bb600864743d4be3")
+ami = os.environ.get("AWS_RAY_AMI", "ami-074e55e3f371ada90")
 security_group_id = os.environ.get("AWS_RAY_SECURITY_GROUP", "sg-0a12b09a5ff961aee")
 subnet_id = os.environ.get("AWS_RAY_SUBNET", "subnet-acd2cfe7")
 instance_role = os.environ.get(


omry:
Thanks.
Is the process for this documented?
1. How do you know you need to update the AMI?
2. How do you update the AMI?

jieru-hu:
> Thanks.
> Is the process for this documented?
> 

> 1. How do you know you need to update the AMI?

The Exception complains that python version does not match (3.8.10 != 3.6.6), which means we do not have conda env available with this version, as a result, the test fall back on the base conda env.  (I will update the message to make it more clear. )

> 2. How do you update the AMI?

I used the script here  to build the AMI https://github.com/facebookresearch/hydra/tree/master/plugins/hydra_ray_launcher/integration_test_tools

I'm going to add a readme file in this folder (for the process.)


I've tracked these in #1602

omry:
What changed and caused the old AMI to break?

jieru-hu:
> What changed and caused the old AMI to break?

a new python version :(  (3.8.10) - we pre-installed all python version's conda env on the AMI - since 3.8.10 is recently released, it was not availabe on the AMI.

omry:
> and we need the two python versions to be the same for cloudpickle.

Can you provide more context on this? Is this restriction documented anywhere in cloudpickle?
I would expect cloudpickle to not be sensitive to minor Python version differences.

jieru-hu:
> > and we need the two python versions to be the same for cloudpickle.
> 
> Can you provide more context on this? Is this restriction documented anywhere in cloudpickle?
> I would expect cloudpickle to not be sensitive to minor Python version differences.

on their github they only say `Cloudpickle can only be used to send objects between the *exact same version of Python.*`. 
I think I ran into some issue with running on differnt python micro versions when I was working on the tests but I'm not 100%. I will look into  relaxing the version requirements to minor version only.

omry:
So yeah, I think it will help to pin our circleci images to something very specific as long as we are willing to update it from time to time.

PR #1632, opened 2021-05-20
Add missing NEWs fragments

jieru-hu:
Add news fragments for some of the closed issues.


jieru-hu:
two files are empty, will fix them.

news/1449.docs
```diff
@@ -0,0 +1 @@
+Documentation for `hydra.searchpath`
```
omry, line 1: Duplicate.

news/1377.api_change
```diff
@@ -0,0 +1 @@
+Support partial failures in MULTIRUN mode
```
omry, line 1: Add a newline at the end of news fragments.

news/1461.bugfix
```diff
@@ -0,0 +1 @@
+`--cfg=hydra` now works for top-level read-only config
```
omry, line 1: The key point is read-only.

```
--cfg=hydra now works when config is read-only
```

news/1529.feature
```diff
@@ -0,0 +1 @@
+Add `env` to Hydra's config group
```
omry, line 1: Let's keep this one internal. Without documentation this means absolutely nothing to users and I don't think we should document it right now.

omry (commented):
In general I have been pretty sparse with news fragments about documentation changes. I try to keep them limited to large items only, not minor changes.

news/1582.docs
```diff
@@ -1 +1 @@
-Documentation for configuring plugin
\ No newline at end of file
+Documentation for configuring plugin
```
Jasha10, line 1:
```suggestion
Document how to configure Hydra plugins
```

Jasha10 (commented):
Just one small suggestion:

PR #1649, opened 2021-06-03
ensure local hyperlinks end with ".md"

### LINKED ISSUE #1640: website bug: relative links get broken on page refresh
# 🐛 Bug
## Description
A "Page Not Found" error is displayed when clicking relative links after performing a page refresh.

## To reproduce

Step 1:
- In a web browser, go to this url: https://hydra.cc/docs/next/plugins/colorlog
Near the bottom of the webpage there is a sentence that says "Check [this page] for more information."
The [this page] link works.

Step 2:
- Refresh the webpage.
The url is changed to https://hydra.cc/docs/next/plugins/colorlog/ (note the trailing slash!).

Step 3:
- Click on the [this page] link. It now gives a "Page Not Found" error.
- This is because, in the `hydra.git/website/docs/plugins/colorlog.md` file,
the [this page] link is defined with a relative path
`[this page](../patterns/configuring_plugins)`
and the trailing slash on the URL messes up the relative path resolution.

I tried this in several web browsers (Firefox, Chrome, Safari). All the
browsers added a trailing slash to the URL when the page was refreshed.

The bug only gets triggered with some of the relative links in the docs.

I'm not sure why the trailing slashes are getting added to the URLs, but I think that [this](https://github.com/facebook/docusaurus/issues/3372) docusaurus issue is related...

## System information
- **Hydra Version** :  master branch
- **Virtual environment type and version** : tested with live webseite and with local development server
- **Operating system** : tested with Ubuntu and MacOS

Jasha10:
Local website hyperlinks now end with ".md", which closes #1640.

This bugfix is motivated by a comment [here](https://github.com/facebook/docusaurus/issues/3372#issuecomment-826759426) in a related Docusaurus issue.

### DIFF
--- a/website/docs/advanced/override_grammar/basic.md
+++ b/website/docs/advanced/override_grammar/basic.md
@@ -251,7 +251,7 @@ key="a,b","c,d"               # Elements can be quoted strings, ChoiceSweep("a,b
 key=[a,b],[c,d]               # Elements can be real lists, ChoiceSweep([a,b], [c,d])
 key={a:10, b:20},{c:30,d:40}  # And dictionaries: ChoiceSweep({a:10, b:20}, {c:30,d:40})
 ```
-More sweeping options are described in the [Extended Grammar page](extended).
+More sweeping options are described in the [Extended Grammar page](extended.md).
 
 :::important
 You may need to quote your choice sweep in the shell.
@@ -260,7 +260,7 @@ You may need to quote your choice sweep in the shell.
 
 ### Functions
 Hydra supports several functions in the command line.
-See the [Extended Grammar page](extended) for more information.
+See the [Extended Grammar page](extended.md) for more information.
 
 ## Working with your shell
 All shells interprets command line inputs and may change what is passed to the process.
--- a/website/docs/advanced/search_path.md
+++ b/website/docs/advanced/search_path.md
@@ -145,7 +145,7 @@ python my_app.py 'hydra.searchpath=[pkg://additonal_conf]'
 
 #### Overriding `--config-dir` from the command line
 This is a less flexible alternative to `hydra.searchpath`. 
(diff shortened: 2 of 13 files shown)


Jasha10:
I used the instructions from `hydra.git/website/README.md` to start a local development server.
The output from `yarn` alerted me if any change I made resulted in a broken link.
In a web browser pointing at `localhost:3000`, I tried all of the updated links. They are working even if the page that contains the link has a trailing slash in the URL.

omry (approved):
Looks good.

Side note, can you take a look at the Windows CI failure?
I think this is because of a change in a dependency.

Jasha10:
From the CI output:
It looks like the `jupyter_core` package is trying to `import win32api`, which is causing an import error:
```
>       import win32api

E       ImportError: DLL load failed while importing win32api: The specified procedure could not be found.
```
I noticed that the `hydra.git/.circleci/config.yml` file specifies a conda environment that explicitly installs pywin32:
```yaml
conda create -n hydra python=<< parameters.py_version >> pywin32 virtualenv==20.0.33 -qy
```
[Here](https://github.com/mhammond/pywin32/issues/1431) is a relevant issue in the `pywin32` github repo.
One suggested workaround in the comments is to pin `pywin32==225`.

omry:
Installing pywin32 was a solution to a related problem. The issue you are pointing to is old. Something changed in the past week to break this.

Jasha10:
It looks like the pywin32 package had a [release](https://github.com/mhammond/pywin32/releases/tag/b301) 5 days ago.

Jasha10:
And there is a [related](https://github.com/mhammond/pywin32/issues/1709) issue from 3 days ago.
In the comments, someone is suggesting to pin `pywin32==300`, which is the previous release.
I'll make a PR with the pin to see if the CI passes...

PR #1655, opened 2021-06-08
enable --resolve flag usage with --help

### LINKED ISSUE #1482: possible to print `$CONFIG` in hydra help with interpolation resolved?
Hi, thank you for the brilliant library once again!

I am using `--help` functionality following the [docs](https://hydra.cc/docs/configure_hydra/app_help), and wondering if the values printed with `$CONFIG` can be 'resolved.
For example, what I get right now is
```yaml
...
load_dataset:
  _target_: datasets.load_dataset
  path: ${hydra:runtime.cwd}/loading_script.py
...
```
Is there a way to get `hydra:runtime.cwd` resolved to an actual path?

Thanks!

p.s. Is there a functional equivalence of python argparse's variable-wise `help`?
```python
parser.add_argument('foo', help='can hydra have something like this to explain each variable?')
```

Jasha10:
This PR adds compatibility between the `--help` flag and `--resolve` flag.
In particular, if the app help displayed by running `python my_app.py --help` contains the `$CONFIG` key, it is now possible to have interpolations in the displayed config resolved by including the `--resolve` flag (`python my_app.py --help --resolve`).

Closes #1482.

### DIFF
--- a/news/1482.feature
+++ b/news/1482.feature
@@ -1,1 +1,1 @@
-Add support resolving interpolations the config printed by `--help` via the `--resolve` flag.
+Add support for resolving interpolations in the config printed by `--help` via the `--resolve` flag.
--- a/news/1482.feature
+++ b/news/1482.feature
@@ -1,0 +1,1 @@
+Enable using the --resolve and --help flags together on the command line to resolve interpolations in app config before printing app help.
--- a/news/1482.feature
+++ b/news/1482.feature
@@ -1,1 +1,1 @@
-Enable using the --resolve and --help flags together on the command line to resolve interpolations in app config before printing app help.
+Add support resolving interpolations the config printed by `--help` via the `--resolve` flag.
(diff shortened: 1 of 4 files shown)


Jasha10:
I decided not to enable `--resolve` together with `--hydra-help` because, if `--hydra-help` contains the `$CONFIG` key, that will often have some interpolation-to-missing values (e.g. `hydra.sweep.subdir: ${hydra.job.num}`, which is an interpolation to a missing value by default), so using `--resolve` would crash with an `omegaconf.errors.InterpolationToMissingValueError`.

news/1482.feature
```diff
@@ -0,0 +1 @@
+Enable using the --resolve and --help flags together on the command line to resolve interpolations in app config before printing app help.
```
omry, line 1:
```suggestion
Add support resolving interpolations the config printed by `--help` via the `--resolve` flag.
```

omry (approved):
Looking good, one nit about the news fragment.

PR #1664, opened 2021-06-10
Allow range(stop) with default start for override grammar

victorjoos:
## Motivation

To make the range function more like the default python range(), and easier to type, I added a check if only start was added to set stop to start and start to zero.

Yes

## Test Plan

tested by adding a parametrize argument in `test_basic_sweeper.py`.

## Related Issues and PRs

N/A

### DIFF
--- a/hydra/_internal/grammar/grammar_functions.py
+++ b/hydra/_internal/grammar/grammar_functions.py
@@ -184,7 +184,9 @@ def choice(
 
 
 def range(
-    start: Union[int, float], stop: Union[int, float], step: Union[int, float] = 1
+    start: Union[int, float],
+    stop: Optional[Union[int, float]] = None,
+    step: Union[int, float] = 1,
 ) -> RangeSweep:
     """
     Range is defines a sweeep over a range of integer or floating-point values.
@@ -193,6 +195,9 @@ def range(
     For a negative step, the contents of the range are still determined by the formula
      r[i] = start + step*i, but the constraints are i >= 0 and r[i] > stop.
     """
+    if stop is None:
+        stop = start
+        start = 0
     return RangeSweep(start=start, stop=stop, step=step)
 
 
--- a/hydra/_internal/grammar/grammar_functions.py
+++ b/hydra/_internal/grammar/grammar_functions.py
@@ -189,7 +189,9 @@ def range(
     step: Union[int, float] = 1,
 ) -> RangeSweep:
     """
-    Range is defines a sweeep over a range of integer or floating-point values.
+    Range defines a sweep over a range of integer or floating-point values.
+    When only start is defined, it is set as the stop value, and start is set at
+    zero.
     For a positive step, the contents of a range r are determined by the formula
      r[i] = start + step*i where i >= 0 and r[i] < stop.
     For a negative step, the contents of the range are still determined by the formula
--- a/news/1664.feature
+++ b/news/1664.feature
@@ -1,0 +1,1 @@
+Allow range() in override grammar to have only one argument (the stop value), e.g. range(3)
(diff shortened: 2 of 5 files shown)


Jasha10:
Thanks for the PR!

It would be good to have a few more tests -- see the `tests/test_overrides_parser.py` file, where most of the `range` tests currently reside.
In particular, I think it would be good to pin down behavior for the following:
- behavior when some inputs are zero or negative, e.g. range(-10, step=1), range(10, step=-1), range(0), ...
  for some of these cases, it is not immediately obvious to me what the behavior should be.
- behavior when called on float inputs

victorjoos:
Aaah, Thanks for the tips !

I was a bit fast in implementing this, didn't notice I had to also modify the parser. Let me modify the right files.

Not sure about the float inputs, but for the other cases I would say the best is to compare to the default python range implementation: 

 - `range(-10)  == []`
 - `range(-2, step=-1) == [0, -1]`
 - `range(0) == []`
 - `range(10, step=-1) == []`

victorjoos:
Actually didn't need to change the parser. I added some tests and a new example in the docs.

news/1664.feature
```diff
@@ -0,0 +1 @@
+Allow range() in override grammar to have only one argument (the stop value).
```
jieru-hu, line 1: pls add a line after this.

jieru-hu (approved):
one nit, otherwise lgtm. Thanks!

jieru-hu:
@omry  could you take a look as well?

omry:
Of course. but as I am in the middle of releasing 1.1 this will wait.

tests/test_basic_sweeper.py
```diff
@@ -46,6 +46,7 @@
             id="batches_of_2",
         ),
         param(["a=range(0,3)"], None, [[["a=0"], ["a=1"], ["a=2"]]], id="range"),
+        param(["a=range(3)"], None, [[["a=0"], ["a=1"], ["a=2"]]], id="range_no_start"),
```
omry, line 49: This is testing the basic sweeper. remove this test, it's not needed here.

tests/test_overrides_parser.py
```diff
@@ -402,6 +402,19 @@ def test_simple_choice_sweep(value: str, expected: Any) -> None:
...
+            "range(-10, step=-1)",
+            RangeSweep(start=0, stop=-10, step=-1),
+            id="no_start_negative",
+        ),
+        param("range(5.5)", RangeSweep(start=0, stop=5.5, step=1), id="no_start_float"),
+        param(
+            "range(5.5, step=0.5)",
+            RangeSweep(start=0, stop=5.5, step=0.5),
+            id="no_start_step_float",
+        ),
```
omry, line 417: Add a few tests to test_eval, it also checks the resulting output and not just the parsed object.

hydra/_internal/grammar/grammar_functions.py
```diff
@@ -184,7 +184,9 @@ def choice(
 
 
 def range(
-    start: Union[int, float], stop: Union[int, float], step: Union[int, float] = 1
+    start: Union[int, float],
+    stop: Optional[Union[int, float]] = None,
+    step: Union[int, float] = 1,
```
omry, line 189: Add a sentence to the function doc describing the new functionality.

news/1664.feature
omry, line 1:
```suggestion
Allow range() in override grammar to have only one argument (the stop value), e.g. range(3)
```

omry (changes requested):
Thanks. Looking pretty good.
See comments.

victorjoos:
Thank you, sorry for the delay. I implemented your suggestions.

omry (approved):
Looking good, thanks.

omry:
We have some existing CI failures we are looking at. Will likely need you to rebase once those are fixed to get CI green again.

jieru-hu:
@victorjoos  could you rebase master and force push to this branch? Many thanks!

PR #1686, opened 2021-06-21
[configen] Add support for `*args` and `**kwargs`

tmke8:
## Motivation

It would be nice if configen could handle `__init__` functions with `*args` and `**kwargs`:

```python
class MyClass:
    def __init__(self, *args: int, foo: str = "bar", **kwargs: str):
        ...
```

`*args` can be translated to `_args_`, and for `**kwargs` there is really not anything that configen can do with it,
so we'll just skip it. So, the above will be turned into

```python
@dataclass
class MyClass Conf:
    _args_: List[int] = MISSING
    foo: str = "bar"
```

There is also the difficulty that hydra seems to only allow `*args` as the first argument. Thus, we'll skip that as well if it appears at any other position.

Yes

## Test Plan

I added tests to verify the new behavior.

## Related Issues and PRs

No related issues to my knowledge.

### DIFF
--- a/tools/configen/configen/configen.py
+++ b/tools/configen/configen/configen.py
@@ -169,8 +169,14 @@ def generate_module(cfg: ConfigenConf, module: ModuleConf) -> str:
         params: List[Parameter] = []
         params = params + default_flags
 
-        for name, p in sig.parameters.items():
-            type_ = original_type = resolved_hints.get(name, sig.empty)
+        for i, (name, p) in enumerate(sig.parameters.items()):
+            if p.kind == p.VAR_KEYWORD:  # `**kwargs` parameter
+                continue
+            elif p.kind == p.VAR_POSITIONAL and i > 0:  # `*args` parameter
+                # hydra supports positional arguments only as the first arg
+                continue
+
+            type_or_empty = original_type = resolved_hints.get(name, sig.empty)
             default_ = p.default
 
             missing_value = default_ == sig.empty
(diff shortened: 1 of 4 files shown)


omry:
Thanks.

> There is also the difficulty that hydra seems to only allow *args as the first argument.

What do you mean here?

omry:
> > Thanks.
> > > There is also the difficulty that hydra seems to only allow *args as the first argument.
> > 
> > 
> > What do you mean here?
> 
> I at first wanted to support this
> 
> ```python
> class MyClass:
>     def __init__(self, num: int, *args: str):
>         ...
> ```
> 
> which would produce this config:
> 
> ```python
> @dataclass
> class MyClassConf:
>     _target_: str = "MyClass"
>     num: int = MISSING
>     _args_: List[str] = MISSING
> ```
> 
> but `instantiate({"_target_": "MyClass", "num": 3, "_args_": ["a", "b", "c"]})` threw an error saying that `num` was passed in _twice_ to the `__init__`, which I assume is because `num` is passed in as a named argument: `MyClass("a", "b", "c", num=3)`? (Instead of `MyClass(3, "a", "b", "c")`.)

Everything defined by name in the instantiated config is passed as named argument.
It's hard to say without debugging instantiate if there is anything to do to make this particular case work. It's the first time someone complains about it.
In any case, I am not sure it has anything to do with `_args_` being first in the config or not.

thomkeh:
>Everything defined by name in the instantiated config is passed as named argument.

Yes, and so it's not possible to pass in an argument that goes _before_ `*args`:

```python
>>> def f(a, *b):
...     print("a:", a)
...     print("b:", b)
...
>>> f(3, 3, 3)
a: 3
b: (3, 3)
>>> f(3, 3, a=3)
  File "<stdin>", line 1, in <module>
TypeError: f() got multiple values for argument 'a'
>>> f(a=3)
a: 3
b: ()
```
In this example, if you want to pass in `a` as a named argument, there is just no way to pass in `b` as well. And so, this
```
@dataclass
class FConf:
    _target_: str = "f"
    a: Any = MISSING
    _args_: List[Any] = MISSING
```
will not work with `f(a, *b)`, because `a` is passed in as a named argument by hydra. (It will however, work fine with `g(*b, a)`.)

And I'm not complaining that hydra can't handle this! I was just explaining my design choices.

omry:
Thanks for the additional context.
I will take some time before I can properly review it because my focus is elsewhere right now.

omry:
Thanks for the contribution, and sorry this sat for so long.

I am going to close this as stale. If configen is still useful but lacking here, the best next step is to start with a new issue describing the concrete use case and the desired API/semantics before sending another PR.

PR #1692, opened 2021-06-22
Remove pin cma version in nevergrad plugin

jieru-hu:
facebookresearch/nevergrad#1160 

is fixed


jieru-hu:
nvm i forgot we need to wait for the new release 🤦🏻‍♀️ will create a task to track this

PR #1701, opened 2021-06-29
Add --upgrade_warnings_as_errors flag

### LINKED ISSUE #1689: Option to fail application if there is any upgrade warnings
from discussion with @omry offline - 
This could be helpful for users to upgrade from Hydra 1.0 to Hydra 1.1.

Jasha10:
This PR is to close #1689.

The main contributions are:
  - create a new subclass `HydraUpgradeWarning` of `UserWarning`
  - In deprecation warnings, use `category=HydraUpgradeWarning`
  - Add a command line flag `--upgrade_warnings_as_errors` that causes `warnings.warn("...", category=HydraUpgradeWarning)` to raise an error instead of a warning.

### DIFF
--- a/hydra/_internal/callbacks.py
+++ b/hydra/_internal/callbacks.py
@@ -24,7 +24,8 @@ class Callbacks:
                 getattr(c, function_name)(**kwargs)
             except Exception as e:
                 warnings.warn(
-                    f"Callback {type(c).__name__}.{function_name} raised {type(e).__name__}: {e}"
+                    f"Callback {type(c).__name__}.{function_name} raised {type(e).__name__}: {e}",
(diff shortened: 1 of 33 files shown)


omry:
I was thinking about something a bit different:
1. create a common function to call to issue a warning.
2. That function can check the presence of an environment variable to decide if to trigger an error as well (sys.exit, maybe just an exception).

This way, we will also be able to replace that function at runtime to add additional logic to the warning (such as custom logging in fbcode).

What do you think?

Jasha10:
> What do you think?

Sounds reasonable to me. I've opened #1703 with an initial implementation.

PR #1703, opened 2021-06-30
Support HYDRA_DEPRECATION_WARNINGS_AS_ERRORS

### LINKED ISSUE #1689: Option to fail application if there is any upgrade warnings

Jasha10:
This PR is to close #1689, replaces #1701.

The main contributions are:
  - create a `deprecation_warning` function for use in the Hydra codebase
  - Add support for a `HYDRA_DEPRECATION_WARNINGS_AS_ERRORS` environment variable, which controls whether `deprecation_warning` results in a warning or an error.


hydra/_internal/utils2.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
Jasha10, line 1: I created this new file because I was getting circular import issues when I put the `deprecation_warning` function in `hydra.utils` or `hydra.core.utils` or `hydra._internal.utils`.

Jasha10:
I did not replace all warnings with the new internal function -- only the deprecation warnings.
@omry, did you have in mind that all warnings should go through the new function call?
The other (non-deprecation) warnings I found are:
  - In `hydra.core.plugins.Plugins`, a warning is issued if a plugin `ImportError` occurs when loading a plugin,
  - In `hydra.override_parser.overrides_visitor.HydraErrorListener`, warnings are issued if there is a lexer/parser issue
  - In `hydra._internal.callbacks.Callbacks`, a warning is issued if a callback raises an `Exception`.
  - In `hydra._internal.core_plugins.structured_config_source.StructuredConfigSource`, a warning is issued if an `Exception` occurs when importing the module that containing the structured config.
  - In `hydra._internal.config_loader_impl.ConfigLoaderImpl`, a warning is issued if a config source is not available.

hydra/_internal/utils2.py
omry: That's odd. can you give more info about the circular import?
what is the cycle?

hydra/compose.py
```diff
@@ -45,13 +46,12 @@ def compose(
...
             dedent(
-                """\
-
-        The strict flag in the compose API is deprecated and will be removed in the next version of Hydra.
-        See https://hydra.cc/docs/upgrades/0.11_to_1.0/strict_mode_flag_deprecated for more info.
-        """
+                """
+                The strict flag in the compose API is deprecated and will be removed in the next version of Hydra.
+                See https://hydra.cc/docs/upgrades/0.11_to_1.0/strict_mode_flag_deprecated for more info.
+                """
```
omry, line 54: This is changing the printed string.
Can you trigger the deprecation and see what it actually looks like before and after?

tests/test_utils.py
```diff
@@ -60,3 +63,22 @@ def test_to_absolute_path_without_hydra(
...
+)
+def test_deprecation_warning(env: Dict[str, str], expected_error: bool) -> None:
+    msg = "Feature FooBar is deprecated"
+    with patch("os.environ", env):
+        if expected_error:
+            with raises(DeprecationWarning, match=re.escape(msg)):
+                deprecation_warning(msg)
+        else:
+            with warns(UserWarning, match=re.escape(msg)):
+                deprecation_warning(msg)
```
omry, line 84: 1. We use monkeypatch in other places. e.g. `monkeypatch.setenv("TEST_ENV", "test_env")`.
2. Can you add a test_app that calls `deprecation_warning()` and test that is fails with a proper error message?

omry (commented):
What I had in mind is to fail only on deprecation warnings.

hydra/_internal/utils2.py
Jasha10: Adding `def deprecation_warning(...)` to `hydra/_internal/utils.py` and putting `from hydra._internal.utils import deprecation_warning` in `hydra/core/utils.py`, we get this cycle:
```text
hydra._internal.utils
-> hydra.core.utils
-> hydra._internal.utils
```

Adding `def deprecation_warning(...)` to `hydra/core/utils.py` and putting `from hydra.core.utils import deprecation_warning` in `hydra/core/default_element.py`, we get this cycle:
```text
hydra.core.utils
-> hydra.core.hydra_config
-> hydra.conf
-> hydra.core.config_store
-> hydra.plugins.config_source
-> hydra.core.default_element
-> hydra.core.utils
```

Adding `def deprecation_warning(...)` to `hydra/utils.py` and putting `from hyd.... import deprecation_warning` in `hydra/core/default_element.py`, we get a similar cycle:
```text
hydra.utils
-> hydra._internal.instantiate._instantiate2
-> hydra._internal.utils
-> hydra.core.utils
-> hydra.core.hydra_config
-> hydra.conf
-> hydra.core.config_store
-> hydra.plugins.config_store
-> hydra.plugins.config_source
-> hydra.core.default_element
-> hydra.utils
```

tests/test_utils.py
Jasha10: 1. Got it. Updated to use `monkeypatch` in https://github.com/facebookresearch/hydra/pull/1703/commits/9931989b9725c8e40d3eb069e074b1a3efbbf3de.
2. In https://github.com/facebookresearch/hydra/pull/1703/commits/ea54062c149030e10321e7b94f2bcc6cc17b5ab9 I added a test app.

hydra/compose.py
Jasha10: I believe you are mistaken, the string is not changed :)
```python
>>> msg1 = dedent(
...                  """\
...
...         The strict flag in the compose API is deprecated and will be removed in the next version of Hydra.
...         See https://hydra.cc/docs/upgrades/0.11_to_1.0/strict_mode_flag_deprecated for more info.
...         """
... )
>>> msg1
'\nThe strict flag in the compose API is deprecated and will be removed in the next version of Hydra.\nSee https://hydra.cc/docs/upgrades/0.11_to_1.0/strict_mode_flag_deprecated for more info.\n'
>>> msg2 = dedent(
...                 """
...                 The strict flag in the compose API is deprecated and will be removed in the next version of Hydra.
...                 See https://hydra.cc/docs/upgrades/0.11_to_1.0/strict_mode_flag_deprecated for more info.
...                 """
... )
>>> msg2
'\nThe strict flag in the compose API is deprecated and will be removed in the next version of Hydra.\nSee https://hydra.cc/docs/upgrades/0.11_to_1.0/strict_mode_flag_deprecated for more info.\n'
>>> assert msg1 == msg2
```
I took away the backslash at the beginning of the string, so the extra newline is no longer necessary.

Jasha10: I triggered the warning (using the `compose` `strict` flag) just to be sure.

hydra/_internal/utils2.py
omry: Okay, put it in a module with a better name.

`hydra._internal.deprecation_warning` ?

hydra/compose.py
omry: Interesting, I tested myself and got a different string, but maybe me copying the input caused the string to change.

tests/test_hydra.py
```diff
@@ -1399,3 +1400,45 @@ def test_frozen_primary_config(
...
+) -> None:
+    cmd = [
+        "tests/test_apps/deprecation_warning/my_app.py",
+        f"hydra.run.dir={tmpdir}",
+    ]
+    if env_deprecation_err:
+        monkeypatch.setenv("HYDRA_DEPRECATION_WARNINGS_AS_ERRORS", "1")
+    _, err = run_python_script(
+        cmd, allow_warnings=True, print_error=False, raise_exception=False
+    )
```
omry, line 1443: There is a run_with_error function that returns stderr even if there is an exception.
Can you try to use it instead of changing run_python_script?

hydra/test_utils/test_utils.py
```diff
@@ -463,3 +464,21 @@ def assert_regex_match(
                 from_name=from_name,
                 to_name=to_name,
             )
+
+
+def assert_multiline_regex_search(
+    pattern: str, string: str, from_name: str = "Expected", to_name: str = "Actual"
+) -> None:
```
omry, line 471: can you give more details why this new function is needed and the existing one not work for it?

jieru-hu:
Thanks @Jasha10 

Could you paste what it looks like what it looks like for one of the deprecation warnings? It should be easy to trigger one:

```
$ git diff
diff --git a/examples/tutorials/basic/your_first_hydra_app/1_simple_cli/my_app.py b/examples/tutorials/basic/your_first_hydra_app/1_simple_cli/my_app.py
index 99a8af7fd..552a84f6a 100644
--- a/examples/tutorials/basic/your_first_hydra_app/1_simple_cli/my_app.py
+++ b/examples/tutorials/basic/your_first_hydra_app/1_simple_cli/my_app.py
@@ -4,7 +4,7 @@ from omegaconf import DictConfig, OmegaConf
 import hydra
 
 
-@hydra.main(config_path=None)
+@hydra.main()
     print(OmegaConf.to_yaml(cfg))

```
Then just run the app.

Jasha10:
Hi @jieru-hu,
Here is the output with the environment variable set:
```python
(hydra) ~/hydra.git $ HYDRA_DEPRECATION_WARNINGS_AS_ERRORS=1 python tmp.py
  File "/Users/jasha10/hydra.git/tmp.py", line 2, in <module>
    @hydra.main()
  File "/Users/jasha10/hydra.git/hydra/main.py", line 29, in main
    deprecation_warning(
  File "/Users/jasha10/hydra.git/hydra/_internal/utils2.py", line 9, in deprecation_warning
    raise DeprecationWarning(message)
DeprecationWarning:
config_path is not specified in @hydra.main().
See https://hydra.cc/docs/next/upgrades/1.0_to_1.1/changes_to_hydra_main_config_path for more information.
```

Here is the output without the environment variable:
```python
(hydra) ~/hydra.git $ python tmp.py
/Users/jasha10/hydra.git/tmp.py:2: UserWarning:
config_path is not specified in @hydra.main().
See https://hydra.cc/docs/next/upgrades/1.0_to_1.1/changes_to_hydra_main_config_path for more information.
  @hydra.main()
```
Do you think it's confusing to have a `DeprecationWarning` instance raised as an error (instead of some other subclass of `Exception`)?

hydra/_internal/utils2.py
Jasha10: Done in https://github.com/facebookresearch/hydra/pull/1703/commits/801693dff7350dad2883dfc0213ae3843e8c0cb8.

tests/test_hydra.py
Jasha10: The `run_with_error` function adds the `-W` flag to the command, which makes warnings get turned into errors. This makes it harder to tell whether `HYDRA_DEPRECATION_WARNINGS_AS_ERRORS` is set or not.

hydra/test_utils/test_utils.py
Jasha10: The existing function processes the inputs line-by-line, using `zip` to match each line of the regex pattern against the corresponding line in the input string.

This was a problem because different CI platforms produce warning messages with different numbers of lines. For example, the test_linux-3.9 CI job produces a two-line warning that looks like this:
```text
tests/test_apps/deprecation_warning/my_app.py:10: UserWarning: Feature FooBar is deprecated
  deprecation_warning("Feature FooBar is deprecated")
```
and the test_linux-3.8 CI job produces a one-line warning that looks like this:
```text
tests/test_apps/deprecation_warning/my_app.py:10: UserWarning: Feature FooBar is deprecated
```

This new `assert_multiline_regex_search` function gives more flexibility to match a single regex pattern against a multi-line block of text, while `assert_regex_match` fails if the regex pattern does not have the same number of lines as the input string.
In commit https://github.com/facebookresearch/hydra/pull/1703/commits/d2ea5528fb70b8ac96f02dce04d7b6e22d95327f I tried to use the existing function `assert_regex_match` instead of this new function `assert_multiline_regex_search`. For that commit, test_linux-3.9 CI job succeeded while many of the other CI jobs failed.

jieru-hu:
Thanks @Jasha10!  Yea, I'm not sure about raising the warnings as well. Maybe it's worth looking into setting [`warnings.filterwarnings("error")`](https://docs.python.org/3/library/warnings.html#warnings.filterwarnings),to fail on deprecation warning.   (I believe this is how pytest fails tests on warnings as well.)
What do you think @omry ?

Jasha10:
In my previous attempt (PR #1701) I tried the `filterwarnings` approach (using `warnings.simplefilter(action="error", category=HydraUpgradeWarning)`), though Omry nixed it :)
See his [comment](https://github.com/facebookresearch/hydra/pull/1701#issuecomment-871004724) on #1701.

Another possibility would be to change type of exception that gets raised. Currently in this PR, the `deprecation_warning` function looks like this:
```python
def deprecation_warning(message: str, stacklevel: int = 1) -> None:
    warnings_as_errors = os.environ.get("HYDRA_DEPRECATION_WARNINGS_AS_ERRORS", None)
    if warnings_as_errors:
        raise DeprecationWarning(message)
    else:
        warnings.warn(message, stacklevel=stacklevel + 1)
```
We could change `raise DeprecationWarning(message)` to something like `raise HydraDeprecationError(message)`. I picked `DeprecationWarning` initially just because it seemed closest-to-correct.

omry:
I think we can raise a new kind of Hydra exception for this, maybe:
```python
raise HydraDeprecationError(message)
```

Jasha10:
> I think we can raise a new kind of Hydra exception for this, maybe:
> 
> ```python
> raise HydraDeprecationError(message)
> ```

Done in https://github.com/facebookresearch/hydra/pull/1703/commits/95d4a3d5638097b65b04e25ee69fb5249fdf1a27.

hydra/errors.py
```diff
@@ -39,3 +39,7 @@ def __init__(
         super(MissingConfigException, self).__init__(message)
         self.missing_cfg_file = missing_cfg_file
         self.options = options
+
+
+class HydraDeprecationError(HydraException):
+    ...
```
Jasha10, line 45: I wasn't sure whether to inherit from `HydraException` or from `CompactHydraException`.

Inheriting from `HydraException`, we get this traceback:
```text
$ HYDRA_DEPRECATION_WARNINGS_AS_ERRORS=1 python tests/test_apps/deprecation_warning/my_app.py
Error executing job with overrides: []
  File "/home/jbss/hydra.git/tests/test_apps/deprecation_warning/my_app.py", line 10, in my_app
    deprecation_warning("Feature FooBar is deprecated")
  File "/home/jbss/hydra.git/hydra/_internal/deprecation_warning.py", line 11, in deprecation_warning
    raise HydraDeprecationError(message)
hydra.errors.HydraDeprecationError: Feature FooBar is deprecated

```
Inheriting from `CompactHydraException`, we get this output:
```text
$ HYDRA_DEPRECATION_WARNINGS_AS_ERRORS=1 python tests/test_apps/deprecation_warning/my_app.py
Error executing job with overrides: []
Feature FooBar is deprecated

```
@omry and @jieru-hu, which is preferable?

omry: Can you try to trigger a realistic deprecation warning a user would see? say if they forget to specify config_path.

Jasha10: Forgetting to specify config_path like this:
```python
import hydra
@hydra.main()
def my_app(cfg):
    pass
my_app()
```
Gives the same traceback for both `HydraException` and `CompactHydraException`:

```text
  File "/home/jbss/hydra.git/my_app.py", line 2, in <module>
    @hydra.main()
  File "/home/jbss/hydra.git/hydra/main.py", line 29, in main
    deprecation_warning(
  File "/home/jbss/hydra.git/hydra/_internal/deprecation_warning.py", line 11, in deprecation_warning
    raise HydraDeprecationError(message)
hydra.errors.HydraDeprecationError:
config_path is not specified in @hydra.main().
See https://hydra.cc/docs/next/upgrades/1.0_to_1.1/changes_to_hydra_main_config_path for more information.
```
I think we are getting the same traceback in this case because the `deprecation_warning` is triggered before `run_job` gets called; the exception-sanitizing logic in `run_and_report` looks for the word `"run_job"` in the traceback.

Triggering a deprecation warning using `TargetConf`:
```python
from hydra import main, types
@main(config_path=None)
def my_app(cfg):
    _ = types.TargetConf()
my_app()
```
Here is the output if we inherit from `HydraException`:
<details><summary>(long traceback)</summary>

```text
$ HYDRA_DEPRECATION_WARNINGS_AS_ERRORS=1 python my_app.py
Error executing job with overrides: []
  File "/home/jbss/hydra.git/hydra/_internal/utils.py", line 211, in run_and_report
    return func()
  File "/home/jbss/hydra.git/hydra/_internal/utils.py", line 368, in <lambda>
    lambda: hydra.run(
  File "/home/jbss/hydra.git/hydra/_internal/hydra.py", line 110, in run
    _ = ret.return_value
  File "/home/jbss/hydra.git/hydra/core/utils.py", line 233, in return_value
    raise self._return_value
  File "/home/jbss/hydra.git/hydra/core/utils.py", line 160, in run_job
  File "/home/jbss/hydra.git/my_app.py", line 4, in my_app
    _ = types.TargetConf()
  File "<string>", line 4, in __init__
  File "/home/jbss/hydra.git/hydra/types.py", line 38, in __post_init__
    deprecation_warning(message=msg)
  File "/home/jbss/hydra.git/hydra/_internal/deprecation_warning.py", line 11, in deprecation_warning
    raise HydraDeprecationError(message)
hydra.errors.HydraDeprecationError:
TargetConf is deprecated since Hydra 1.1 and will be removed in Hydra 1.2.

  File "/home/jbss/hydra.git/my_app.py", line 5, in <module>
    my_app()
  File "/home/jbss/hydra.git/hydra/main.py", line 48, in decorated_main
    _run_hydra(
  File "/home/jbss/hydra.git/hydra/_internal/utils.py", line 367, in _run_hydra
    run_and_report(
  File "/home/jbss/hydra.git/hydra/_internal/utils.py", line 251, in run_and_report
    assert mdl is not None
AssertionError
```
</details>
And here is the output if we inherit from `CompactHydraException`:

```text
$ HYDRA_DEPRECATION_WARNINGS_AS_ERRORS=1 python my_app.py
Error executing job with overrides: []

TargetConf is deprecated since Hydra 1.1 and will be removed in Hydra 1.2.

```

Jasha10: In my opinion the `CompactHydraException` is much cleaner, but there is a problem:
someone might be confused if the output just says "Foobar is deprecated", since normally deprecations do not cause a program to fail immediately.

What do you think about adding the word `"HydraDeprecationError"` to the output? In the `deprecation_warning` function, we could replace
```python
raise HydraDeprecationError(message)
```
with

```python
raise HydraDeprecationError("HydraDeprecationError: " + message)
```
to produce this output (inheriting from `CompactHydraException`):

```text
$ HYDRA_DEPRECATION_WARNINGS_AS_ERRORS=1 python my_app.py
Error executing job with overrides: []
HydraDeprecationError:
TargetConf is deprecated since Hydra 1.1 and will be removed in Hydra 1.2.

```

omry: Deprecation warnings typically tell you where the problem is. I don't we should keep the more verbose version here.
This is just a debugging tool and not something users will constantly use.

Jasha10: Ok.

Jasha10:
@omry Will this feature still be useful? Should be ready to merge if your comments above are resolved.

hydra/_internal/deprecation_warning.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+import os
+import warnings
+
+from hydra.errors import HydraDeprecationError
+
+
+def deprecation_warning(message: str, stacklevel: int = 1) -> None:
+    warnings_as_errors = os.environ.get("HYDRA_DEPRECATION_WARNINGS_AS_ERRORS", None)
```
omry, line 9: nit: dict.get() returns None by default for non-existing keys.

This should be equivalent:
```suggestion
    warnings_as_errors = os.environ.get("HYDRA_DEPRECATION_WARNINGS_AS_ERRORS")
```

hydra/test_utils/test_utils.py
omry: I see. How does this function handle the case where the number of lines does not match?
Can it be used to replace the other implementation completely?

tests/test_hydra.py
omry: run_python_script is also preexisting, right?

I believe you can pass an env dictionary to run_python_script instead of mocking the environment like this.

omry (commented):
Sorry for the response.
Added a few small comments.

This is not urgent because we did not yet start to migrate people to 1.1 internally (we are still working on getting 1.0 apps to run in the same on both versions).

hydra/_internal/deprecation_warning.py
Jasha10: Oh, nice! I think I was confusing `dict.get` with `dict.pop`, which would raise an error if no default is supplied.

tests/test_hydra.py
Jasha10: > run_python_script is also preexisting, right?

Yes.

> I believe you can pass an env dictionary to run_python_script instead of mocking the environment like this.

You're right. Made the update in https://github.com/facebookresearch/hydra/pull/1703/commits/2c4d9513018ec0e722cf21c626f0c2a5f65065d1.

Jasha10:
> This is not urgent because we did not yet start to migrate people to 1.1 internally (we are still working on getting 1.0 apps to run in the same on both versions).

Ok, got it.

hydra/test_utils/test_utils.py
Jasha10: > I see. How does this function handle the case where the number of lines does not match?

This function only makes one call to `re.search`. The pattern that gets matched can contain newline characters `'\n'`.
For example, I used a question-mark pattern `'\n?'` to match an "optional newline" in one of the new `test_hydra.py` test cases.

> Can it be used to replace the other implementation completely?

Yes, that should be possible.

omry (approved):
lgtm.

@jieru-hu, looks like the Ray tests are failing for an unrelated reason.

jieru-hu:
> lgtm.
> 
> @jieru-hu, looks like the Ray tests are failing for an unrelated reason.

Yes, will fix them.

PR #1735, opened 2021-07-24
Fixed +experinent=foo causing unexpected changes to defaults list

### LINKED ISSUE #1706: [Bug] adding ColorLog override will disable other overrides
Hi
here i follow the basic tutorial and experiment section :
main.py:
```
from omegaconf import DictConfig, OmegaConf
import hydra

@hydra.main(config_path="conf", config_name="config")
    print(OmegaConf.to_yaml(cfg))

if __name__ == "__main__":
    my_app()
```

config.yaml:

```
defaults:
  - segment : color
  #- override hydra/job_logging: colorlog
  #- override hydra/hydra_logging: colorlog
volume: 1
```

segment/color.yaml :
`color_num=1`
segment/color2.yaml :
`color_num=10`

experiment/test.yaml :
```
# @package _global_
defaults:
  - override /segment: color2
volume: 9999
```

running the
 `:\python main.py +experiment=test`
will return `volume:9999`
BUT while un-commenting colorlog override in config.yaml the volume returned will be `volume:1`
why adding colorlog impact the other config ?
btw the colorlog do his job and the colorizing the output

thanks

omry:
This is undoing a previous bug fix for a related problem that attempted to insert the external overrides in the right place.
The idea behind forcing the order is just to communicate to the user that overrides are special and applies to what comes before. In practice the implementation should work correctly even if they are not.

The solution introduced here is to exempt external overrides from the order check.

Closes #1706

### DIFF
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -98,6 +98,7 @@ class Overrides:
                         group=override.key_or_group,
                         package=override.package,
                         value=value,
+                        external_append=True,
                     )
                 )
             else:
@@ -364,7 +365,9 @@ def _update_overrides(
             assert d.group is not None
             legacy_hydra_override = not d.is_override() and d.group.startswith("hydra/")
 
-        if seen_override and not (d.is_override() or legacy_hydra_override):
+        if seen_override and not (
+            d.is_override() or d.is_external_append() or legacy_hydra_override
+        ):
             assert isinstance(last_override_seen, GroupDefault)
             pcp = parent.get_config_path()
             okey = last_override_seen.get_override_key()
@@ -467,14 +470,7 @@ def _create_defaults_tree_impl(
         self_added = _validate_self(containing_node=parent, defaults=defaults_list)
 
     if is_root_config:
-        # To ensure config overrides are last, insert the external overrides before the first config override.
-        insert_idx = len(defaults_list)
-        for idx, default in enumerate(defaults_list):
-            if default.is_override():
-                insert_idx = idx
-                break
-
-        defaults_list[insert_idx:insert_idx] = overrides.append_group_defaults
+        defaults_list.extend(overrides.append_group_defaults)
 
     _update_overrides(defaults_list, overrides, parent, interpolated_subtree)
 
--- a/hydra/core/default_element.py
+++ b/hydra/core/default_element.py
@@ -261,6 +261,9 @@ class InputDefault:
     def is_override(self) -> bool:
         raise NotImplementedError()
 
+    def is_external_append(self) -> bool:
+        raise NotImplementedError()
+
 
 @dataclass
 class VirtualRoot(InputDefault):
@@ -304,6 +307,9 @@ class VirtualRoot(InputDefault):
         raise NotImplementedError()
 
     def is_override(self) -> bool:
+        return False
+
+    def is_external_append(self) -> bool:
         return False
 
 
@@ -420,6 +426,9 @@ class ConfigDefault(InputDefault):
     def is_override(self) -> bool:
         return False
 
+    def is_external_append(self) -> bool:
+        return False
+
 
 _legacy_interpolation_pattern: Pattern[str] = re.compile(r"\${defaults\.\d\.")
 
@@ -436,6 +445,8 @@ class GroupDefault(InputDefault):
     deleted: Optional[bool] = None
 
     config_name_overridden: bool = field(default=False, compare=False, repr=False)
+    # True if this item was added using +foo=bar from the external overrides
+    external_append: bool = field(default=False, compare=False, repr=False)
 
     def __post_init__(self) -> None:
         assert self.group is not None and self.group != ""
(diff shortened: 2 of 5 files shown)


Jasha10:
Here is the approach I took:
```git
$ git diff master
diff --git a/hydra/_internal/defaults_list.py b/hydra/_internal/defaults_list.py
index 0888952f7..487fca9f3 100644
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -352,8 +352,7 @@ def _update_overrides(
     parent: InputDefault,
     interpolated_subtree: bool,
 ) -> None:
-    seen_override = False
-    last_override_seen = None
+    seen_overrides: Dict[str, GroupDefault] = {}
     for d in defaults_list:
         if d.is_self():
             continue
@@ -364,16 +363,18 @@ def _update_overrides(
             assert d.group is not None
             legacy_hydra_override = not d.is_override() and d.group.startswith("hydra/")

-        if seen_override and not (d.is_override() or legacy_hydra_override):
-            assert isinstance(last_override_seen, GroupDefault)
+        if (
+            not (d.is_override() or legacy_hydra_override)
+            and d.get_override_key() in seen_overrides
+        ):
             pcp = parent.get_config_path()
-            okey = last_override_seen.get_override_key()
-            oval = last_override_seen.get_name()
+            okey = d.get_override_key()
+            oval = seen_overrides[okey].get_name()
             raise ConfigCompositionException(
                 dedent(
                     f"""\
-                    In {pcp}: Override '{okey} : {oval}' is defined before '{d.get_override_key()}: {d.get_name()}'.
-                    Overrides must be at the end of the defaults list"""
+                    In {pcp}: Override '{okey} : {oval}' is defined before '{okey} : {d.get_name()}'.
+                    Each override must come after the default it is overriding"""
                 )
             )

@@ -392,8 +393,7 @@ def _update_overrides(
                 warnings.warn(msg, UserWarning)

             if d.override:
-                seen_override = True
-                last_override_seen = d
+                seen_overrides[d.get_override_key()] = d
                 if interpolated_subtree:
                     # Since interpolations are deferred for until all the config groups are already set,
                     # Their subtree may not contain config group overrides
@@ -467,14 +467,7 @@ def _create_defaults_tree_impl(
         self_added = _validate_self(containing_node=parent, defaults=defaults_list)

     if is_root_config:
-        # To ensure config overrides are last, insert the external overrides before the first config override.
-        insert_idx = len(defaults_list)
-        for idx, default in enumerate(defaults_list):
-            if default.is_override():
-                insert_idx = idx
-                break
-
-        defaults_list[insert_idx:insert_idx] = overrides.append_group_defaults
+        defaults_list.extend(overrides.append_group_defaults)

     _update_overrides(defaults_list, overrides, parent, interpolated_subtree)

diff --git a/tests/defaults_list/test_defaults_tree.py b/tests/defaults_list/test_defaults_tree.py
index 3a0bae03f..2dcb977f7 100644
--- a/tests/defaults_list/test_defaults_tree.py
+++ b/tests/defaults_list/test_defaults_tree.py
@@ -621,13 +621,26 @@ def test_defaults_tree_with_package_overrides__group_override(
                 match=re.escape(
                     dedent(
                         """\
-                        In override_wrong_order: Override 'group1 : file2' is defined before 'group1: file1'.
-                        Overrides must be at the end of the defaults list"""
+                        In override_wrong_order: Override 'group1 : file2' is defined before 'group1 : file1'.
+                        Each override must come after the default it is overriding"""
                     )
                 ),
             ),
             id="test_override_wrong_order_in_defaults_list",
         ),
+        param(
+            "override_same_level",
+            ["+group2=file1"],
+            DefaultsTreeNode(
+                node=ConfigDefault(path="override_same_level"),
+                children=[
+                    GroupDefault(group="group1", value="file2"),
+                    ConfigDefault(path="_self_"),
+                    GroupDefault(group="group2", value="file1"),
+                ],
+            ),
+            id="override_same_level_with_append",
+        ),
     ],
 )
 def test_override_option_from_defaults_list(
```

omry:
> The approach I took lets overrides come before non-overrides as long as they have a different override key

Nice.

Given the underlying motivation for requiring that overrides are end the end is just for clarity and does not serve a functional purpose, I think it's better to just exempt externally appended config groups from the check as I did.
The code becomes simpler instead of more complicated:
as an example, I am not sure I fully understand why your approach is really solving the problem and other similar problems because it's harder to understand the solution.

Jasha10 (approved):
> Given the underlying motivation for requiring that overrides are end the end is just for clarity and does not serve a functional purpose, I think it's better to just exempt externally appended config groups from the check as I did.

Fair enough.
Also, I realized my diff would be a breaking change. The following test would pass on master or on this PR branch, but it fails on my diff with the error message `Override 'group1 : file1' is defined before 'group1 : file2'`:
```text
$ cat tests/defaults_list/data/override_only.yaml
defaults:
  - override group1: file1
```
```
$ cat tests/defaults_list/test_defaults_tree.py | sed '2042,2052!d'
        param(
            "override_only",
            ["+group1=file2"],
            DefaultsTreeNode(
                node=ConfigDefault(path="override_only"),
                children=[
                    GroupDefault(group="group1", value="file1"),
                ],
            ),
            id="append_to_override_only",
        ),
```

jieru-hu (approved):

PR #1741, opened 2021-07-30
handle frame in stack trace that doesn't have a module

### LINKED ISSUE #1739: [Bug] failure while sanitiainzg exception stack trace in some cases
In some scenarios, one of the insepected frames in the stack trace does not have a module, resulting in a failure to print the underlying error.

The following looks like it's resolving it:

```diff
++ hydra/_internal/utils.py
@@ -248,6 +248,8 @@
                 while end is not None:
                     frame = end.tb_frame
                     mdl = inspect.getmodule(frame)
+                    if mdl is None:
+                        break
-                     assert mdl is not None
```

Jasha10:
Closes #1739

Not sure how to test this.
Currently there are not any unit tests of the `run_and_report` function.

### DIFF
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -7,7 +7,7 @@ import sys
 from dataclasses import dataclass
 from os.path import dirname, join, normpath, realpath
 from traceback import print_exc, print_exception
-from types import FrameType
+from types import FrameType, TracebackType
 from typing import Any, Callable, List, Optional, Sequence, Tuple, Union
 
 from omegaconf.errors import OmegaConfBaseException
@@ -223,7 +223,7 @@ def run_and_report(func: Any) -> Any:
                 # It is possible to add additional libraries to sanitize from the bottom later,
                 # maybe even make it configurable.
 
-                tb: Any = ex.__traceback__
+                tb = ex.__traceback__
                 search_max = 10
                 # strip Hydra frames from start of stack
                 # will strip until it hits run_job()
@@ -243,7 +243,7 @@ def run_and_report(func: Any) -> Any:
                     sys.exit(1)
 
                 # strip OmegaConf frames from bottom of stack
-                end = tb
+                end: Optional[TracebackType] = tb
                 num_frames = 0
                 while end is not None:
                     frame = end.tb_frame
@@ -276,6 +276,7 @@ def run_and_report(func: Any) -> Any:
                     added = added + 1
                     cur.tb_next = FakeTracebackType()
                     cur = cur.tb_next
+                    assert iter_tb.tb_next is not None
                     iter_tb = iter_tb.tb_next
 
                 print_exception(etype=None, value=ex, tb=final_tb)  # type: ignore
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -284,8 +284,18 @@ def run_and_report(func: Any) -> Any:
                 sys.stderr.write(
                     "\nSet the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.\n"
                 )
-            except Exception:
-                print_exception(etype=None, value=ex, tb=ex.__traceback__)
+            except Exception as ex2:
+                sys.stderr.write(
+                    "An error occurred during Hydra's exception formatting:"
+                    + os.linesep
+                )
+                if ex2.__traceback__ is not None:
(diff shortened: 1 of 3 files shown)


omry:
Yeah, it's a hard one to test.

I am actually not sure that the best handling is to break.
Maybe we should revert to printing the full stacktrace in such a case?

omry:
you can test by triggering the code path to see what happens.

omry:
I would say that in any case there is any exception at all while trying print a nice exception in that branch we should revert to printing the full exception.

Jasha10:
I wrapped the nice-exception logic in a try/except block.

hydra/_internal/utils.py
```diff
@@ -213,75 +213,79 @@ def run_and_report(func: Any) -> Any:
...
+                        cur = cur.tb_next
+                        assert iter_tb.tb_next is not None
+                        iter_tb = iter_tb.tb_next
+
+                    print_exception(etype=None, value=ex, tb=final_tb)  # type: ignore
+                sys.stderr.write(
+                    "\nSet the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.\n"
+                )
+            except Exception:
+                print_exception(etype=None, value=ex, tb=ex.__traceback__)
```
omry, line 288: Probably better to raise the original exception instead after printing some kind of warning.
```suggestion
                # print something to stderr here.
                raise ex
```

omry:
Can you check the CI failures?

hydra/_internal/utils.py
Jasha10: Ok, as of commit https://github.com/facebookresearch/hydra/pull/1741/commits/7de5282b18c33f511d4c15f174120c0ea0ce6cb5 a warning is printed including the filename / line number where the second exception occurred.

Jasha10:
> Can you check the CI failures?

Ok, I've fixed most of the failures.
There are still a few ray_launcher failures, but I think they're unrelated.

hydra/_internal/utils.py
omry: The second exception is happening only in the processing logic.
It seems like an overkill.

You can simplify to something like:
```
                sys.stderr.write(
                    "An error occurred during Hydra's exception formatting:"
                    + os.linesep + repr(ex2)
                )
                raise ex
```

Jasha10: Done in https://github.com/facebookresearch/hydra/pull/1741/commits/cf7bbbfcfeab1d754fb9823f1292033c12004150.

tests/test_utils.py
```diff
@@ -60,3 +65,120 @@ def test_to_absolute_path_without_hydra(
     path = str(Path(path))
     expected = str(Path(expected).absolute())
     assert utils.to_absolute_path(path) == expected
+
+
+class TestRunAndReport:
+    class DemoFunctions:
+        """Demo inputs for `run_and_report`"""
```
omry, line 72: Can you explain the testing logic?
The test is surprisingly large.

Jasha10: > The test is surprisingly large.

Sorry about that.
This is not just testing the new functionality. It is also testing the pre-existing `run_and_report` behavior.

The `run_and_report` function has the following signature:
```
def run_and_report(func: Any) -> Any: ...
```

There are three test methods:
- `test_success` tests when `run_and_report` runs `func` successfully.
- `test_failure` tests when `func` raises an exception and `run_and_report` is able to print a nicely-formatted error message.
- `test_simplified_traceback_failure` tests when `func` raises an exception and `run_and_report` is *not* able to print a nicely-formatted error message -- this is the case where the original exception gets re-raised.

There are four `@staticmethod` definitions. These are passed in to `run_and_report` as the `func` argument. They are designed to trigger different logic code paths inside of `run_and_report`:
- `success_func` raises no exception
- `simple_error` raises an `AssertionError`
- `run_job_wrapper` triggers the special logic in `run_and_report` that looks for a function called `"run_job"` in the stack, and strips away the leading stack frames
- `omegaconf_job_wrapper` triggers the special logic in `run_and_report` that looks for the `omegaconf` module in the stack, and strips away the bottom stack frames

omry: Gotcha.
Thanks for adding this. Can you add a comment describing this in the test class?
(does not need to be as detailed).

omry (approved):
Overall looks good.
See comment.

omry:
Looking good, don't forget a bugfix news fragment.

PR #1744, opened 2021-08-01
Update dev.txt

tanujdhiman:
By specifying the version of Pytest it will download exactly the same otherwise it will download all versions and test one by one either it is compatible or not with other dependencies.

Also I found this `6.2.4` is compatible with others (updated too).

Here is some sample output when I was running without specifying the version ::

pytest

Thanks


omry:
Why does pip need to download all versions? I have never seen it happening.
Does it happen on a clean conda env where you install the dev dependencies?

tanujdhiman:
Yes @omry . I tried it in a clean env.

Thanks

omry:
Thanks. Can you do some more digging to see what is the root cause? It makes no sense to me that pytest users need to pin a specific version or face very long install times.

tanujdhiman:
Yeah sure. I'll do more research why this occur. Then get back to you!! 

Thanks

tanujdhiman:
Hi @omry 

> Yeah sure. I'll do more research why this occur. Then get back to you!!
> 
> Thanks

I found why this occurs. Sometimes when we do not specify the version of a library then pip has become confused which version is suitable so pip will start download all versions and trying to make one by one suitable with other dependencies. 

Hope you are understanding !!
Thanks

omry:
So this sounds like a pip issue.
I tested on a clean environment and was not able to reproduce the issue (see https://pastebin.ubuntu.com/p/CN3YR3kj6n/). What version of pip are you using? I got those results with pip 21.2.2.

If you get different results, please provide exact repro instructions (including how you create the conda environment).

I will not pin to a specific pytest version before I see the problem (pinning packages to specific versions should be done as a last resort).

tanujdhiman:
Thanks @omry 
Yes it is a Pip issue. I update this and found good results. But one thing I saw if anybody like me have not upgraded their pip version then he/she will again got this problem and then they make an issue in this repository. In my opinion to set the particular version for the particular library so that they will get good results not caring about their pip is upgraded or pip is not upgraded.

I've put all accurate version of all dependencies and find no issues now.

Hope you understand !!

omry:
I understand what you mean, but I disagree.
As I said, pinning to a specific version should be done as last resort. Not as a workaround for a long-fixed bug in pip.

Thanks for the PR, but I am not interested in merging it.

tanujdhiman:
Thanks @omry

PR #1726, opened 2021-07-20
Update optuna dependency and some cleanups

jieru-hu:
closes #1718

PR #1746, opened 2021-08-02
Follow-up 1726. Update optuna sweeper dependency and some cleanups.

toshihikoyanase:
## Motivation

This PR is a follow-up for #1726. Since #1726 has not merged yet, this PR includes the changes of the original PR.

*Changes*
- Set lower bound of Optuna version (>=2.5.0). Since the code in #1726 uses `study.ask` and `study.tell`, it does not work with `optuna<2.5.0`. See https://github.com/facebookresearch/hydra/pull/1726#discussion_r679915412.
- Update the scenario of `test_optuna_example` to save computation time. It also makes the test robust to Optuna's changes. See  "Test Plan" for more details.

Yes

## Test Plan

This PR changes the test scenario of `test_optuna_example`. The current test function tries to find the global optima by the random sampler, but it requires many trials (about 1000 trials) to make it robust to the change of seed values or the sampler
implementation.

The new test function compares the results in Optuna's storage and the results returned by the plugin. I think this is a straight forward way to verify the plugin behavior. Since it does not require the global optima, we can use the same number of trials as the current master.

(See also https://github.com/facebookresearch/hydra/pull/1726#discussion_r678299458)

## Related Issues and PRs

The original PR is #1726 .

### DIFF
--- a/plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
+++ b/plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
@@ -1,4 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+import os
 from pathlib import Path
 from typing import Any, List
 
@@ -11,6 +12,7 @@ from hydra.test_utils.test_utils import (
     run_python_script,
 )
 from omegaconf import DictConfig, OmegaConf
+import optuna
 from optuna.distributions import (
     BaseDistribution,
     CategoricalDistribution,
@@ -118,12 +120,16 @@ def test_launch_jobs(hydra_sweep_runner: TSweepRunner) -> None:
 
 @mark.parametrize("with_commandline", (True, False))
 def test_optuna_example(with_commandline: bool, tmpdir: Path) -> None:
+    storage = "sqlite:///" + os.path.join(str(tmpdir), "test.db")
+    study_name = "test-optuna-example"
     cmd = [
         "example/sphere.py",
         "--multirun",
         "hydra.sweep.dir=" + str(tmpdir),
-        "hydra.sweeper.n_trials=100",
-        "hydra.sweeper.n_jobs=8",
+        "hydra.sweeper.n_trials=20",
+        "hydra.sweeper.n_jobs=1",
+        "hydra.sweeper.storage=" + storage,
+        "hydra.sweeper.study_name=" + study_name,
         "hydra/sweeper/sampler=random",
         "hydra.sweeper.sampler.seed=123",
     ]
@@ -134,14 +140,16 @@ def test_optuna_example(with_commandline: bool, tmpdir: Path) -> None:
         ]
     run_python_script(cmd)
     returns = OmegaConf.load(f"{tmpdir}/optimization_results.yaml")
+    study = optuna.load_study(storage=storage, study_name=study_name)
+    best_trial = study.best_trial
     assert isinstance(returns, DictConfig)
     assert returns.name == "optuna"
-    assert returns["best_params"]["x"] == 0
+    assert returns["best_params"]["x"] == best_trial.params["x"]
     if with_commandline:
         assert "y" not in returns["best_params"]
     else:
-        assert returns["best_params"]["y"] == 0
-    assert returns["best_value"] == 0
+        assert returns["best_params"]["y"] == best_trial.params["y"]
+    assert returns["best_value"] == best_trial.value
 
 
 @mark.parametrize("with_commandline", (True, False))
--- a/plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
+++ b/plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
@@ -126,7 +126,7 @@ def test_optuna_example(with_commandline: bool, tmpdir: Path) -> None:
         "example/sphere.py",
         "--multirun",
         "hydra.sweep.dir=" + str(tmpdir),
-        "hydra.sweeper.n_trials=40",
+        "hydra.sweeper.n_trials=20",
         "hydra.sweeper.n_jobs=1",
         f"hydra.sweeper.storage={storage}",
         f"hydra.sweeper.study_name={study_name}",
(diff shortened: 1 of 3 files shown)


plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
```diff
@@ -124,16 +118,18 @@ def test_launch_jobs(hydra_sweep_runner: TSweepRunner) -> None:
...
+    study_name = "test-optuna-example"
     cmd = [
         "example/sphere.py",
         "--multirun",
         "hydra.sweep.dir=" + str(tmpdir),
         "hydra.sweeper.n_trials=20",
-        "hydra.sweeper.n_jobs=8",
+        "hydra.sweeper.n_jobs=1",
+        "hydra.sweeper.storage=" + storage,
+        "hydra.sweeper.study_name=" + study_name,
```
omry, line 132:
```suggestion
        f"hydra.sweeper.storage={storage}",
        f"hydra.sweeper.study_name={study_name}",
```

plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
```diff
@@ -144,18 +140,18 @@ def test_optuna_example(with_commandline: bool, tmpdir: Path) -> None:
...
     assert returns.name == "optuna"
-    assert returns["best_params"]["x"] == 0
+    assert returns["best_params"]["x"] == best_trial.params["x"]
     if with_commandline:
         assert "y" not in returns["best_params"]
     else:
-        assert returns["best_params"]["y"] == 0
-    assert returns["best_value"] == 0
+        assert returns["best_params"]["y"] == best_trial.params["y"]
+    assert returns["best_value"] == best_trial.value
```
omry, line 152: Can we use constants here (with a reasonable epsilon?)
The current testing looks like it will be too lenient to  functional issues in Optuna.

plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
toshihikoyanase: Thanks!

plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
toshihikoyanase: Thank you for your comment.

I understand your concern since the current Optuna repository does not have CI jobs which execute search performance tests.

I tried to estimated the reasonable epsilon with [a notebook](https://colab.research.google.com/drive/1HIvjys_NhinskSR1AFlby0NsLstRlZUD?usp=sharing). I repeated 1000 time of random search (that is the search algorithm used in this example), and found that the 95th percentile value of the best values was 6.25.  

The acceptance range is shown in the following contour plot with the red rectangle. I guess it is reasonably small considering that the search algorithm was random search. What do you think?

image

The suggested change is as follows:
```python
    # 95th percentile with 1000 different seed values.
    assert returns["best_value"] <= 6.25
```

omry: Thanks for evaluating 
Seems like a very large range (and also, if I understand correctly, 95th percentile means it will fail with one in 20 random seeds).

Is there an algorithm that converges faster than a random search we can test with?

toshihikoyanase: I applied the following three changes to the test case.
1. Employ 99the percentile (it will fail with one in 100 random seeds.)
2. Use TPE sampler instead of the random search
3. Increase the number of trials from 20 to 40.

The threshold values were changed as follows:
1. `6.25` to `12.25`
2. `12.25` to `2.27`
3. `2.27` to `0.25`

So, I applied all changes in the commit https://github.com/facebookresearch/hydra/pull/1746/commits/268cbb7ae399d611c5e88e38b0dadbfb200e7886.

The change increased the computation time, but it was still shorter than that of random search with 100 trials.
```
random with 20 trials
2.90s call     plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py::test_optuna_example[False]
2.85s call     plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py::test_optuna_example[True]

random with 100 trials (from #1726)
8.89s call     plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py::test_optuna_example[False]
8.79s call     plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py::test_optuna_example[True]

tpe with 20 trials
2.97s call     plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py::test_optuna_example[False]
2.95s call     plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py::test_optuna_example[True]

tpe with 40 trials
5.15s call     plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py::test_optuna_example[False]
4.82s call     plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py::test_optuna_example[True]
```

omry: Given that the seed is fixed in the test, I think 95th percentile is acceptable.
What bounds do we get for 95th percentile, tpe and 20 iterations?

Go ahead and proceed with whatever values you are comfortable with after this evaluation. I think it will be a good idea to add a comment to this discussion to the test.

toshihikoyanase: Thank you for your suggestion.
The 95th percentile with TPE and 20 iterations is 2.27.
I updated the bound and added a code comment on this discussion in this commit https://github.com/facebookresearch/hydra/pull/1746/commits/e463885adacf2713ddee97f3ef8d381633ff57ed

omry (approved):
lgtm. thanks.

Jieru, can you take over from here?

jieru-hu:
so sorry for getting back to this late! the PR lgtm. Thanks so much!
 @toshihikoyanase  could you rebase master again? I will merge this after that. 

Thanks again!

toshihikoyanase:
@jieru-hu Thank you for your review! I rebased master, but the CI failed at the test of nevergrad as follows:

```console
E   nevergrad.common.errors.InefficientSettingsWarning: DE algorithms are inefficient with budget < 60
=========================== short test summary info ===========================
ERROR tests\test_nevergrad_sweeper_plugin.py - nevergrad.common.errors.Ineffi...
!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
============================== 1 error in 1.74s ===============================
```
Do you have any ideas about it?

jieru-hu:
Thanks @toshihikoyanase 
Looks like the tests were broken by a new warnings introduced by the new nevergrad version released yesterday. 
I will fix the tests and let you know.

jieru-hu:
@toshihikoyanase  the CI has been fixed on the main branch. Could you try rebase? Sorry for the hassle!

toshihikoyanase:
@jieru-hu Thank you for resolving the CI issue! All the CI jobs have passed. Please take a look.

jieru-hu:
many thanks @toshihikoyanase !!

PR #1750, opened 2021-08-03
warning if legacy hydra override comes before group default

### LINKED ISSUE #1748: Legacy overriding of Hydra config groups results in an error if not at the end of the list.
If primary defaults list has an Hydra override that is not at the end of the list, Hydra 1.1 fails to start.
We should make this a warning for smoother migration. 
```
defaults:
  - hydra/launcher: basic
  - db: mysql
```

```
$ python my_app.py
/Users/omry/dev/hydra/hydra/_internal/defaults_list.py:389: UserWarning: In config: Invalid overriding of hydra/launcher:
Default list overrides requires 'override' keyword.
See https://hydra.cc/docs/next/upgrades/1.0_to_1.1/defaults_list_override for more information.

  warnings.warn(msg, UserWarning)
In config: Override 'hydra/launcher : basic' is defined before 'db: mysql'.
Overrides must be at the end of the defaults list

```

Jasha10:
Closes #1748.

- Keep track of legacy overrides separately from non-legacy overrides.
- Issue a warning for out-of-order legacy overrides. As before, raise an Exception for out-of-order non-legacy overrides.

### DIFF
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -385,15 +385,7 @@ def _update_overrides(
                     )
                 )
             elif seen_legacy_override:
-                warnings.warn(
-                    dedent(
-                        f"""\
-                        In {pcp}: Legacy override '{okey} : {oval}' is defined before '{dkey} : {dval}'.
-                        Overrides should be at the end of the defaults list.
-                        """
-                    ),
-                    UserWarning,
-                )
+                pass  # for now, we allow legacy overrides anywhere in the defaults list
 
         if isinstance(d, GroupDefault):
             if legacy_hydra_override:
@@ -404,6 +396,7 @@ def _update_overrides(
                     f"""\
                     In {parent.get_config_path()}: Invalid overriding of {d.group}:
                     Default list overrides requires 'override' keyword.
+                    Note: in Hydra 1.2, overrides will only be allowed to appear at the end of the defaults list.
                     See {url} for more information.
                     """
                 )
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -354,6 +354,7 @@ def _update_overrides(
     interpolated_subtree: bool,
 ) -> None:
     seen_override = False
+    seen_legacy_override = False
     last_override_seen = None
     for d in defaults_list:
         if d.is_self():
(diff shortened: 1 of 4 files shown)


hydra/_internal/defaults_list.py
```diff
@@ -365,20 +366,33 @@ def _update_overrides(
...
             pcp = parent.get_config_path()
             okey = last_override_seen.get_override_key()
             oval = last_override_seen.get_name()
-            raise ConfigCompositionException(
-                dedent(
-                    f"""\
-                    In {pcp}: Override '{okey} : {oval}' is defined before '{d.get_override_key()}: {d.get_name()}'.
-                    Overrides must be at the end of the defaults list"""
+            dkey = d.get_override_key()
+            dval = d.get_name()
```
Jasha10, line 377: I added these two lines to work around linter requirements (line length <= 120 characters).

tests/defaults_list/test_defaults_tree.py
```diff
@@ -831,6 +831,63 @@ def test_legacy_hydra_overrides_from_primary_config_2(
...
+    assert (
+        dedent(
+            """\
+            Legacy override 'hydra/help : custom1' is defined before 'group1 : file1'.
+            Overrides should be at the end of the defaults list.
+            """
+        )
+        in recwarn.list[1].message.args[0]
+    )
+
```
omry, line 889: Instead of producing two warnings, we can produce just the first one.

In the area where we check the order, we can treat hydra overrides as if they are prefixed by `override`, which means they will just produce the first warning.

We can also update the first warning to indicate that overrides should also be at the end of the defaults list if it fits nicely.

What do you think?

Jasha10: > Instead of producing two warnings, we can produce just the first one.
> ...
> We can also update the first warning to indicate that overrides should also be at the end of the defaults list if it fits nicely.

Sounds good. Done in https://github.com/facebookresearch/hydra/pull/1750/commits/96cca574b07ef027e7938c442a311af9b749cede.

> In the area where we check the order, we can treat hydra overrides as if they are prefixed by `override`, which means they will just produce the first warning.

I still think we need to track legacy overrides specially -- normal overrides (with the `override` prefix) will cause `ConfigCompositionException` if they are out of order, and we want to prevent this for the legacy overrides.

hydra/_internal/defaults_list.py
```diff
@@ -390,13 +396,17 @@ def _update_overrides(
...
                     """
                 )
                 deprecation_warning(msg)
 
             if d.override:
-                seen_override = True
+                if legacy_hydra_override:
+                    seen_legacy_override = True
+                else:
+                    seen_override = True
```
omry, line 409: Left overs or actually doing something?

hydra/_internal/defaults_list.py
```diff
@@ -390,13 +390,15 @@ def _update_overrides(
                     f"""\
                     In {parent.get_config_path()}: Invalid overriding of {d.group}:
                     Default list overrides requires 'override' keyword.
+                    Note: in Hydra 1.2, overrides will only be allowed to appear at the end of the defaults list.
```
omry, line 393: This information is appears in red in the migration link. I don't think we need it in the error message.

omry (approved):
Overall looking good. See my comment about the update to the error message.

omry:
So just so we clear on the new behavior.
If a user have a config like:
```yaml
defaults:
  - hydra/help: custom1
  - group1: file1
```

They will get a warning telling them to use override when overriding hydra.
When they fix it, they will get an error telling them overrides have to be at the end of he defaults list.
Is this correct?

I think we need a news fragment, something like:
```
Relaxed handling of legacy hydra overrides in the defaults list to make migration from Hydra 1.0 to 1.1 easier.
```

Jasha10:
> They will get a warning telling them to use override when overriding hydra.
> When they fix it, they will get an error telling them overrides have to be at the end of he defaults list.
> Is this correct?

Yes.

> I think we need a news fragment

Done!

PR #1754, opened 2021-08-04
[ray test ]only assert minor version

jieru-hu:
We've been trying to make sure the micro versions of python matches for the circleCI instance and the AWS instance launched by the ray launcher (due to cloudpickle), however this breaks whenever Python releases a new micro version.
It seems like only making sure the minor version matches works, so changing the check to minor version only.


jieru-hu:
seems like there might be a sev going on with AWS now? I tried to create two test AMIs but they've been stuck in pending for 1hr+

will check back later to see if they are ready


ami-0bb3a0b5a04b966c5 
ami-01a4c55662261264a

Jasha10 (approved):
lgtm

Issue #1765, opened 2021-08-12
[Bug] ConfigKeyError: Key 'defaults' not in <ConfigDataClass>


karthikprasad:
# 🐛 Bug
## Description
Composing structured configs with defaults list results in the above error

## Checklist
- [x] I checked on the latest version of Hydra
- [x] I created a minimal repro (See [this](https://stackoverflow.com/help/minimal-reproducible-example) for tips).

## To reproduce

https://drive.google.com/file/d/193vWa5zkRDBAPtJ55tS9Xlj9bOCEJNl0/view?usp=sharing
**hydra_issue_1765.py**
```python
from dataclasses import dataclass, field
from typing import Any, List, Optional
import hydra
from hydra.core.config_store import ConfigStore
from hydra.utils import instantiate
from omegaconf import OmegaConf, MISSING
from hydra.experimental import initialize, compose

@dataclass
class Foo:
    a: int = 1

@dataclass
class Reducer:
    _target_: str = "__main__.T"
    foo: Foo = MISSING
    defaults: List[Any] = field(default_factory=lambda: [{"/foo": "base_foo"}])

@dataclass
class Trainer:
    _target_: str = "__main__.T"
    reducer: Reducer = MISSING
    defaults: List[Any] = field(default_factory=lambda: [{"/reducer": "base_reducer"}])


class T:
  def __init__(self, **kwargs):
     print(self.__class__, kwargs)

ConfigStore.instance().store(
    name="base_trainer", node=Trainer(), group="trainer"
)
ConfigStore.instance().store(
    name="base_reducer", node=Reducer(), group="reducer"
)
ConfigStore.instance().store(
    name="base_foo", node=Foo(), group="foo"
)

@hydra.main(config_path=None, config_name=None)
def run(cfg):
  print(cfg)

if __name__ == "__main__":
    print(ConfigStore.instance().repo)
    run()
```

To repro, run:
``` bash
python hydra_issue_1765.py +trainer=base_trainer
```

** Stack trace/error message **
```
ConfigKeyError                            Traceback (most recent call last)

/usr/local/lib/python3.7/dist-packages/hydra/_internal/config_loader_impl.py in load_configuration(self, config_name, overrides, run_mode, from_shell)
    149                 run_mode=run_mode,
--> 150                 from_shell=from_shell,
    151             )

55 frames

[...]

PR #1782, opened 2021-08-20
Fixes an edge case with defaults list in a dataclass node that is MISSING

omry:


hydra/_internal/config_loader_impl.py
```diff
@@ -522,6 +522,21 @@ def _compose_config_from_defaults_list(
...
+            if isinstance(cfg, DictConfig):
+                if cfg._is_missing() or cfg._is_none():
+                    return
+                with flag_override(cfg, ["readonly", "struct"], False):
+                    if getattr(cfg, "__REMOVE_TOP_LEVEL_DEFAULTS__", False):
+                        cfg.pop("defaults", None)
+                        cfg.pop("__REMOVE_TOP_LEVEL_DEFAULTS__")
+
+                    for _key, value in cfg.items_ex(resolve=False):
+                        strip_defaults(value)
```
Jasha10, line 536: I don't think the `flag_override` is necessary for this part.
```suggestion
                for _key, value in cfg.items_ex(resolve=False):
                    strip_defaults(value)
```

omry: Did you test without?

Jasha10: Yes, the tests are passing :)

hydra/_internal/config_loader_impl.py
```diff
@@ -522,6 +522,21 @@ def _compose_config_from_defaults_list(
...
                     ).with_traceback(sys.exc_info()[2])
 
+        # # remove remaining defaults lists from all nodes.
+        def strip_defaults(cfg: Any) -> None:
+            if isinstance(cfg, DictConfig):
+                if cfg._is_missing() or cfg._is_none():
+                    return
+                with flag_override(cfg, ["readonly", "struct"], False):
+                    if getattr(cfg, "__REMOVE_TOP_LEVEL_DEFAULTS__", False):
+                        cfg.pop("defaults", None)
```
jieru-hu, line 532: is it possible to have defaults not in cfg yet `__REMOVE_TOP_LEVEL_DEFAULTS__ = True`, that would sound like a bug?

jieru-hu (approved):

hydra/_internal/config_loader_impl.py
omry: If a user decides to add __REMOVE_TOP_LEVEL_DEFAULTS__ to their config.

Jasha10:
Closes #1765

omry:
What's the status of this?

Jasha10:
Sorry for letting this slide. Will merge shortly.

hydra/_internal/config_loader_impl.py
```diff
@@ -522,6 +522,21 @@ def _compose_config_from_defaults_list(
                         f"In '{default.config_path}': Validation error while composing config:\n{e}"
                     ).with_traceback(sys.exc_info()[2])
 
+        # # remove remaining defaults lists from all nodes.
+        def strip_defaults(cfg: Any) -> None:
+            if isinstance(cfg, DictConfig):
```
Jasha10, line 527: It took a while to convince myself that an `if isinstance(cfg, ListConfig)` branch is not needed here.

If a list contains a missing value, you cannot use a higher-level defaults list to merge into that missing value. This is because list-merge results in the whole list being overwritten (rather than in the entries of the list being modified).

I discovered bug https://github.com/omry/omegaconf/issues/788 while working on this.

PR #1791, opened 2021-08-25
disambiguate get_override_key treatment of global

### LINKED ISSUE #1784: [Bug] key collision when using option@_global_ in defaults list
When a defaults list contains entries for both `option` and `option@_global_`, there is a key collision resulting in `ConfigCompositionException`.

Consider the following setup:
```
$ tree
.
├── conf
│   ├── config.yaml
│   └── db
│       ├── mysql.yaml
│       └── sqlite.yaml
└── my_app.py

2 directories, 4 files
```
```python
# my_app.py
import hydra

@hydra.main(config_path="conf", config_name="config")
def my_app(cfg) -> None:
    print(OmegaConf.to_yaml(cfg))

my_app()
```
```yaml
# conf/db/mysql.yaml
name: mysql
```
```yaml
# conf/db/sqlite.yaml
name: sqlite
```
**If `config.yaml` has a defaults list with just one entry, `db: mysql`, the result is as expected:**
```text
$ cat conf/config.yaml
defaults:
  - db: mysql

$ python my_app.py
db:
  name: mysql
```
**If `config.yaml` has a defaults list with just one entry, `db@_global_: sqlite`, the result is as expected:**
```text
$ cat conf/config.yaml
defaults:
  - db@_global_: sqlite

$ python my_app.py
name: sqlite
```
**If `config.yaml` has a defaults list with both of these entries, there is an unexpected key collision:**
```text
$ cat conf/config.yaml
defaults:
  - db: mysql
  - db@_global_: sqlite

$ python my_app.py
Multiple values for db. To override a value use 'override db: sqlite'

```
The expected output in this scenario would be the following:
```text
name: sqlite
db:
  name: mysql
```

Jasha10:
This PR closes #1784 by modifying the `InputDefaults.get_override_key` method.

#### Changes Introduced
Previously, `get_override_key` had this behavior:
- Group `"foo"` and package `"foo"` -> `override key == "foo"`
- Group `"foo"` and package `"bar"` -> `override key == "foo@bar"`
- Group `"foo"` and package `""` -> `override key == "foo"`
With this PR, `get_override_key` has this behavior:
- Group `"foo"` and package `"foo"` -> `override key == "foo"`
- Group `"foo"` and package `"bar"` -> `override key == "foo@bar"`
- Group `"foo"` and package `""` -> `override key == "foo@_global_"`

#### Updates to Pre-Existing Tests
This PR introduces some new tests, and it also updates two old tests that were relying on the old key-collision behavior.
These two tests had to be updated:
```
test_defaults_list.py::test_include_nested_group_global[option_override:include_nested_config_item_global]
test_defaults_tree.py::test_experiment_overriding_global_group[include_absolute_config:override_with_global_default]
```
In the first updated tests, the command-line override `"group1/group2=file2"` was being used, while `"group1/group2@_global_=file2"` should have been used instead.
In the second updated test, the default element `override /group1@_global_: file2` was being used to override group `"group1"` at package `"group1"`, while the element `override /group1@_global_.group1: file2` should have been used instead.

### DIFF
--- a/tests/defaults_list/data/experiment/override_with_global_default2.yaml
+++ b/tests/defaults_list/data/experiment/override_with_global_default2.yaml
@@ -1,0 +1,2 @@
+defaults:
+  - override /group1@_global_: file2
--- a/tests/defaults_list/test_defaults_list.py
+++ b/tests/defaults_list/test_defaults_list.py
@@ -674,6 +674,79 @@ def test_include_nested_group_global(
     "config_name, overrides, expected",
     [
         param(
+            "group_default_at_global",
+            [],
+            [
+                ResultDefault(
+                    config_path="group1/file1",
+                    package="",
+                    parent="group_default_at_global",
+                ),
+                ResultDefault(
+                    config_path="group_default_at_global",
+                    package="",
+                    is_self=True,
+                ),
+            ],
+            id="group_default_at_global",
+        ),
+        param(
+            "group_default_at_global",
+            ["+experiment=override_with_global_default2"],
+            [
+                ResultDefault(
+                    config_path="group1/file2",
+                    package="",
+                    parent="group_default_at_global",
+                ),
+                ResultDefault(
+                    config_path="group_default_at_global",
+                    package="",
+                    is_self=True,
+                ),
+                ResultDefault(
+                    config_path="experiment/override_with_global_default2",
+                    package="experiment",
+                    parent="group_default_at_global",
+                ),
+            ],
+            id="include_absolute_config:override_with_global_default2",
+        ),
+        param(
+            "two_group_defaults_different_pkgs_global",
+            [],
+            [
+                ResultDefault(
+                    config_path="group1/file1",
+                    parent="two_group_defaults_different_pkgs_global",
+                    package="group1",
+                ),
+                ResultDefault(
+                    config_path="group1/file2",
+                    parent="two_group_defaults_different_pkgs_global",
+                    package="",
+                ),
+                ResultDefault(
+                    config_path="two_group_defaults_different_pkgs_global",
+                    package="",
+                    is_self=True,
+                ),
+            ],
+        ),
+    ],
+)
+def test_group_global(
+    config_name: str, overrides: List[str], expected: List[ResultDefault]
+) -> None:
+    _test_defaults_list_impl(
+        config_name=config_name, overrides=overrides, expected=expected
+    )
+
+
+@mark.parametrize(
+    "config_name, overrides, expected",
+    [
+        param(
             "include_nested_group_global_foo",
             [],
             [
--- a/tests/defaults_list/test_defaults_list.py
+++ b/tests/defaults_list/test_defaults_list.py
@@ -639,7 +639,7 @@ def test_group_default_pkg1(
         ),
         param(
             "include_nested_group_global_",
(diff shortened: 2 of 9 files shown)


tests/defaults_list/test_defaults_tree.py
```diff
@@ -682,6 +682,19 @@ def test_override_option_from_defaults_list(
...
             id="two_group_defaults_different_pkgs:override_second",
         ),
+        param(
+            "two_group_defaults_different_pkgs_global",
+            [],
+            DefaultsTreeNode(
+                node=ConfigDefault(path="two_group_defaults_different_pkgs_global"),
+                children=[
+                    GroupDefault(group="group1", value="file1"),
+                    GroupDefault(group="group1", value="file2", package="_global_"),
```
Jasha10, line 692: I've noticed `package="_global_"` showing up in the `GroupDefaults` for the new tests I'm adding to this file.
In all the other tests, a global package shows up as `package=""`.
Is this inconsistency something I should try to fix?

tests/defaults_list/data/experiment/override_with_global_default2.yaml
```diff
@@ -0,0 +1,2 @@
+defaults:
+  - override /group1@_global_: file2
```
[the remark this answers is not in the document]
omry: It's because the file the experiment is overriding, `group_default_at_global`, is defined as 👍🏼 
```yaml
defaults:
  - group1@_global_: file1
```

tests/defaults_list/test_defaults_list.py
```diff
@@ -670,6 +670,79 @@ def test_include_nested_group_global(
...
+                    parent="two_group_defaults_different_pkgs_global",
+                    package="",
+                ),
+                ResultDefault(
+                    config_path="two_group_defaults_different_pkgs_global",
+                    package="",
+                    is_self=True,
+                ),
+            ],
+        ),
```
omry, line 735: missing id for this test.

tests/defaults_list/test_defaults_list.py
```diff
@@ -670,6 +670,79 @@ def test_include_nested_group_global(
...
+                    package="",
+                    is_self=True,
+                ),
+                ResultDefault(
+                    config_path="experiment/override_with_global_default2",
+                    package="experiment",
+                    parent="group_default_at_global",
+                ),
+            ],
+            id="include_absolute_config:override_with_global_default2",
```
omry, line 713: copy paste id? include_absolute_config does not make sense to me.

omry (approved):
A few nits. Overall I am okay with this change.
I believe it will also address the issue you opened about `_here_`.

Don't forget news fragment.

tests/defaults_list/test_defaults_list.py
Jasha10: Yes. Updated in https://github.com/facebookresearch/hydra/pull/1791/commits/806eb0b95082eb1e6398c2d655e189015af94361.

tests/defaults_list/test_defaults_list.py
Jasha10: Added in https://github.com/facebookresearch/hydra/pull/1791/commits/806eb0b95082eb1e6398c2d655e189015af94361.

Jasha10:
This PR also closed #1803

PR #1818, opened 2021-09-08
Optuna get override value element as str

### LINKED ISSUE #1811: [Bug] delete field when using optuna sweeper would throw error
# 🐛 Bug
## Description

I was using optuna sweeper but I noticed that if i delete a field by `~field` it would throw an error.

## To reproduce
Let's [start with this](https://github.com/facebookresearch/hydra/tree/main/examples/advanced/defaults_list_interpolation). Let's modify `conf/db/sqlite.yaml` to be this
```yaml
name: sqlite

model:
  name: boring
```
then create a directory called `exp`, and create a file `exp/exp.yaml`
```yaml
# @package _global_

defaults:
  - override /hydra/sweeper: optuna
  - override /db: sqlite
```
Now run the command `python my_app.py -m +exp=exp ~db.model` it would have the error
> Could not delete from config. The value of 'db.model' is {'name': 'boring'} and not None.

However if I did `python my_app.py +exp=exp ~db.model` (not activating sweeper), then the code would run correctly, making the db to be sqlite (the default is mysql) and the `model.name` part deleted. 

## System information
- **Hydra Version** :  1.0.0.rc1
- **Python version** : 3.7.10
- **Operating system** : Ubuntu 18.04.2 LTS

jieru-hu:
Closes #1811 

following how it is done in [BasicSweeper](https://github.com/facebookresearch/hydra/blob/b16baa73df33e5a2e7ac811983f9ef60e6f130d8/hydra/_internal/core_plugins/basic_sweeper.py#L99-L102) 

It seems like nevergrad also has the same issue, if this looks good, i can follow up with a fix for the other sweepers.


plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
```diff
@@ -153,7 +153,7 @@ def sweep(self, arguments: List[str]) -> None:
         assert self.hydra_context is not None
         assert self.job_idx is not None
 
-        parser = OverridesParser.create()
+        parser = OverridesParser.create(config_loader=self.hydra_context.config_loader)
```
omry, line 156: Is there a reason for this change?

omry:
Are you missing tests and a news fragment?

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
jieru-hu: unintended change, removed :)

Jasha10 (approved):
lgtm!

Issue #1834, opened 2021-09-24
Dict[non-string, ...] crashes when looking for key `__HYDRA_REMOVE_TOP_LEVEL_DEFAULTS__`


Jasha10:
# 🐛 Bug
## Description
There is a bug in the logic to remove top-level defaults lists from structured configs.
Specifically, if a `DictConfig` instance has `key_type` that is typed as something like `int` or as a subclass of `Enum`, then the defaults list logic raises a `ConfigCompositionException` when `getattr(cfg, "__HYDRA_REMOVE_TOP_LEVEL_DEFAULTS__")` is called.

## To reproduce

** Minimal Code/Config snippet to reproduce **
Here is `app.py`:
```python
from enum import Enum
from dataclasses import dataclass, field
from typing import Dict
import hydra
from hydra.core.config_store import ConfigStore

cs = ConfigStore.instance()

class Category(Enum):
    X = 0
    Y = 1
    Z = 2

@dataclass
class Conf:
    enum_dict: Dict[Category, str] = field(default_factory=dict)
    int_dict: Dict[int, str] = field(default_factory=dict)
    str_dict: Dict[str, str] = field(default_factory=dict)

cs.store(name="conf", node=Conf)

@hydra.main(config_path=None, config_name="conf")
def app(cfg):
    assert cfg == {"enum_dict": {}, "int_dict": {}, "str_dict": {}}

if __name__ == "__main__":
    app()
```
Here is the output from running `app.py`:
```text
$ HYDRA_FULL_ERROR=1 python app.py
Traceback (most recent call last):
  File "/home/jasha10/hydra.git/hydra/_internal/config_loader_impl.py", line 140, in load_configuration
    return self._load_configuration_impl(
  File "/home/jasha10/hydra.git/hydra/_internal/config_loader_impl.py", line 248, in _load_configuration_impl
    cfg = self._compose_config_from_defaults_list(
  File "/home/jasha10/hydra.git/hydra/_internal/config_loader_impl.py", line 537, in _compose_config_from_defaults_list
    strip_defaults(cfg)
  File "/home/jasha10/hydra.git/hydra/_internal/config_loader_impl.py", line 535, in strip_defaults
    strip_defaults(value)
  File "/home/jasha10/hydra.git/hydra/_internal/config_loader_impl.py", line 530, in strip_defaults
    if getattr(cfg, "__HYDRA_REMOVE_TOP_LEVEL_DEFAULTS__", False):
[...]

PR #1839, opened 2021-09-30
Fix recent invalid processing of structured configs

pixelb:
Fixes issue #1834

We can't insert an aribtrary "__REMOVE_TOP_LEVEL_DEFAULTS__" key
in the config, as that will fail various base OmegaConf checks,
due to not being a valid enum etc, giving errors like:
  Key '__HYDRA_REMOVE_TOP_LEVEL_DEFAULTS__' is incompatible with the enum type ...

Instead we set that flag using _set_flag on the cfg object.

This was tested with:

  pytest tests/test_compose.py -k test_missing_node_with_defaults_list

### DIFF
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -435,10 +435,7 @@ def _has_config_content(cfg: DictConfig) -> bool:
         return False
 
     for key in cfg.keys():
-        if not OmegaConf.is_missing(cfg, key) and key not in (
-            "defaults",
-            "__HYDRA_REMOVE_TOP_LEVEL_DEFAULTS__",
-        ):
+        if not OmegaConf.is_missing(cfg, key) and key != "defaults":
             return True
     return False
 
--- a/tests/test_compose.py
+++ b/tests/test_compose.py
@@ -3,6 +3,7 @@ import re
 import subprocess
 import sys
 from dataclasses import dataclass, field
+from enum import Enum
 from pathlib import Path
 from textwrap import dedent
 from typing import Any, Dict, List, Optional
@@ -703,3 +704,23 @@ def test_missing_node_with_defaults_list(hydra_restore_singletons: Any) -> None:
 
     cfg = compose("trainer/base_trainer")
     assert cfg == {"trainer": {"reducer": {}}}
+
+
+@mark.usefixtures("initialize_hydra_no_path")
+def test_enum_with_removed_defaults_list(hydra_restore_singletons: Any) -> None:
+    class Category(Enum):
+        X = 0
+        Y = 1
+        Z = 2
+
+    @dataclass
+    class Conf:
+        enum_dict: Dict[Category, str] = field(default_factory=dict)
+        int_dict: Dict[int, str] = field(default_factory=dict)
+        str_dict: Dict[str, str] = field(default_factory=dict)
+
+    cs = ConfigStore.instance()
+    cs.store(name="conf", node=Conf)
+
+    cfg = compose("conf")
+    assert cfg == {"enum_dict": {}, "int_dict": {}, "str_dict": {}}
--- a/hydra/_internal/config_loader_impl.py
+++ b/hydra/_internal/config_loader_impl.py
@@ -527,9 +527,9 @@ class ConfigLoaderImpl(ConfigLoader):
                 if cfg._is_missing() or cfg._is_none():
                     return
                 with flag_override(cfg, ["readonly", "struct"], False):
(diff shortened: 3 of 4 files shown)


hydra/_internal/defaults_list.py
```diff
@@ -435,10 +435,7 @@ def _has_config_content(cfg: DictConfig) -> bool:
         return False
 
     for key in cfg.keys():
-        if not OmegaConf.is_missing(cfg, key) and key not in (
-            "defaults",
-            "__HYDRA_REMOVE_TOP_LEVEL_DEFAULTS__",
-        ):
+        if not OmegaConf.is_missing(cfg, key) and key not in ("defaults",):
```
Jasha10, line 438:
```suggestion
        if not OmegaConf.is_missing(cfg, key) and key != "defaults":
```

omry:
The test you used does not expose the reported issue.
See the minimal repro in #1834 for an example of how to test it.

A few additional points:
Using flags is a nice solution here but they are not designed for it. 
1. Setting a flag invalidates the flags cache in the config subtree which is potentially expensive.
2. The flag should be cleared to avoid contaminated the config (this will get serialized with the config object).
3. Missing news fragment (see [this](https://hydra.cc/docs/development/documentation) for info)
4. Missing test as I pointed out.

Orthogonal to this fix, we should also probably fix OmegaConf to not fail when hasattr is called if the key is mismatching (it should return False).

pixelb:
Yes I should augment the tests here, 
and I see I can remove the flag with _set_flag("blah", None).

@omry re NEWS, this issue wasn't released in a release, so I'm not sure it needs an entry?
I generally avoid adding anything to NEWS inconsequential to users upgrading from vN -> vN+1

omry:
> @omry re NEWS, this issue wasn't released in a release, so I'm not sure it needs an entry?
I generally avoid adding anything to NEWS inconsequential to users upgrading from vN -> vN+1

Ah yeah, good point.
No need to create a news fragment here, you are right.

jieru-hu (approved):
lgtm.

setting a flag looks like a clean solution. fwiw, i got curious and tried to run `tests/benchmark.py` which covers this change with a slight modification:
```
    benchmark.pedantic(
        config_loader.load_configuration,
        kwargs={
            "config_name": "config",
            "overrides": [],
            "run_mode" : RunMode.RUN,
        },
        iterations=10,
        rounds=10,
    )
```

### on main
```
----------------------------------------------- benchmark: 1 tests -----------------------------------------------
Name (time in ms)           Min      Max     Mean  StdDev   Median      IQR  Outliers      OPS  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------
test_config_loading     66.9423  81.3681  75.4913  5.4585  78.1522  10.0342       4;0  13.2466      10          10
------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
```
### on top of this PR
```
----------------------------------------------- benchmark: 1 tests ----------------------------------------------
Name (time in ms)           Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------
test_config_loading     54.6083  65.1431  57.5654  3.0593  57.1400  3.3454       1;1  17.3716      10          10
-----------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
============================================================================================================================
```

it looks like for the basic scenarios that the benchmark covers, there does not seem to be an regression.

Jasha10 (approved):
Lgtm!

tests/test_compose.py
```diff
@@ -703,3 +704,23 @@ class Trainer:
...
+    class Conf:
+        enum_dict: Dict[Category, str] = field(default_factory=dict)
+        int_dict: Dict[int, str] = field(default_factory=dict)
+        str_dict: Dict[str, str] = field(default_factory=dict)
+
+    cs = ConfigStore.instance()
+    cs.store(name="conf", node=Conf)
+
+    cfg = compose("conf")
+    assert cfg == {"enum_dict": {}, "int_dict": {}, "str_dict": {}}
```
omry, line 726: To confirm, this test fails before you apply the fix, right?

omry (approved):
lgtm, thanks.

tests/test_compose.py
pixelb: Yes it does fail before the fix

Issue #1815, opened 2021-09-06
Support optuna.samplers.GridSampler


GillianGrayson:
Hi!

I tried to use `optuna.samplers.GridSampler` for some model with this example:

```yaml
defaults:
  - override /hydra/sweeper: optuna

hydra:
  sweeper:
    _target_: hydra_plugins.hydra_optuna_sweeper.optuna_sweeper.OptunaSweeper
    storage: null
    study_name: null
    n_jobs: 1

    sampler:
      _target_: optuna.samplers.GridSampler
      search_space: {"model.n_d": [ 4, 8, 16 ]}
```
but I got this error:

```
Key 'search_space' not in 'TPESamplerConfig'
    full_key: hydra.sweeper.sampler.search_space
    reference_type=SamplerConfig
    object_type=TPESamplerConfig
```

It is strange, because I create GridSampler, not TPESampler.

## System information
- **Hydra Version** : 
hydra-core==1.1.0
hydra-optuna-sweeper==1.1.0 
- **Python version** : 3.8 
- **Operating system** : Windows 10

GillianGrayson:
As far as I understood, `GridSampler` isn't mentioned here:
https://github.com/facebookresearch/hydra/blob/main/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py

Is this sampler supported?

jieru-hu:
hi @GillianGrayson 
you are right, `GridSampler`  is not supported. Otherwise it'd be listed in the config.py.

Supporting this should be relatively straightforward (it would require adding a [config dataclass](https://github.com/facebookresearch/hydra/blob/main/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py#L27-L42) and [registering it with the config store](https://github.com/facebookresearch/hydra/blob/main/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py#L199-L204))

jieru-hu:
While `GridSampler` is not supported yet, Hydra's `BasicSweeper` actually supports grid search out of the box with a rich override syntax available - the only thing missing is you cannot define the sweep in a config file (which we will be evaluating for next Hydra release.)

https://hydra.cc/docs/tutorials/basic/running_your_app/multi-run
https://hydra.cc/docs/advanced/override_grammar/basic

cc @GillianGrayson

PR #1840, opened 2021-10-02
Add GridSampler to Optuna Sweeper

beerzyp:

Yes

## Related Issues and PRs

Add GridSampler support, issue https://github.com/facebookresearch/hydra/issues/1815.

### DIFF
--- a/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
+++ b/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
@@ -102,6 +102,17 @@ class MOTPESamplerConfig(SamplerConfig):
 
 
 @dataclass
+class GridSamplerConfig(SamplerConfig):
+    """
+    https://optuna.readthedocs.io/en/stable/reference/generated/optuna.samplers.GridSampler.html
+    """
+
+    _target_: str = "optuna.samplers.GridSampler"
+
+    search_space: Optional[Dict[str, Any]] = None
+
+
+@dataclass
 class DistributionConfig:
 
     # Type of distribution. "int", "float" or "categorical"
@@ -200,5 +211,12 @@ ConfigStore.instance().store(
     group="hydra/sweeper/sampler",
     name="motpe",
     node=MOTPESamplerConfig,
+    provider="optuna_sweeper",
+)
+
+ConfigStore.instance().store(
+    group="hydra/sweeper/sampler",
+    name="grid",
+    node=GridSamplerConfig,
     provider="optuna_sweeper",
 )


Jasha10:
@beerzyp thanks very much for the PR, we will review it shortly :)

jieru-hu:
thanks @beerzyp  for the PR.

I assume you are following my recommendation in #1815 - in which I didn't lay out everything that needs to be changed. I should have put more details there. sorry.

1. the change doesn't actually run - if you try to run an application with sampler=grid, you'd see
```
  File "/Users/jieru/workspace/hydra-fork/hydra/hydra/_internal/instantiate/_instantiate2.py", line 62, in _call_target
    return _target_(*args, **kwargs)
TypeError: Error instantiating 'optuna.samplers._grid.GridSampler' : __init__() got an unexpected keyword argument 'seed'
```
the reason being GridSamplerConfig cannot extend SamplerConfig - it does not accept a `seed` param.

2. GridSampler fits a bit awkwardly with the Optuna config: 1) it expects a `search_space` param to be passed in to the init the class. 2) Optuna already has a `hydra.sweeper.search_space` config available. 

A bit more context on how the Optuna sweeper works:
the Optuna config has [custom logic](https://github.com/facebookresearch/hydra/blob/main/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py#L35-L60) parsing `hydra.sweeper.search_space`. To support `GridSampler`, I think we will need some custom logic in the Optuna sweeper to check if the Sampler is a GridSampler and if so, override the search_space params accordingly. If we directly pass a list of params to the Sampler like how it is implemented now, we actually miss out all the power Hydra's [override syntax](https://hydra.cc/docs/advanced/override_grammar/basic) provides and the behavior doesn't align well with the Optuna sweeper's current user experience. 

While GridSampler is not supported yet, Hydra's BasicSweeper actually provides a grid search out of the box, here is the doc
https://hydra.cc/docs/advanced/override_grammar/extended#sweeps
https://hydra.cc/docs/tutorials/basic/running_your_app/multi-run

jieru-hu (changes requested):
see my comment.

cschell:
@jieru-hu, I have a question about that statement

> While GridSampler is not supported yet, Hydra's BasicSweeper actually provides a grid search out of the box

This is not true for distributed scenarios, right? So, for example, I run multiple optuna-sweeper instances within Kubernetes jobs, all instances coordinated by a central optuna-RDBStorage. If I understand this correctly, I have no way to coordinate the hyperparameter-runs when I use the basic hydra sweeper, right?

jieru-hu:
@cschell sorry I just saw your comment! Yes, you are right - I didn't consider the Distributed scenarios in this case. thanks for pointing that out!

jieru-hu:
closing due to inactivity. we are going to support this in the upcoming release.

PR #1842, opened 2021-10-02
Implement shell completion for +group and ~group

### LINKED ISSUE #1841: Improvements to shell completion
# 🚀 Feature Request

Hydra's [command line completion](https://hydra.cc/docs/tutorials/basic/running_your_app/tab_completion/#internaldocs-banner) currently does not implement completions for [appending to or deleting from the defaults list](https://hydra.cc/docs/advanced/override_grammar/basic#modifying-the-defaults-list) (e.g. `+db=mysql` or `~db` or `~db=mysql`).

```text
# this works:
python my_app.py group=opt{TAB} -> group=option
# this does not work:
python my_app.py +group=opt{TAB} -> ...
```

Jasha10:
This PR adds logic Hydra's shell-completion, handling the cases of '+group' and '~group'.
Closes #1841.

Completions enabled by this PR:
- `+gro` -> `+group=`  (just like if no plus is given)
- `+group=` -> `+group=option`  (just like if no plus is given)
- `+` -> `+group=`
- `~gro` -> `~group`  (note no trailing equal sign) 
- `~group=` -> `~group=option`  (just like if no tilde is given)
- `~` -> `~group`

Edit:
I was not able to get the tilde completions ('~group...') working with fish or zsh; those shells have special completion logic for words starting with tilde.

### DIFF
--- a/hydra/plugins/completion_plugin.py
+++ b/hydra/plugins/completion_plugin.py
@@ -158,6 +158,10 @@ class CompletionPlugin(Plugin):
         return matches
 
     def _query_config_groups(self, word: str) -> Tuple[List[str], bool]:
+        if word.startswith("+") or word.startswith("~"):
+            prefix, word = word[0], word[1:]
+        else:
+            prefix = ""
         last_eq_index = word.rfind("=")
         last_slash_index = word.rfind("/")
         exact_match: bool = False
@@ -197,6 +201,7 @@ class CompletionPlugin(Plugin):
                         name = name + "/"
                     matched_groups.append(name)
 
+        matched_groups = [f"{prefix}{group}" for group in matched_groups]
         return matched_groups, exact_match
 
     def _query(self, config_name: Optional[str], line: str) -> List[str]:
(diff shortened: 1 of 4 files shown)


Jasha10:
It seems that CI is failing because zsh is treating the tilde character '~' in a special way.

pixelb (commented):
I see bash and zsh are different wrt ~ expansion.

bash will fall back to treating ~blah as literal if there is no blah user,
while zsh will give an error.

I.e. bash could start giving errors if there was a ~hydra or ~group user on the system.

I wonder would the most general fix be to quote the ~,
i.e. with \\~hydra etc. in the tests to avoid this issue.

One is generally going to need to quote tilde while on the command line anyway.

omry:
> One is generally going to need to quote tilde while on the command line anyway.

For bash at least, I never had to quote `~` (As you are saying, it might conflict in the rare case that there is a user matching the config group name. 
if zsh is more strict about it we can just exclude it from the test.

pixelb (approved):
This change is fine in isolation.

The test adjustments can be done separately

Issue #1843, opened 2021-10-04
[Bug] LexerNoViableAltException: word.p=a|b (error comes from the vertical bar)


zhaoyanpeng:
hi there,

looks hydra does not support escaping a vertical bar? I checked the doc and it says

> ... by escaping them with a `\`. These characters are: `\()[]{}:=, \t` (the last two ones being the whitespace and tab characters).

any reason for not escaping vertical bar? one of my running parameters is a regex expression, which contains a vertical bar, but I cannot send it in from the command line, e.g, `python main.py word.p=a\|b`

Jasha10:
I can reproduce the issue:
```text
$ cd ~/hydra.git/examples/tutorials/basic/your_first_hydra_app/1_simple_cli/
$ python my_app.py '+foo=ab'
foo: ab

$ python my_app.py '+foo=a|b'
LexerNoViableAltException: +foo=a|b
                                 ^
See https://hydra.cc/docs/next/advanced/override_grammar/basic for details

Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.
$ python my_app.py '+foo=a\|b'
LexerNoViableAltException: +foo=a\|b
                                  ^
See https://hydra.cc/docs/next/advanced/override_grammar/basic for details

Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.
```

jieru-hu:
the following works for me:
```
$ python examples/tutorials/basic/your_first_hydra_app/2_config_file/my_app.py '+foo="a|b"'
db:
  driver: mysql
  user: omry
  password: secret
foo: a|b

```

zhaoyanpeng:
thanks @jieru-hu , that works! quoting only the right-hand-side also works: `python main.py word.p='"a|b"'`

PR #1844, opened 2021-10-04
Docs: add notes on command line quoting

Jasha10:
This PR adds two notes to the documentation on command-line overrides.
Related to #1843.

### DIFF
--- a/website/docs/advanced/override_grammar/basic.md
+++ b/website/docs/advanced/override_grammar/basic.md
@@ -147,6 +147,19 @@ Quoted strings can accept any value between the quotes, but some characters need
 </div>
 </div>
 
+It may be necessary to use multiple pairs of quotes to prevent your
+shell from consuming quotation marks before they are passed to hydra.
+
+```shell
+$ python my_app.py +foo="{a: 10}"  # the string `+foo={a: 10}` is passed to Hydra by bash
+foo:
+  a: 10
+
+$ python my_app.py '+foo="{a: 10}"'  # the string `+foo="{a: 10}"` is passed to Hydra by bash
+foo: '{a: 10}'
+
+```
+
 ### Whitespaces in unquoted values
 Unquoted Override values can contain non leading or trailing whitespaces.
 For example, `msg=hello world` is a legal override (key is `msg` and value is the string `hello world`).
--- a/website/docs/advanced/override_grammar/basic.md
+++ b/website/docs/advanced/override_grammar/basic.md
@@ -196,11 +196,13 @@ As an example, in the following `dir` is set to the string `job{a=1,b=2,c=3}`:
 $ python my_app.py 'dir=job\{a\=1\,b\=2\,c\=3\}'
 ```
 
-Some characters, such as the bar symbol `'|'`, are not legal in unquoted strings.
-Strings containing such characters must be quoted:
+Hydra's parser considers some characters, such as the left-bracket symbol `'['`, to be illegal in unquoted strings.
+Strings containing such characters must be quoted, or the offending character(s) escaped with backslashes:
 
 ```shell
-$ python my_app.py 'dir="A|B"'
+$ python my_app.py 'dir=A[B'    # error
+$ python my_app.py 'dir="A[B"'  # ok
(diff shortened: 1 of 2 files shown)


omry:
I suspect `|` can be supported in unquoted strings as we are not using it for anything else in the grammar.

website/docs/advanced/override_grammar/basic.md
```diff
@@ -173,6 +186,13 @@ As an example, in the following `dir` is set to the string `job{a=1,b=2,c=3}`:
 $ python my_app.py 'dir=job\{a\=1\,b\=2\,c\=3\}'
 ```
 
+Some characters, such as the bar symbol `'|'`, are not legal in unquoted strings.
+Strings containing such characters must be quoted:
+
+```shell
+$ python my_app.py 'dir="A|B"'
+```
+
```
pixelb, line 195: Since "|" is also special in shell, it would be good to distinguish to what "|" is not legal.

website/docs/advanced/override_grammar/basic.md
```diff
@@ -147,6 +147,19 @@ Quoted strings can accept any value between the quotes, but some characters need
 </div>
 </div>
 
+It may be necessary to use multiple pairs of quotes to prevent your
+shell from consuming quotation marks before they are passed to hydra.
+
+```shell
+$ python my_app.py +foo="{a: 10}"  # the string `+foo={a: 10}` is passed to Hydra by bash
+foo:
+  a: 10
```
jieru-hu, line 156: I think this might be confusing. since we recommend quote key=value pare in single quote, maybe update this to 
```
python my_app.py '+foo={a: 10}'
```

Jasha10: Updated in https://github.com/facebookresearch/hydra/pull/1844/commits/d30001657e95418436b6bb234d02131c2b29b2ac.

Jasha10:
Rebased to incorporate recent CI fixes on the `main` branch.

Jasha10:
> This is one of the most common questions we get - here is a list of tips from Omry (copy pasted from discussions before), could you also include them in the doc?
> ...

Done in https://github.com/facebookresearch/hydra/pull/1844/commits/df625f866a2622f09ba19940065f7c31fdf8b4d8.

website/docs/advanced/override_grammar/basic.md
Jasha10: PR #1850 is merged, so Hydra's parser now accepts the pipe character in unquoted strings.
I've updated the above to use the left-bracket character `'['` and have added some language to disambiguate that Hydra's parser is rejecting the character (as opposed to the shell rejecting it).

website/docs/advanced/override_grammar/basic.md
```diff
@@ -147,6 +147,29 @@ Quoted strings can accept any value between the quotes, but some characters need
...
+Here are some best practices around quoting in CLI overrides:
+- Quote the whole key=value pair with single quotes, as in the first two
+  examples above. These quotes are for the benefit of the shell.
+- Do not quote keys.
+- Only quote values if they contain a space. It will work if you always quote
+  values, but it will turn numbers/dicts/lists into strings (as in the first
+  example above).
+- When you are quoting values, use double quotes to avoid collision with the
+  outer single quoted consumed by the shell.
+
```
pixelb, line 172: These are good basic rules that should cater for almost all cases.

BTW double quoting in the shell is more general than single quoting, as you can specify any character through the use of escapes (you can't have literal single quotes within single quoted args for example). But I wouldn't bother mentioning that in the doc as it's shell specific

pixelb (approved):
I see you've squashed the 1.1 update into this also.
In general it's probably best to cherry-pick a separate commit to the 1.1 branch

jieru-hu:
@pixelb 
right now we only deploy the doc changes on main branch ([link](https://github.com/facebookresearch/hydra/blob/main/.github/workflows/deploy_website.yml)), so I think it is OK to update the doc changes in one commit, and probably fine not to cherry-pick to 1.1_branch (since those changes won't be deployed in the current setup.) What do you think?

PR #1848, opened 2021-10-09
Support escaped tilde completion

Jasha10:
This PR is a follow-up to #1842.
The goal is to enable shell completion to work for deletions that are preceded
by a backslash escape character (e.g. `\~group`).

As of this writing, I've got the following working:

In zsh:
  - `\~` -> `\~group`
  - `\~gro` -> `\~group`

In fish:
  - `\~` -> `\~group`
  - `\~gro` -> `\~group`

In bash: I have not been able to get the same behavior.
I did manage to get the following bash completions working (note the completion does not have a backslash):
  - `\~` -> `~group`
  - `\~gro` -> `~group`
However, I decided not to include this behavior in the PR because I do not
think it is useful (as removing the backslash from the completion would defeat
the point of escaping the tilde).

### DIFF
--- a/hydra/_internal/core_plugins/zsh_completion.py
+++ b/hydra/_internal/core_plugins/zsh_completion.py
@@ -18,7 +18,7 @@ class ZshCompletion(CompletionPlugin):
         self.delegate = BashCompletion(config_loader)
 
     def install(self) -> None:
-        self.delegate.install(shell="zsh")
+        self.delegate.install("zsh")
 
     def uninstall(self) -> None:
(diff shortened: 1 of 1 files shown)


omry:
My understanding was that this was about changing the tests to pass `\~` to shells that needs it instead of changing actual completion logic.

pixelb:
Right just confirming Omry is correct here.
I.e. I was only suggesting to adjust the tests, so the shells would be doing unambiguous processing.
Hydra should only be dealing with unescaped ~

PR #1850, opened 2021-10-10
grammar: Allow pipe character `|` in unquoted strings

Jasha10:
This PR is a follow-up to Omry's comment [here](https://github.com/facebookresearch/hydra/pull/1844#issuecomment-933897125) suggesting that Hydra's override grammar allow the pipe character `|` in unquoted strings.

If we decide to move forward with this PR, I think we should update [OmegaConf's grammar](https://github.com/omry/omegaconf/blob/master/omegaconf/grammar/OmegaConfGrammarLexer.g4) too.


tests/test_overrides_parser.py
```diff
@@ -84,6 +84,8 @@ def eq(item1: Any, item2: Any) -> bool:
         # bool
         param("true", True, id="value:bool"),
         param(".", ".", id="value:dot"),
+        param("|", "|", id="value:pipe"),
+        param("a|b", "a|b", id="value:a-pipe-b"),
```
odelalleau, line 88: To be consistent with testing of over special characters in unquoted strings, I'd suggest to just add | to UNQUOTED_SPECIAL (https://github.com/facebookresearch/hydra/blob/f797eadc5d3245a282829ce9e50899ced0c035a4/tests/test_overrides_parser.py#L34) instead of adding these two tests.

odelalleau (changes requested):
Looks good to me, but yes we'll need the same change in OmegaConf's grammar as well.

odelalleau:
Oh, can you add a news fragment as well?

tests/test_overrides_parser.py
Jasha10: Done in https://github.com/facebookresearch/hydra/pull/1850/commits/96df2e0c94a5f32827a6c3adc0d9a6a7999f4abd. And I can confirm that the tests fail on the `main` branch.

PR #1870, opened 2021-10-28
Ax sweeper - sampling in log space

SiddhantRanade:
## Motivation

Allow users to use log_space from Ax with hydra

Yes


SiddhantRanade:
This works now!

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
```diff
@@ -312,6 +312,7 @@ def create_range_param_using_interval_override(override: Override) -> Dict[str,
         "name": key,
         "type": "range",
         "bounds": [value.start, value.end],
+        "log_scale": "log" in value.tags,
```
Jasha10, line 315: I would slightly prefer to use `"log_scale": "log_scale" in value.tags` instead of `"log_scale": "log" in value.tags` here. The motivation is:
  1) that way the API would be consistent between using the command line vs using yaml files, and
  2) there would be less chance for users to be confused by thinking that they are e.g. taking the logarithm of the values (which is not the case).

What do you think, @SiddhantRanade?

Jasha10 (changes requested):
Overall it looks good!
One point regarding the API:

plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
Jasha10: Please disregard my previous comment -- I hadn't seen pixelb's comment above. It seems that the `log` tag is already being used by other sweepers (nevergrad and optuna).

Jasha10 (approved):

SiddhantRanade:
`tag(log, interval(a,b))` was chosen because that is consistent with the
Optuna sweeper plugin and shorter to type on the command line. For those
reasons I think it's fine as is.

In a config file we would need to have `log_scale: true` because that is
what Ax expects.

The API is inconsistent either way because on the command line we need to
use `interval` while in a config file we need to use `range`. I think it's
perfectly fine as is. The command line specification follows "the Hydra
way" while the config file specification follows whatever the specific
sweeper expects.

@Jasha10 Let me know if you still think we should change the tag to
`log_scale`.

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -107,6 +107,33 @@ def test_jobs_configured_via_config(hydra_sweep_runner: TSweepRunner) -> None:
         assert math.isclose(best_parameters["quadratic.y"], -1.0, abs_tol=1e-4)
 
 
+def test_jobs_configured_via_config_logscale(hydra_sweep_runner: TSweepRunner) -> None:
```
jieru-hu, line 110: thanks for adding the tests. 

there is a lot of duplication going on here, could you use [pytest params](https://docs.pytest.org/en/6.2.x/example/parametrize.html) to avoid all these duplication?

for example, it seems that `param` and the `overrides` are the only params that are different, you could set those two as pytest parameters to avoid copying the whole test. you can see an example in the same test file https://github.com/facebookresearch/hydra/blob/main/plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py#L36-L52

same comment goes for `test_jobs_configured_via_cmd_logscale`

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -107,6 +107,33 @@ def test_jobs_configured_via_config(hydra_sweep_runner: TSweepRunner) -> None:
...
+        assert sweep.returns is None
+        returns = OmegaConf.load(f"{sweep.temp_dir}/optimization_results.yaml")
+        assert isinstance(returns, DictConfig)
+        assert returns["optimizer"] == "ax"
+        assert len(returns) == 2
+        best_parameters = returns.ax
+        # the point of this test is to make sure that Ax runs with logscale.
+        # It's hard to get the correct answer just trying values.
+        assert "quadratic.x" in best_parameters
+        assert "quadratic.y" in best_parameters
```
jieru-hu, line 134: i set 
```
quadratic.x:
  type: range
  bounds: [1e-6, 1]
  log_scale: True
```
for log scale, and `math.isclose(best_parameters["quadratic.x"], 0.0, abs_tol=1e-4)` returns True :)

jieru-hu (changes requested):
thanks for updating the PR. pls see comments inline. :)

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -154,6 +211,7 @@ def test_jobs_configured_via_cmd_and_config(hydra_sweep_runner: TSweepRunner) ->
         assert len(returns) == 2
         best_parameters = returns.ax
         assert math.isclose(best_parameters["quadratic.x"], -2.0, abs_tol=1e-4)
+        # TODO: y should actually be -1
```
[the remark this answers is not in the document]
SiddhantRanade: Yes, if you look at the optimization problem, the solution is y=-1 in this case. Since the sweeper only does a few tries, it happens to output +1. I don't think this is the right way to test.

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
```diff
@@ -129,9 +156,39 @@ def test_jobs_configured_via_cmd(hydra_sweep_runner: TSweepRunner) -> None:
         assert len(returns) == 2
         best_parameters = returns.ax
         assert math.isclose(best_parameters["quadratic.x"], -2.0, abs_tol=1e-4)
+        # TODO: y should actually be -2
```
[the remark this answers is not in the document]
SiddhantRanade: Similar to the above case.

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
SiddhantRanade: Sure, but see the above comments. I don't think it's right to actually expect the correct answer to this optimization problem with the sweeper in this way.

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
SiddhantRanade: Will fix.

Jasha10:
> @Jasha10 Let me know if you still think we should change the tag to `log_scale`.

No, don't change it; your reasoning above makes sense. I hadn't realized that the `log` tag is already being used by other sweepers.

Thanks @SiddhantRanade.

plugins/hydra_ax_sweeper/tests/test_ax_sweeper_plugin.py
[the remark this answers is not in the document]
jieru-hu: The tests have been stable  and it would be hard to predict the sweeper's behavior beyond a couple of steps so the tests make the best efforts here.

Please do feel free to suggest alternatives via a issue or a PR. Adding a TODO here makes it harder to track.  

Thanks!

jieru-hu:
closing due to no activity. i've create a new PR here https://github.com/facebookresearch/hydra/pull/2018

PR #1874, opened 2021-11-01
Fix CI.

jieru-hu:
`aiohttp` (ray's dependency) released 3.8.0 yesterday, it is installing an `examples` top level package and as a result make it not possible for the hydra's examples module being discovered. for now, im pinning the version. 
This issue has already been fixed in aiohttp and should become available in the next release.


jieru-hu:
minor change. merging...

pixelb:
is this needed on 1.1 branch also?

jieru-hu:
Yes :) will create a PR today

PR #1878, opened 2021-11-06
unpin mypy

### LINKED ISSUE #1326: Upgrade mypy
we pinned mypy==0.790 for now.
more info can be found here https://github.com/facebookresearch/hydra/pull/1325

Jasha10:
Closes #1326

Edit:
I've updated this PR and the tests are now passing.
Commits:
- https://github.com/facebookresearch/hydra/commit/e7816cc83fe4e277a0f1c9afff64774de9cbbd10: unpin mypy
- https://github.com/facebookresearch/hydra/commit/27bd6fc3e44502090f734323ba4ea754d1f7dbe6: update noxfile.py and .mypy.ini
- https://github.com/facebookresearch/hydra/commit/2365aaa4ae5252b2af6f784b1e196f9cc2606b95: modify source files so that newer mypy version will pass

Comments on this diff:
The latest version of mypy errors if it detects multiple modules with the same name ([mypy#4008](https://github.com/python/mypy/issues/4008)). This means simultaneously checking many files named `my_app.py` will result in a mypy error (assuming that there is no `__init__.py` next to `my_app.py`). Guido van Rossum commented on mypy#4008, saying "You should really run mypy separately over each package in that case". This motivates many of the changes made in the second and third commits to this PR.


Jasha10:
Related to the failure on the most recent commit: https://github.com/python/mypy/issues/4008

Jasha10:
Rebased against `main` to incorporate recent code updates.

pixelb (approved):

tools/release/release.py
```diff
@@ -123,17 +123,20 @@ def build_package(cfg: Config, pkg_path: str, build_dir: str) -> None:
...
+        assert cur.pre is not None
         prefix = cur.pre[0]
         num = cur.pre[1] + 1
         new_version = f"{cur.major}.{cur.minor}.{cur.micro}.{prefix}{num}"
     elif cur.is_postrelease:
-        prefix = cur.post[0]
-        num = cur.post[1] + 1
+        assert cur.post is not None
+        num = cur.post + 1
         new_version = f"{cur.major}.{cur.minor}.{cur.micro}.{prefix}{num}"
```
pixelb, line 140: This looks a bit strange. prefix is mentioned in the f string, but no longer set?
Also why change cur.post[1] to cur.post ?

pixelb (approved):

tools/release/release.py
Jasha10: > Also why change cur.post[1] to cur.post?

The type of `cur.post` is `Optional[int]`, so `cur.post[1]` does not make sense.
As for why the previous version had `cur.post[0]` and `cur.post[1]`: I suspect that this branch of the if-statement was copy/pasted from the `if cur.is_prerelease` branch above.

There is currently no test coverage for this `release.py` file.

Jasha10: > This looks a bit strange. prefix is mentioned in the f string, but no longer set?

Updated in https://github.com/facebookresearch/hydra/pull/1878/commits/f0d30ae370a0cd2faed886fcba7070ee91a40fa9. Thanks!

PR #1886, opened 2021-11-11
Support defining basic sweep through config file

### LINKED ISSUE #1376: [Feature Request] Define a sweep in config.yaml file rather than command line
# 🚀 Feature Request

Currently a sweep can only be defined on the command line. Could it be supported through the config files as well?

## Motivation / pitch
Situations where this particularly makes sense: `python train.py seed="range(0,10)" xval_fold="range(0,5)" actual_hyperparam_I_want_to_change=1,2,3`
-- if for any new hyperparameter combination we always want to run over those several seeds and cross-validation folds.

For those cases I'd like to be able to have those sweeps in `config.yaml`, to declutter the command line.

Couldn't find prior discussions about this, but somewhat related is #394 

cc @liujas000

jieru-hu:
Closes #1376 

part of the efforts to standardize how we configure Sweeper's search space for all Hydra Sweepers.

### DIFF
--- a/hydra/_internal/core_plugins/basic_sweeper.py
+++ b/hydra/_internal/core_plugins/basic_sweeper.py
@@ -20,7 +20,7 @@ import itertools
 import logging
 import time
 from collections import OrderedDict
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional, Sequence
 
@@ -56,7 +56,9 @@ class BasicSweeper(Sweeper):
     Basic sweeper
     """
 
-    def __init__(self, max_batch_size: Optional[int], params: Optional[Dict[str, str]] = None) -> None:
+    def __init__(
+        self, max_batch_size: Optional[int], params: Optional[Dict[str, str]] = None
+    ) -> None:
         """
         Instantiates
         """
--- a/hydra/_internal/core_plugins/basic_sweeper.py
+++ b/hydra/_internal/core_plugins/basic_sweeper.py
@@ -19,9 +19,10 @@ python foo.py a=range(1,4) b=10,20
 import itertools
 import logging
 import time
-from dataclasses import dataclass
+from collections import OrderedDict
+from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, Iterable, List, Optional, Sequence
+from typing import Any, Dict, Iterable, List, Optional, Sequence
 
 from omegaconf import DictConfig, OmegaConf
 
@@ -39,6 +40,7 @@ from hydra.types import HydraContext, TaskFunction
 class BasicSweeperConf:
     _target_: str = "hydra._internal.core_plugins.basic_sweeper.BasicSweeper"
     max_batch_size: Optional[int] = None
+    params: Dict[str, str] = field(default_factory=dict)
 
 
 ConfigStore.instance().store(
@@ -54,7 +56,7 @@ class BasicSweeper(Sweeper):
     Basic sweeper
     """
 
-    def __init__(self, max_batch_size: Optional[int]) -> None:
+    def __init__(self, max_batch_size: Optional[int], params: Dict[str, str]) -> None:
         """
         Instantiates
         """
@@ -62,6 +64,7 @@ class BasicSweeper(Sweeper):
         self.overrides: Optional[Sequence[Sequence[Sequence[str]]]] = None
         self.batch_index = 0
         self.max_batch_size = max_batch_size
+        self.params = params
 
         self.hydra_context: Optional[HydraContext] = None
         self.config: Optional[DictConfig] = None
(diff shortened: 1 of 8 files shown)


website/docs/tutorials/basic/running_your_app/2_multirun.md
```diff
@@ -21,6 +25,27 @@ The following sweeps over all combinations of the dbs and schemas.
...
+```yaml
+hydra:
+  sweeper:
+    params:
+      db: mysql,postgresql
+      schema: warehouse,support,school
+```
+
+The syntax are consistent for both input configs and commandline overrides. The commandline will override the definition 
+in input configs. If we run the same application with the above input config and a new commandline override:
```
Jasha10, line 40: I think that the word "override" is overloaded (as here it refers both to the sweeper overriding the config settings in different jobs, and to the command line sweep taking precedence over the the sweep from the input config).
```suggestion
The syntax are consistent for both input configs and commandline overrides.
If a sweep is specified in both an input config and at the command line,
then the commandline sweep will take precedence over the sweep defined 
in the input config. If we run the same application with the above input config and a new commandline override:
```

hydra/_internal/core_plugins/basic_sweeper.py
```diff
@@ -54,14 +56,15 @@ class BasicSweeper(Sweeper):
     Basic sweeper
     """
 
-    def __init__(self, max_batch_size: Optional[int]) -> None:
+    def __init__(self, max_batch_size: Optional[int], params: Dict[str, str]) -> None:
```
Jasha10, line 59: Is it possible that there is some downstream code that uses or inherits from `BasicSweeper`? If so, we could supply a default value for `params` so that this change will be less breaking to callers of `BasicSweeper.__init__`.
Here's what I'm thinking:
```python
def __init__(self, max_batch_size: Optional[int], params: Optional[Dict[str, str]] = None) -> None:
    if params is None:
        params = {}
```

hydra/_internal/core_plugins/basic_sweeper.py
```diff
@@ -101,12 +104,13 @@ def split_arguments(
     ) -> List[List[List[str]]]:
 
         lists = []
+        merged_overrides = OrderedDict()
```
Jasha10, line 107: I think the word "merged" here could be confusing (as no call is being made to `OmegaConf.merge`).
Maybe use "combined" instead?

Jasha10:
> Closes #1376 

I feel like this is one part of closing #1376, but does not close it completely -- our other sweeper plugins will still need modification to support defining sweeps through an input config, right?

website/docs/tutorials/basic/running_your_app/2_multirun.md
jieru-hu: sg! thanks for the edits

hydra/_internal/core_plugins/basic_sweeper.py
jieru-hu: sure - we can just make it default to an empty dict right?

hydra/_internal/core_plugins/basic_sweeper.py
jieru-hu: yes `merged` is probably not the best here - maybe just `final_overrides`? since we are neither merging nor combining the overrides here...

jieru-hu:
> I feel like this is one part of closing #1376, but does not close it completely -- our other sweeper plugins will still need modification to support sweeping via an input config, right?


We need a separate issue tracking HPO config standardization. Other sweepers already support config files :)

Jasha10:
> Other sweepers already support config files :)

Oh, great! Ok, agreed about closing #1376

website/docs/tutorials/basic/running_your_app/2_multirun.md
```diff
@@ -21,6 +25,29 @@ The following sweeps over all combinations of the dbs and schemas.
...
+  sweeper:
+    params:
+      db: mysql,postgresql
+      schema: warehouse,support,school
+```
+
+The syntax are consistent for both input configs and commandline overrides.
+If a sweep is specified in both an input config and at the command line,
+then the commandline sweep will take precedence over the sweep defined 
+in the input config. If we run the same application with the above input config and a new commandline override:
```
Jasha10, line 42: Since the syntax is not yet standardized across all of our supported sweepers, I think this could be confusing to a user who e.g. wants to use the nevergrad sweeper and finds that `hydra.sweeper.params` is not supported.

Jasha10: (Then again, these docs makes sense for the next release assuming that all the sweepers will be standardized)

jieru-hu: I've added an info section to highlight the difference. I will remove it later once the other HPOs are updated.

Jasha10 (approved):
lgtm!

PR #1900, opened 2021-11-23
fix broken website links

antonk52:
## Motivation

Broken links on website that can lead to misleading "Page not found" pages.

Repro steps
- go to [hydra.cc/docs/next/plugins/nevergrad_sweeper/](https://hydra.cc/docs/next/plugins/nevergrad_sweeper/) (note trailing slash, without it the issue does not reproduce)
- click on `this page` link
- see "page not found"

The fix is to force trailing slashes for all pages and to throw during website build to prevent broken links being added in the future.

Yes

## Test Plan

- Navigate to `nevergrad_sweeper` page by pasting URL to the browser bar without a trailing slash
- Expect trailing slash to be added on page load
- Navigating to `this page` link should work


Jasha10 (approved):
Thanks!

Issue #1669, opened 2021-06-11
[Bug] Cannot use `target` kwarg when instantiating with `hydra.utils.instantiate()`


fabiocarrara:
# 🐛 Bug
## Description
`hydra.utils.instantiate()` seems to fail when trying to instantiate a class with a keyword argument called `target`.
It seems like the two keys `_target_` and `target` are clashing, and this raises a 
`TypeError: _call_target() got multiple values for argument 'target'`.

## Checklist
- [x] I checked on the latest version of Hydra
- [x] I created a minimal repro (See [this](https://stackoverflow.com/help/minimal-reproducible-example) for tips).

## To reproduce

**Minimal Code/Config snippet to reproduce**
Available as a Colab Gists:
https://colab.research.google.com/gist/fabiocarrara/6bf5abaecf3f84c1e83239f8af3b8219/hydra_instantiate_bug.ipynb

**Stack trace/error message**
```
TypeError                                 Traceback (most recent call last)
<ipython-input-4-616e52dc8a9b> in <module>()
----> 1 foo = hydra.utils.instantiate(cfg.bar)
      2 foo.target

/usr/local/lib/python3.7/dist-packages/hydra/_internal/instantiate/_instantiate2.py in instantiate(config, *args, **kwargs)
    178         _convert_ = config.pop(_Keys.CONVERT, ConvertMode.NONE)
    179 
--> 180         return instantiate_node(config, *args, recursive=_recursive_, convert=_convert_)
    181     else:
    182         raise InstantiationException(

/usr/local/lib/python3.7/dist-packages/hydra/_internal/instantiate/_instantiate2.py in instantiate_node(node, convert, recursive, *args)
    247                         )
    248                     kwargs[key] = _convert_node(value, convert)
--> 249             return _call_target(target, *args, **kwargs)
    250         else:
    251             # If ALL or PARTIAL non structured, instantiate in dict and resolve interpolations eagerly.

TypeError: _call_target() got multiple values for argument 'target'
```

## Expected Behavior
`target` should be a permitted key name for callable/constructor's kwargs.

## System information
- **Hydra Version** :  1.1
- **Python version** : 3.7.10 (Colab)
- **Virtual environment type and version** : none
- **Operating system** :  Colab

omry:
Thanks for reporting, we will take a look.

Jasha10:
Here is a minimal repro:
```python
# repro.py
from hydra.utils import instantiate
from omegaconf import OmegaConf
[...]

PR #1905, opened 2021-11-30
add _partial_ to instantiate api

### LINKED ISSUE #1283: [Feature Request] Partial instantiate/call
First off, thanks for the wonderful library!!

# 🚀 Feature Request

I would like to be able to partially instantiate/call a class/function. For example:

```yaml
# config.yaml
loss_fn:
  _target_: main.cross_entropy_loss
  _partial_: True
  label_smooth_amount: 0.1
```

```python
# main.py
def cross_entropy_loss(logits, labels, label_smooth_amount):
    ...

loss_fn = hydra.utils.instantiate(cfg.loss_fn)   
# loss_fn = functools.partial(cross_entropy_loss, label_smooth_amount=0.1)
```

## Motivation

Currently, to make the above use-case work, I would do the following:
```yaml
# config.yaml
loss_fn:
  _target_: main.get_cross_entropy_loss
  label_smooth_amount: 0.1
```

```python
# main.py
def get_cross_entropy_loss(label_smooth_amount):
    def cross_entropy_loss(logits, labels):
        ...
    return cross_entropy_loss

loss_fn = hydra.utils.instantiate(cfg.loss_fn)
```

This is acceptable for code I write, but does not work well when I am trying to configure functions that libraries provide. Many libraries follow a more functional style (compared to PyTorch and TensorFlow), so losses/activations/metrics are provided as simple functions as opposed to callable objects. For example, [Flax for JAX](https://flax.readthedocs.io/en/latest/flax.linen.html#activation-functions) (and several other neural network libraries for JAX) defines all its activation functions and pooling layers as straightforward functions instead of classes, making partial instantiation crucial for configuration.

Also, code will often be more clear when there are more simple functions and fewer higher order functions/classes. Partial instantiation will prevent code from having too many of the latter.

## Pitch

Having an optional `_partial_` entry in the config (similar to `_recursive_` and `_convert_`) in my view is the most straightforward way to achieve this. By default this would be `False`, and when `True`, partially instantiates/calls the class/function instead of actually instantiating/calling. 

Another option is to introduce two new methods: `hydra.utils.partial_instantiate` and `hydra.utils.partial_call`. This removes the need for another config entry, and makes it more clear at the call-site what's going on. There is one major disadvantage: it's not clear how this would work with `_recursive_=True`. Would all the recursive instantiations be partial? You probably don't want that. Will only the top level instantiation be partial? This would limit some use cases as well. 

For this reason, I think the `_partial_` entry makes the most sense.

Yes! I was planning to make an `_pop_is_partial` function (like [`pop_convert_mode`](https://github.com/facebookresearch/hydra/blob/40105229ab1eb115e27d5685cd8cddf4647bd0cd/hydra/_internal/utils.py#L652)), then add the appropriate `functools.partial` calls [here](https://github.com/facebookresearch/hydra/blob/master/hydra/utils.py#L100-L111).

jieru-hu:
Closes #1283

add `_partial_` support to the instantiation API


tests/instantiate/__init__.py
```diff
@@ -2,13 +2,45 @@
...
+def _convert_type(obj: Any) -> Any:
+    if isinstance(obj, DictConfig):
+        obj = dict(obj)
+    elif isinstance(obj, ListConfig):
+        obj = list(obj)
+    return obj
+
+
+def partial_equal(obj1: Any, obj2: Any) -> bool:
+    obj1, obj2 = _convert_type(obj1), _convert_type(obj2)
```
Jasha10, line 22: Many places in this file there is the pattern `A == B or partial_equal(A, B)`.
This could be simplified by moving the `A == B` part inside of the `partial_equal` function as below:

```suggestion
def partial_equal(obj1: Any, obj2: Any) -> bool:
    if obj1 == obj2:
        return True

    obj1, obj2 = _convert_type(obj1), _convert_type(obj2)
```

This would enable simplifying the call-sites of `partial_equal`.
For example, the expression below on line 30
```python
            if not (obj1[i] == obj2[i] or partial_equal(obj1[i], obj2[i])):
                return False
```
could be replaced with this:
```python
            if not partial_equal(obj1[i], obj2[i]):
                return False
```

Jasha10: I've tested out the change locally to make sure it works. If you agree with the idea, I can push a commit to this PR branch!

hydra/_internal/instantiate/_instantiate2.py
```diff
@@ -45,7 +47,7 @@ def _extract_pos_args(*input_args: Any, **kwargs: Any) -> Tuple[Any, Any]:
     return output_args, kwargs
 
 
-def _call_target(_target_: Callable, *args, **kwargs) -> Any:  # type: ignore
+def _call_target(_target_: Callable, partial: Optional[bool], *args, **kwargs) -> Any:  # type: ignore
```
Jasha10, line 50: Could we get away with `bool` instead of `Optional[bool]`?
(Same for the signature of the `instantiate_node` function below)

tests/instantiate/__init__.py
jieru-hu: nice!! feel free to push, thanks jasha!

hydra/_internal/instantiate/_instantiate2.py
jieru-hu: the  reason it is `None` is a bit nuanced. with configs like `_recursive_`,`_convert_`,   a child node  will always inherit the  parent node config if the config is not specifically set in the child node. But  `_partial_` is different, i think it only makes sense to  set `_partial_` per node, which means if `_partial_` is not set in  a node,  it should default to False instead of inherit that from the parent.

tests/instantiate/__init__.py
Jasha10: Done in https://github.com/facebookresearch/hydra/pull/1905/commits/bf689d30fd92d6f1fa46179202c9205df99e5a04

hydra/_internal/instantiate/_instantiate2.py
Jasha10: I agree that `_partial_` should be set per-node...
 why not make the default `partial=False` instead of `partial=None`?

jieru-hu: you are right, Pr updated! :)

tests/instantiate/__init__.py
```diff
@@ -212,9 +253,9 @@ def __init__(self, value: Any, left: Any = None, right: Any = None) -> None:
     def __eq__(self, other: Any) -> Any:
         if isinstance(other, type(self)):
             return (
-                self.value == other.value
-                and self.left == other.left
-                and self.right == other.right
+                (partial_equal(self.value, other.value))
+                and (partial_equal(self.left, other.left))
+                and (partial_equal(self.right, other.right))
```
Jasha10, line 258:
```suggestion
                partial_equal(self.value, other.value)
                and partial_equal(self.left, other.left)
                and partial_equal(self.right, other.right)
```

tests/instantiate/__init__.py
```diff
@@ -236,7 +277,9 @@ def __init__(
 
     def __eq__(self, other: Any) -> Any:
         if isinstance(other, type(self)):
-            return self.dictionary == other.dictionary and self.value == other.value
+            return (partial_equal(self.dictionary, other.dictionary)) and (
+                partial_equal(self.value, other.value)
+            )
```
Jasha10, line 282:
```suggestion
            return partial_equal(self.dictionary, other.dictionary) and partial_equal(
                self.value, other.value
            )
```

Jasha10 (commented):
Removing straggler parentheses from my previous commit:

tests/instantiate/__init__.py
Jasha10: These diffs should be compliant with the `black` formatter.

hydra/_internal/instantiate/_instantiate2.py
```diff
@@ -235,18 +251,21 @@ def instantiate_node(
...
             _target_ = _resolve_target(node.get(_Keys.TARGET))
             kwargs = {}
             for key, value in node.items():
                 if key not in exclude_keys:
                     if recursive:
                         value = instantiate_node(
-                            value, convert=convert, recursive=recursive
+                            value,
+                            convert=convert,
+                            recursive=recursive,
```
Jasha10, line 264:
```suggestion
                            value, convert=convert, recursive=recursive
```

hydra/_internal/instantiate/_instantiate2.py
```diff
@@ -256,15 +275,19 @@ def instantiate_node(
                 for key, value in node.items():
                     # list items inherits recursive flag from the containing dict.
                     dict_items[key] = instantiate_node(
-                        value, convert=convert, recursive=recursive
+                        value,
+                        convert=convert,
+                        recursive=recursive,
```
Jasha10, line 280:
```suggestion
                        value, convert=convert, recursive=recursive
```

hydra/_internal/instantiate/_instantiate2.py
```diff
@@ -256,15 +275,19 @@ def instantiate_node(
...
                 return dict_items
             else:
                 # Otherwise use DictConfig and resolve interpolations lazily.
                 cfg = OmegaConf.create({}, flags={"allow_objects": True})
                 for key, value in node.items():
                     cfg[key] = instantiate_node(
-                        value, convert=convert, recursive=recursive
+                        value,
+                        convert=convert,
+                        recursive=recursive,
```
Jasha10, line 290:
```suggestion
                        value, convert=convert, recursive=recursive
```

hydra/_internal/instantiate/_instantiate2.py
```diff
@@ -214,14 +222,22 @@ def instantiate_node(
...
+
     # If OmegaConf list, create new list of instances if recursive
     if OmegaConf.is_list(node):
         items = [
-            instantiate_node(item, convert=convert, recursive=recursive)
+            instantiate_node(
+                item,
+                convert=convert,
+                recursive=recursive,
+            )
```
Jasha10, line 240:
```suggestion
            instantiate_node(item, convert=convert, recursive=recursive)
```

Jasha10 (commented):
A few more nits:

Jasha10 (changes requested):
This PR suffers from the same bug as https://github.com/facebookresearch/hydra/issues/1669:
```python
>>> instantiate({"_target_": "builtins.dict", "_partial_": True})
functools.partial(<class 'dict'>)
>>> instantiate({"_target_": "builtins.dict", "_partial_": True, "partial": "unrelated"})
...
TypeError: _call_target() got multiple values for argument 'partial'
```

Jasha10 (approved):

Jasha10:
@jieru-hu I think we need a news fragment for this.

PR #1915, opened 2021-12-03
Improve instantiate handling of callable _target_ kwarg

Jasha10:
This PR closes #1914.

Commits
- https://github.com/facebookresearch/hydra/commit/90468cbbaf7963f6cd89658f5e7c73b8c2c54073: add failing test

  As of this commit, the following fails:
  ```
  pytest tests/instantiate -k 'test_instantiate_with_callable_target_keyword[instantiate2-nested]'
  ```
- https://github.com/facebookresearch/hydra/commit/1122135b11dbcf9e02c2d8ffc6c0785512898417: patch `_convert_target_to_string`

PR #1916, opened 2021-12-05
Improve `instantiate` handling of exceptions raised by user code

### LINKED ISSUE #1911: Instantiation exception also raises an exception sometimes
# 🐛 Bug
## Description

When there's an instantiation exception, raising a chained exception sometimes also fails because it doesn't take an argument:

https://github.com/facebookresearch/hydra/blob/9759f87673be1d41fe33279e1b7980f8efb38b19/hydra/_internal/instantiate/_instantiate2.py#L63-L66

## Checklist

## To reproduce

Try to instantiate an object such that it raises an exception that doesn't take any argument.

## Expected Behavior

Maybe it should try to show another type of exception instead? Or chaining (`from e`).

## System information
- **Hydra Version** :  1.1
- **Python version** : 3.8
- **Virtual environment type and version** : conda 4.11
- **Operating system** : Ubuntu 18.04

Jasha10:
This PR fixes a bug that arises when `instantiate` is invoked on a callable that raises a custom exception that has a non-standard signature.

## Background info about the bug:
This PR closes #1911.

A call to `instantiate({"_target_": target_callable, **kwargs})` works by locating the `target_callable` and calling it on the given arguments `**kwargs`.
The call to `target_callable` is performed inside of the [`hydra._internal.instantiate._instantiate2._call_target`](https://github.com/facebookresearch/hydra/blob/86b544bf1b635058094ff29690d85b5091147312/hydra/_internal/instantiate/_instantiate2.py#L62) function:
```python
def _call_target(_target_: Callable, *args, **kwargs) -> Any:  # type: ignore
    """Call target (type) with args and kwargs."""
    try:
        ...
        return _target_(*args, **kwargs)
    except Exception as e:
        raise type(e)(
            f"Error instantiating '{_convert_target_to_string(_target_)}' : {e}"
```
This `_call_target` function has a `try`/`except` block that catches exceptions that might be raisd by the `target_callable`.
In the `_call_target` function's `except` block, another exception is raised with an error message giving context about the exception. The exception raised in the `except` block is of the same type as the exception that was originally caught.
This is problematic because we must assume that the exception class `type(e)` can be called with one argument; we are making the assumption here that `type(e).__init__` accepts a string error message as it's first and only argument.
This assumption fails if the user's code raises a custom exception whose `__init__` function takes a non-standard number of arguments (as was the case in bug report #1911).

## Proposed Changes
This PR fixes the reported bug by not raising an exception of `type(e)` if something goes wrong with the call to `_target_`.
Instead, an instance of `hydra.errors.InstantiationException` is raised *from* `e` (a lá [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining)).

Since this PR changes the type of exception that will be raised if user code throws an exception, it should be considered an API change.

Commits:
  - https://github.com/facebookresearch/hydra/commit/2505edae29bff0b4e9252d5f5ef54d66ff5c658d: Add failing tests
  - https://github.com/facebookresearch/hydra/commit/a2dbbf87a876db74961e7f1edca8652992a91827: raise InstantiationException instead of type(e)
  - https://github.com/facebookresearch/hydra/commit/247a3ed7051a0ed71ada0843df07c1ff51db15bf: add news fragment
  - https://github.com/facebookresearch/hydra/pull/1916/commits/95494eeb09df52ad7a18d19bafabeb6132b646e7: update test regex patterns for python3.6 compatibility.

I'm not planning to merge this PR until after #1905 and #1915 are finished, as those PRs also touch code related to `instantiate`.

### DIFF
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -60,13 +60,19 @@ def _call_target(_target_: Callable, _partial_: bool, *args, **kwargs) -> Any:  
         for v in kwargs.values():
             if OmegaConf.is_config(v):
                 v._set_parent(None)
-        if _partial_:
-            return functools.partial(_target_, *args, **kwargs)
-        return _target_(*args, **kwargs)
     except Exception as e:
         raise type(e)(
             f"Error instantiating '{_convert_target_to_string(_target_)}' : {e}"
         ).with_traceback(sys.exc_info()[2])
+
+    try:
+        if _partial_:
+            return functools.partial(_target_, *args, **kwargs)
+        return _target_(*args, **kwargs)
+    except Exception as e:
+        raise InstantiationException(
+            f"Error instantiating '{_convert_target_to_string(_target_)}' : {repr(e)}"
+        ) from e
 
 
 def _convert_target_to_string(t: Any) -> Any:
--- a/news/1911.api_change
+++ b/news/1911.api_change
@@ -1,0 +1,1 @@
+If user code raises an exception when called by `instantiate`, raise an `InstantiateError` exception instead of an instance of the same exception class that was raised by the user code.
--- a/tests/instantiate/__init__.py
+++ b/tests/instantiate/__init__.py
@@ -3,7 +3,7 @@ import collections
 import collections.abc
 from dataclasses import dataclass
 from functools import partial
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, NoReturn, Optional, Tuple
 
 from omegaconf import MISSING, DictConfig, ListConfig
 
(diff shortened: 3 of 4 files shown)


rsokl:
This will be very useful for improving compatibility with Hydra + pydantic! Currently, this [pydantic-based example](https://mit-ll-responsible-ai.github.io/hydra-zen/explanation/type_refinement.html#using-third-party-runtime-type-checkers) hits this exact issue, because the validation exception class that pydantic raises does not use a standard `__init__`.

Jasha10:
Nice!

jieru-hu (approved):
lgtm! :)

Jasha10:
Rebased against `main` to resolve merge conflict.

PR #1929, opened 2021-12-18
allow NoneType values in optuna CategorialDistribution

Jasha10:
`optuna.distributions.CategoricalDistribution` allows its `choices` parameter to include `None` values.
(see the note near the top of the [Optuna CategoricalDistribution docs page](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.distributions.CategoricalDistribution.html?highlight=CategoricalDistribution#optuna.distributions.CategoricalDistribution))
Previously, Hydra's optuna plugin disallowed `Non` values for `CategoricalDistribution` sweeps. This PR enables `None` values.


jieru-hu:
any context on why this was not supported in the first place? 🤔

Jasha10:
My assumption is that it was an oversight.

pixelb (approved):

YannDubs:
See #1986, I think that the optuna sweeper does not correctly always deal with `NoneType`.

PR #1952, opened 2022-01-11
remove handling of deprecated "old optional" defaults list syntax

Jasha10:
In previous versions of hydra, the following two defaults lists were equivalent:
```yaml
defaults:
  - optional foo: bar
```
```yaml
defaults:
  - foo: bar
    optional: true
```
This PR removes support for the ~former~ latter syntax.

### DIFF
--- a/hydra/_internal/config_repository.py
+++ b/hydra/_internal/config_repository.py
@@ -15,6 +15,7 @@ from omegaconf import (
     read_write,
 )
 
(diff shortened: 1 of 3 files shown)


Jasha10:
I'm marking this PR as a draft until the new Hydra base_version functionality is in place.

pixelb (approved):
looks good. thanks

Issue #1786, opened 2021-08-22
Cannot  access HydraConfig in sweep using Hydra resolver


cnut1648:
# 🚀 Feature Request
Hi I am wondering if it is possible to use `hydra:` namespace in the sweep.

## Motivation
I would like to customize the sweep dir, but however it seems that when I need to do a value interpolation I got the following error message
```
omegaconf.errors.InterpolationResolutionError: ValueError raised while resolving interpolation: HydraConfig was not set
    full_key: hydra.sweep.dir
    reference_type=SweepDir
    object_type=SweepDir
```
My config is (created from [this example](https://github.com/facebookresearch/hydra/blob/master/examples/configure_hydra/workdir/conf/config.yaml))
```yaml
# @package _global_
hydra:
  run:
    dir: run_dir
  sweep:
    dir: ${log_dir}
    subdir: ${hydra.job.num}

a: a1
work_dir: ${hydra:runtime.cwd}
log_dir: "${work_dir}/../my_long_logs"
```
It seems that the `hydra:` namespace can't be accessed during the sweep. and my current workaround is making `log_dir` to be
```
log_dir: "../my_long_logs"
```
## Additional context
The python version is python 3.6. and the hydra version is 1.1.1 (the latest)

jieru-hu:
Hi @cnut1648  could you share how are you running the sweep? We do not support sweeping over Hydra configs.

cnut1648:
Hi @jieru-hu in this case I just use something like
```
python main.py -m a=1,2,3
```
The actual command I think does not matter here since I only would like to make `hydra.sweep.dir` interpolate with `work_dir`. So the sweep dir would be created to `<work_dir>/../my_long_logs` instead of the default dir.

jieru-hu:
hey @cnut1648 
sorry for getting back to you late - as you can see from the stacktrace, HydraConfig is not initialized at that time so OmegaConf could not run the `hydra` custom resolver. if you update your config to be `work_dir: ${hydra.runtime.cwd}` it should work as expected.

cnut1648:
Thank you @jieru-hu! I think this works and I can create an another field that is interpolated with the `hydra.runtime.cwd`. But I have yet another question, now I changed my config to be

```yaml
work_dir: ${hydra.runtime.cwd}
log_dir: "${work_dir}/../my_long_logs"
```
I noticed that I cannot access either `cfg.work_dir` or `cfg.log_dir` in the runtime. I got the following error

```
[...]

PR #1979, opened 2022-01-20
hydra resolver in output dir config

jieru-hu:
ran into this when i was investigating #1786 
an regression introduced by https://github.com/facebookresearch/hydra/pull/1867 :)

### DIFF
--- a/hydra/core/utils.py
+++ b/hydra/core/utils.py
@@ -109,6 +109,9 @@ def run_job(
     old_cwd = os.getcwd()
     orig_hydra_cfg = HydraConfig.instance().cfg
 
+    # init Hydra config for config evaluation
+    HydraConfig.instance().set_config(config)
+
     output_dir = str(OmegaConf.select(config, job_dir_key))
     if job_subdir_key is not None:
         # evaluate job_subdir_key lazily.
@@ -121,6 +124,7 @@ def run_job(
         with open_dict(config.hydra.runtime):
             config.hydra.runtime.output_dir = os.path.abspath(output_dir)
 
+    # update Hydra config
     HydraConfig.instance().set_config(config)
     _chdir = None
     try:
--- a/tests/test_hydra.py
+++ b/tests/test_hydra.py
@@ -1594,13 +1594,13 @@ def test_hydra_verbose_1897(tmpdir: Path, multirun: bool) -> None:
 def test_hydra_resolver_in_output_dir(tmpdir: Path, multirun: bool) -> None:
     from hydra import __version__
 
-    output_dir = str(tmpdir) + "${val}"
+    output_dir = str(Path(tmpdir) / "${hydra:runtime.version}")
 
     cmd = [
         "tests/test_apps/hydra_resolver_in_output_dir/my_app.py",
         f"hydra.run.dir='{output_dir}'",
-        f"hydra.sweep.subdir='{output_dir}'",
-        "hydra.sweep.dir=.",
+        "hydra.sweep.subdir='${hydra:runtime.version}'",
+        f"hydra.sweep.dir={str(Path(tmpdir))}",
         "hydra.job.chdir=False",
     ]
 
@@ -1609,6 +1609,7 @@ def test_hydra_resolver_in_output_dir(tmpdir: Path, multirun: bool) -> None:
 
     out, _ = run_python_script(cmd)
 
-    expected_output_dir = os.path.join(str(tmpdir) + f"{__version__}")
-    assert Path(expected_output_dir, "my_app.log").exists()
+    expected_output_dir = str(Path(tmpdir) / f"{__version__}")
+    expected_log_file = Path(expected_output_dir) / "my_app.log"
+    assert expected_log_file.exists()
     assert expected_output_dir in out
--- a/tests/test_hydra.py
+++ b/tests/test_hydra.py
@@ -1585,3 +1585,30 @@ def test_hydra_verbose_1897(tmpdir: Path, multirun: bool) -> None:
     if multirun:
         cmd += ["+a=1,2", "-m"]
     run_python_script(cmd)
+
+
+@mark.parametrize(
+    "multirun",
+    [False, True],
+)
+def test_hydra_resolver_in_output_dir(tmpdir: Path, multirun: bool) -> None:
+    from hydra import __version__
+
+    output_dir = str(tmpdir) + "${val}"
+
+    cmd = [
+        "tests/test_apps/hydra_resolver_in_output_dir/my_app.py",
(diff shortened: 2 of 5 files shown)


hydra/core/utils.py
```diff
@@ -131,6 +134,7 @@ def run_job(
         with open_dict(config.hydra.runtime):
             config.hydra.runtime.output_dir = os.path.abspath(output_dir)
 
+    # update Hydra config
     HydraConfig.instance().set_config(config)
```
Jasha10, line 138: I think this one can be deleted.

jieru-hu: This is still needed since we need  to update the  value of `hydra.runtime.output_dir` - if not updated an example Hydra app like the following won't work as  expected
```python3
@hydra.main(config_path=".", config_name="config")
def my_app(_: DictConfig) -> None:
    print(HydraConfig.instance().get().runtime.output_dir)

```

Jasha10: I see! Currently, the core tests are still passing even if the second call to `set_config` is deleted. Could you please add the statement `print(HydraConfig.instance().get().runtime.output_dir)` to the new `my_app.py` test file and check the output?

jieru-hu: makes sense, added! :) Thanks

tests/test_hydra.py
```diff
@@ -1585,3 +1585,30 @@ def test_hydra_verbose_1897(tmpdir: Path, multirun: bool) -> None:
...
+    ]
+
+    if multirun:
+        cmd += ["-m"]
+
+    out, _ = run_python_script(cmd)
+
+    output_dir = os.path.join(str(tmpdir) + f"{__version__}")
+    assert Path(output_dir, "my_app.log").exists()
+    assert output_dir in out
```
Jasha10, line 1614: This diff avoids using the same name "`output_dir`" for different purposes:

```suggestion
    expected_output_dir = os.path.join(str(tmpdir) + f"{__version__}")
    assert Path(expected_output_dir, "my_app.log").exists()
    assert expected_output_dir in out
```

hydra/core/utils.py
Jasha10: It feels to me like a hack to call `HydraConfig.instance().set_config` twice.
The root issue in my mind is that `HydraConfig.instance().set_config` is creating a copy that is different from the original object (which is `config.hydra`). I think there should just be one copy of the hydra config.

I have a way to make these two copies share memory, so that either one can be updated without needing to call `HydraConfig.set_config` a second time.
Here is the diff comparing with the latest commit (https://github.com/facebookresearch/hydra/pull/1979/commits/04d7876eb3241cfa352c25e59af96b388d3f67ff) on this PR branch:
```diff
diff --git a/hydra/core/hydra_config.py b/hydra/core/hydra_config.py
index 6ebcca763f..6ba721cf5b 100644
--- a/hydra/core/hydra_config.py
+++ b/hydra/core/hydra_config.py
@@ -1,7 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 from typing import Any, Optional

-from omegaconf import DictConfig, OmegaConf
+from omegaconf import DictConfig, OmegaConf, read_write

 from hydra.conf import HydraConf
 from hydra.core.singleton import Singleton
@@ -23,6 +23,10 @@ class HydraConfig(metaclass=Singleton):
         # into the user's config.
         self.cfg = OmegaConf.masked_copy(cfg, "hydra")  # type: ignore
         self.cfg.hydra._set_parent(cfg)  # type: ignore
+        # Ensure cfg.hydra and self.cfg.hydra refer to the same object:
+        with read_write(cfg):
+            with open_dict(cfg):
+                cfg.hydra = self.get()

     @staticmethod
     def get() -> HydraConf:
diff --git a/hydra/core/utils.py b/hydra/core/utils.py
index 602f7f3a91..e993294521 100644
--- a/hydra/core/utils.py
+++ b/hydra/core/utils.py
@@ -134,8 +134,6 @@ def run_job(
         with open_dict(config.hydra.runtime):
             config.hydra.runtime.output_dir = os.path.abspath(output_dir)

-    # update Hydra config
-    HydraConfig.instance().set_config(config)
     _chdir = None
     try:
         ret = JobReturn()
```
@jieru-hu, what do you think about this idea?
I ran the tests locally and they seem to be passing.

Jasha10: Resolved via PMs.

jieru-hu: resolving per offline discussions cc @Jasha10

tests/test_hydra.py
```diff
@@ -1585,3 +1585,30 @@ def test_hydra_verbose_1897(tmpdir: Path, multirun: bool) -> None:
...
+        "hydra.sweep.dir=.",
+        "hydra.job.chdir=False",
+    ]
+
+    if multirun:
+        cmd += ["-m"]
+
+    out, _ = run_python_script(cmd)
+
+    expected_output_dir = os.path.join(str(tmpdir) + f"{__version__}")
```
Jasha10, line 1612:
```suggestion
    output_dir = str(tmpdir) + "/${val}"

    cmd = [
        "tests/test_apps/hydra_resolver_in_output_dir/my_app.py",
        f"hydra.run.dir='{output_dir}'",
        f"hydra.sweep.subdir='{output_dir}'",
        "hydra.sweep.dir=.",
        "hydra.job.chdir=False",
    ]

    if multirun:
        cmd += ["-m"]

    out, _ = run_python_script(cmd)

    expected_output_dir = os.path.join(str(tmpdir), f"{__version__}")
```

You need a comma to create a subdirectory.
```python
>>> os.path.join("tmpdir" + "version")
'tmpdirversion'
>>> os.path.join("tmpdir", "version")
'tmpdir/version'
```

Jasha10 (changes requested):

tests/test_hydra.py
```diff
@@ -1585,3 +1585,30 @@ def test_hydra_verbose_1897(tmpdir: Path, multirun: bool) -> None:
...
+
+
+@mark.parametrize(
+    "multirun",
+    [False, True],
+)
+def test_hydra_resolver_in_output_dir(tmpdir: Path, multirun: bool) -> None:
+    from hydra import __version__
+
+    output_dir = str(tmpdir) + "${val}"
```
Jasha10, line 1597: What do you think about replacing `${val}` with `${hydra:runtime.version}` here? We could then get rid of the config file (to make the test simpler and easier to understand for future maintainers).

tests/test_hydra.py
jieru-hu: thanks for catching the typo. fixed.

Jasha10 (approved):
Looks good!

PR #2005, opened 2022-02-03
bump black version to >=22.1.0

Jasha10:
Version 22.1.0 of the black formatter has [recently been released](https://github.com/psf/black/releases/tag/22.1.0).
According to the release notes, this is the first non-beta release of black, and it
is the first black release subject to the new [stability policy](https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy).

This PR updates our pin on black to `>=22.1.0`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
         files: 'tools/.*'
 
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.1.0
     hooks:
       - id: black
         language_version: python3.6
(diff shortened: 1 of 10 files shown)


jieru-hu:
probably need to update this as well https://github.com/facebookresearch/hydra/blob/main/.pre-commit-config.yaml#L13
and pin the version?

Jasha10:
Good catch regarding the pre-commit config!

As for the file `requirements/dev.txt`, I chose to use inequality `>=22.1.0` instead of a hard pin `==22.1.0` because that way we can get bugfixes and such when new versions of black are released. I figured that using a soft pin should not break our CI more than once per year -- according to black's new stability policy, linked above, they will only make formatting updates once per calendar year.

pixelb (approved):

PR #2011, opened 2022-02-03
Exception formatting: handle case where `module is None`

### LINKED ISSUE #2010: [Bug] AssertionError in multithread context
I ran into an AssertionError using Hydra with python's `threading` library.

In theory, Hydra is not thread safe. However, we should be able to support using Hydra in a multithread application if Hydra is only used from one thread.

Here is the relevant part of the traceback:
```
An error occurred during Hydra's exception formatting:
AssertionError()
  File "/home/jasha10/miniconda3/envs/pysc/lib/python3.9/site-packages/hydra/_internal/utils.py", line 252, in run_and_report
    assert mdl is not None
AssertionError
```

### LINKED ISSUE #2342: [Bug] Hydra's exception formatting cannot handle exception raised in C extension
# 🐛 Bug
## Description
I use `hydra` and `av`(PyAV), when `av` raise an error inside my hydra app, hydra cannot handle it correctly.
Because `inspect.getmodule` won't work for C extension.
```python
end: Optional[TracebackType] = tb
num_frames = 0
while end is not None:
    frame = end.tb_frame
    mdl = inspect.getmodule(frame)
    assert mdl is not None
    name = mdl.__name__
    if name.startswith("omegaconf."):
        break
    end = end.tb_next
    num_frames = num_frames + 1
```

## Checklist

## To reproduce

**Minimal Code/Config snippet to reproduce**
```python
# foo.pyx
def foo():
    raise Exception

# main.py
import pyximport
pyximport.install() # to directly import pyx without `cythonize -i foo.pyx`

import hydra
import foo
@hydra.main(version_base='1.2')
def app(_):
    foo.foo()
app()
```

**Stack trace/error message**
```
Error executing job with overrides: []
An error occurred during Hydra's exception formatting:
AssertionError()
  File "/home/youjiacheng/miniconda3/envs/env/lib/python3.10/site-packages/hydra/_internal/utils.py", line 254, in run_and_report
    assert mdl is not None
AssertionError

  File "/home/youjiacheng/test/main.py", line 6, in <module>
    app()
  File "/home/youjiacheng/miniconda3/envs/env/lib/python3.10/site-packages/hydra/main.py", line 90, in decorated_main
    _run_hydra(
  File "/home/youjiacheng/miniconda3/envs/env/lib/python3.10/site-packages/hydra/_internal/utils.py", line 389, in _run_hydra
    _run_app(
  File "/home/youjiacheng/miniconda3/envs/env/lib/python3.10/site-packages/hydra/_internal/utils.py", line 452, in _run_app
    run_and_report(
  File "/home/youjiacheng/miniconda3/envs/env/lib/python3.10/site-packages/hydra/_internal/utils.py", line 296, in run_and_report
    raise ex
  File "/home/youjiacheng/miniconda3/envs/env/lib/python3.10/site-packages/hydra/_internal/utils.py", line 213, in run_and_report
    return func()
  File "/home/youjiacheng/miniconda3/envs/env/lib/python3.10/site-packages/hydra/_internal/utils.py", line 453, in <lambda>
    lambda: hydra.run(
  File "/home/youjiacheng/miniconda3/envs/env/lib/python3.10/site-packages/hydra/_internal/hydra.py", line 132, in run
    _ = ret.return_value
  File "/home/youjiacheng/miniconda3/envs/env/lib/python3.10/site-packages/hydra/core/utils.py", line 260, in return_value
    raise self._return_value
  File "/home/youjiacheng/miniconda3/envs/env/lib/python3.10/site-packages/hydra/core/utils.py", line 186, in run_job
  File "/home/youjiacheng/test/main.py", line 5, in app
    foo.foo()
  File "foo.pyx", line 2, in foo.foo
    raise Exception
Exception
```

## Expected Behavior

Following traceback is produced by renaming `foo.pyx` to `foo.py`. 
```
Error executing job with overrides: []
  File "/home/youjiacheng/test/main.py", line 5, in app
    foo.foo()
  File "/home/youjiacheng/test/foo.py", line 2, in foo
    raise Exception
Exception

```

```python
end: Optional[TracebackType] = tb
num_frames = 0
while end is not None:
    frame = end.tb_frame
    mdl = inspect.getmodule(frame)
    if mdl is not None:
        name = mdl.__name__
        if name.startswith("omegaconf."):
            break
    end = end.tb_next
    num_frames = num_frames + 1
```

## System information
- **Hydra Version** :  1.2.0
- **Python version** : 3.10.4
- **Virtual environment type and version** : conda 4.13.0
- **Operating system** : CentOS Linux release 7.9.2009 (Core)

Jasha10:
Fix an issue where Hydra's exception-handling logic could raise an `AssertionError` if the module from which in the exception's originates cannot be determined.

Closes #2010. Closes #2342.
See #2342 for a way to reproduce the bug.

Details:

Hydra's exception-handling logic automatically removes omegaconf-related stack frames before printing the handled traceback. This requires inspecting the stack frames (when an exception arises) to determine which stack frames belong to OmegaConf. Hydra uses the function `inspect.getmodule` to determine the module to which a given stack frame belongs.

In the case where the module from which a stack frame originated cannot be determined, `inspect.getmodule` returns `None`. Previously, this would cause Hydra to throw an `AssertionError`. This PR adds logic to handle the case where the module cannot be determined, avoiding the `AssertionError`.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -4,7 +4,7 @@ import os
 import re
 from pathlib import Path
 from textwrap import dedent
-from typing import Any, Optional
+from typing import Any, NoReturn, Optional
 from unittest.mock import patch
 
 from omegaconf import DictConfig, OmegaConf
@@ -216,25 +216,58 @@ class TestRunAndReport:
         stderr_output = mock_stderr.read()
         assert_regex_match(expected_traceback_regex, stderr_output)
 
+    def test_simplified_traceback_with_no_module(self) -> None:
+        """
+        Test that simplified traceback logic can succeed even if
+        `inspect.getmodule(frame)` returns `None` for one of
+        the frames in the stacktrace.
+        """
+        demo_func = self.DemoFunctions.run_job_wrapper
+        expected_traceback_regex = dedent(
+            r"""
+            Traceback \(most recent call last\):$
+              File "[^"]+", line \d+, in nested_error$
+                assert False, "nested_err"$
+            AssertionError: nested_err$
+            assert False$
+            Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace\.$
+            """
+        )
+        mock_stderr = io.StringIO()
+        with raises(SystemExit, match="1"), patch("sys.stderr", new=mock_stderr):
+            # Patch `inspect.getmodule` so that it will return None. This simulates a
+            # situation where a python module cannot be identified from a traceback
+            # stack frame. This can occur when python extension modules or
+            # multithreading are involved.
+            with patch("inspect.getmodule", new=lambda *args: None):
+                run_and_report(demo_func)
+        mock_stderr.seek(0)
+        stderr_output = mock_stderr.read()
+        assert_regex_match(expected_traceback_regex, stderr_output)
+
     def test_simplified_traceback_failure(self) -> None:
         """
         Test that a warning is printed and the original exception is re-raised
         when an exception occurs during the simplified traceback logic.
         """
         demo_func = self.DemoFunctions.run_job_wrapper
+
+        def throws(*args: Any, **kwargs: Any) -> NoReturn:
+            assert False, "Error thrown"
+
         expected_traceback_regex = dedent(
             r"""
             An error occurred during Hydra's exception formatting:$
-            AssertionError\(.*\)$
+            AssertionError\(.*Error thrown.*\)$
             """
         )
         mock_stderr = io.StringIO()
         with raises(AssertionError, match="nested_err"), patch(
             "sys.stderr", new=mock_stderr
         ):
-            # patch `inspect.getmodule` so that an exception will occur in the
-            # simplified traceback logic:
-            with patch("inspect.getmodule", new=lambda *args: None):
+            # patch `traceback.print_exception` so that an exception will occur
+            # in the simplified traceback logic:
+            with patch("traceback.print_exception", new=throws):
                 run_and_report(demo_func)
         mock_stderr.seek(0)
         stderr_output = mock_stderr.read()
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -4,10 +4,10 @@ import inspect
 import logging.config
 import os
 import sys
+import traceback
 import warnings
 from dataclasses import dataclass
 from os.path import dirname, join, normpath, realpath
-from traceback import print_exc, print_exception
 from types import FrameType, TracebackType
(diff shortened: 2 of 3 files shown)


jieru-hu:
could you paste an example before and after? :)

Jasha10:
Unfortunately I forgot how to reproduce this issue.

Jasha10:
I'm closing this PR for now as I'm no longer able to reproduce the original issue.

tests/test_utils.py
```diff
@@ -216,25 +216,56 @@ def test_failure(self, demo_func: Any, expected_traceback_regex: str) -> None:
...
+                assert False, "nested_err"$
+            AssertionError: nested_err$
+            assert False$
+            Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace\.$
+            """
+        )
+        mock_stderr = io.StringIO()
+        with raises(SystemExit, match="1"), patch("sys.stderr", new=mock_stderr):
+            # patch `inspect.getmodule` so that an exception will occur in the
+            # simplified traceback logic:
```
YouJiacheng, line 239: This comment seems copied from original test, and should be removed here.

Jasha10:
I'm reopening this PR per discussion in #2343.

tests/test_utils.py
YouJiacheng: But this PR make the exception not occur in the simplified traceback logic, so I think the comment should be modified.
E.g. "patch `inspect.getmodule` so that it return None in the simplified traceback logic, but there should not be an exception."

tests/test_utils.py
```diff
@@ -216,25 +216,56 @@ def test_failure(self, demo_func: Any, expected_traceback_regex: str) -> None:
...
+        assert_regex_match(expected_traceback_regex, stderr_output)
+
     def test_simplified_traceback_failure(self) -> None:
         """
         Test that a warning is printed and the original exception is re-raised
         when an exception occurs during the simplified traceback logic.
         """
         demo_func = self.DemoFunctions.run_job_wrapper
+
+        def throws(*args: Any, **kwargs: Any) -> None:
```
YouJiacheng, line 253: `NoReturn` is more suitable than `None`, per https://docs.python.org/3/library/typing.html#typing.NoReturn and https://mypy.readthedocs.io/en/stable/more_types.html#the-noreturn-type.
Not important, just a minor suggestion.

tests/test_utils.py
Jasha10: You're right. Thanks.

PR #2012, opened 2022-02-03
noxfile: call _upgrade_basic from every session

Jasha10:
This PR improves our noxfile, insuring that the `_upgrade_basic` function
(which installs/upgrades pip+setuptools) is called before every nox session.

This diff is based on offline discussion with @jieru-hu regarding
the CI failure that occurred in PR #2009.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -184,7 +184,6 @@ def select_plugins(session, directory: str) -> List[Plugin]:
 
 
 def install_dev_deps(session):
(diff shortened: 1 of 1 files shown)


Jasha10:
With this diff, `_upgrade_basic` gets called exactly once by each function decorated with `@nox.session`.

jieru-hu (approved):
🚀

PR #2014, opened 2022-02-04
allow list as toplevel instantiate input

### LINKED ISSUE #1950: [Feature Request] `instantiate` support for ListConfig at top level
The below text is copied directly from a post by a user of the internal Meta Hydra users group:

---

It seems that Hydra doesn't support recursive instantiation for list objects yet. For example, I would expect the following code to work, but it's raising an InstantiationException: Top level config has to be OmegaConf DictConfig, plain dict, or a Structured Config class or instance.
```python
hydra.utils.instantiate([
    {"_target_": "torch.nn.Linear", "in_features": 3, "out_features": 4},
    {"_target_": "torch.nn.Linear", "in_features": 4, "out_features": 5},
])
# hydra.errors.InstantiationException: Top level config has to be OmegaConf DictConfig, plain dict, or a Structured Config class or instance
```
As another example, we often instantiate a list of callbacks and pass them to a trainer. Although the same thing can be done with a list comprehension, you can probably imagine its usefulness when composed with other structures. This looks like a simple and intuitive feature, so I'm curious if we plan to support it, if it hasn't been done already.

Jasha10:
This PR enables the [`instantiate`](https://hydra.cc/docs/next/advanced/instantiate_objects/overview/) function to take a list or ListConfig as top-level input. Closes #1950.

Consider the following example:
```python
# demo.py

lst = [
    {"_target_": "pathlib.Path", "_args_": ["foo"]},
    {"_target_": "pathlib.Path", "_args_": ["bar"]},
]

paths = instantiate(lst)
print(paths)
```
### Before this diff:
```python
$ python demo.py
  File "/home/jasha10/hydra.git/demo.py", line 9, in <module>
    paths = instantiate(lst)
  File "/home/jasha10/hydra.git/hydra/_internal/instantiate/_instantiate2.py", line 193, in instantiate
    raise InstantiationException(
hydra.errors.InstantiationException: Top level config has to be OmegaConf DictConfig, plain dict, or a Structured Config class or instance
```
### After this diff:
```python
$ python demo.py
[PosixPath('foo'), PosixPath('bar')]
```

### DIFF
--- a/tests/instantiate/test_instantiate.py
+++ b/tests/instantiate/test_instantiate.py
@@ -1,5 +1,4 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
-
 import copy
 import pickle
 import re
--- a/tests/instantiate/test_instantiate.py
+++ b/tests/instantiate/test_instantiate.py
@@ -1,4 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
 import copy
 import pickle
 import re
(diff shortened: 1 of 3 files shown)


pixelb (approved):

tests/instantiate/test_instantiate.py
```diff
@@ -1,4 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
```
Jasha10, line 2:
```suggestion
```
Reverting addition of accidental whitespace

libokj:
Is it possible to instantiate a list of objects from a yaml file?

Issue #1856, opened 2021-10-14
[Feature Request] Support python 3.10


jieru-hu:
including hydra-core and plugins.

Jasha10:
Is it worth supporting Python3.10 on the Hydra 1.1 branch as well?

PR #2015, opened 2022-02-04
Fix CI (for latest pytest version)

Jasha10:
A new pytest release came out earlier today.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -79,6 +79,10 @@ def find_dirs(path: str):
             yield fullname
 
 
+def _install_pytest(session):
+    session.install("pytest<7.0.0")
+
+
 def install_hydra(session, cmd):
     # needed for build
     session.install("read-version", silent=SILENT)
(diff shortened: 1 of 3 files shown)


Jasha10:
The CI failure on commit https://github.com/facebookresearch/hydra/pull/2015/commits/b9b3b2f15184accbed4159d59568c3176514c9d7 is coming from our jupyter notebook test. I've opened an [issue](https://github.com/computationalmodelling/nbval/issues/180) against the `nbval` repository. For I'll try pinning pytest...

noxfile.py
```diff
@@ -373,6 +377,7 @@ def test_core(session):
 @nox.session(python=PYTHON_VERSIONS)
 def test_plugins(session):
     _upgrade_basic(session)
+    _install_pytest(session)
```
jieru-hu, line 380: nit: since we need to always install pytest along with the upgrade_basic, could we merge them into one method, something like install_basic?

jieru-hu (commented):
thanks! one minor comment :)

noxfile.py
Jasha10: Actually the `lint` and `lint_plugins` nox sessions do not require install pytest (but they do still use upgrade_basic).

Jasha10: I have just opened PR #2016 which only pins pytest for the jupyter notebook session. I'm thinking of closing this PR in favor of that one. What do you think?

(One benefit of not pinning pytest is that the new pytest version should allow us to move forward with [python3.10 support](https://github.com/facebookresearch/hydra/issues/1856), which was previously blocked).

PR #2018, opened 2022-02-07
allow ax to sample log scale

jieru-hu:
original PR here https://github.com/facebookresearch/hydra/pull/1870

taking over due to no activity. I've addressed the comments to  not copy paste the test but instead use pytest param


website/docs/plugins/ax_sweeper.md
```diff
@@ -62,6 +62,8 @@ ax.modelbridge.dispatch_utils: Using Bayesian Optimization generation strategy:
...
 
+Tagging the interval with `log` will tell Ax to sample in log space (e.g. `python example/banana.py -m banana.x=tag(log, interval(1, 1000))`). When specifying a range sweep in a config file, include `log_scale: true` to achieve the same.
+
```
Jasha10, line 66: I think it would be helpful to give a concrete example for how to use `log_scale` in a config file. Would it make sense to move this paragraph later in the docs so that it comes after the [banana yaml example](https://github.com/facebookresearch/hydra/blob/e7ad04b300aba0c59d65cf1384e2f6f8a677cd8e/website/docs/plugins/ax_sweeper.md?plain=1#L68-L76)? That way, we could include the following to demonstrate usage of `log_scale`:
```yaml
banana.y:
 type: range
 bounds: [-5, 10.1]
 log_scale: true
```

Jasha10 (commented):
The implementation looks good. See my comment on the docs.
Also, this needs a news fragment.

plugins/hydra_ax_sweeper/news/1870.feature
```diff
@@ -0,0 +1 @@
+Add the ability to sample values from the log scale.
```
jieru-hu, line 1: @Jasha10 news segment is here :)

plugins/hydra_ax_sweeper/news/1870.feature
```diff
@@ -0,0 +1 @@
+Support sampling in log scale.
```
jieru-hu, line 1: @Jasha10 the  NEWS segment is here :)

Jasha10: Oh, awesome! I missed it somehow xD

website/docs/plugins/ax_sweeper.md
Jasha10: Looks good!

PR #2019, opened 2022-02-08
importlib plugin discovery

### LINKED ISSUE #1830: [Bug] Unable to use OptunaSweeperConf using the Python API
# 🐛 Bug
## Description

Hi,

First of all, many thanks for that wonderful tool you've made. It is, for sure, extremely helpful to have it when production environment is necessary.
I have met an error/bug when I wanted to use the **OptunaSweeperConf** using the Python API (instead of using the traditional YAML files). 

## Checklist

## To reproduce

```python
# Standard libraries

# Third-party libraries
import hydra
from hydra.conf import HydraConf
from hydra_plugins.hydra_optuna_sweeper.config import OptunaSweeperConf


@dataclass
class ExampleConfig():
    hydra: HydraConf = HydraConf(sweeper=OptunaSweeperConf())


ConfigStore.instance().store(name='config', node=ExampleConfig)


@hydra.main(config_path=None, config_name='config')
def main(config: DictConfig) -> None:
    """Main function."""
    print(config)


if __name__ == '__main__':
    main()

```

And the stack trace with the returned error:

```
In 'config': Validation error while composing config:
Merge error: OptunaSweeperConf is not a subclass of BasicSweeperConf. value: {'_target_': 'hydra_plugins.hydra_optuna_sweeper.optuna_sweeper.OptunaSweeper', 'defaults': [{'sampler': 'tpe'}], 'sampler': '???', 'direction': <Direction.minimize: 1>, 'storage': None, 'study_name': None, 'n_trials': 20, 'n_jobs': 2, 'search_space': {}}
    full_key: 
    object_type=dict

```

## Expected Behavior

I would like to override the hydra config group, directly into the structure config. To do so, I import the **HydraConf** from hydra and install the **hydra-optuna-sweeper** plugin.
The error seems to indicate that the **OptunaSweeperConf** must inherit from the **BasicSweeperConf** structured config (current implementation: https://github.com/facebookresearch/hydra/blob/v1.1.1/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py#L133).
The idea is to be able to override hydra config group directly from a python script. Does that make sense?

## System information
- **Hydra Version** :  1.1.1
- **Hydra Optuna Sweeper Version** :  1.1.1
- **Python version** : 3.8.11
- **Virtual environment type and version** : MiniConda environment
- **Operating system** : LinuxMint

Jasha10:
This PR hopefully closes #1830 by using importlib to load plugin modules instead of using pkgutil.

The diff ensures that, if user code loades a plugin or other object (such as a dataclass) from a plugin package, the object loaded in the user code will have the same identity as the object loaded by Hydra's plugin discovery mechanism.

### DIFF
--- a/hydra/core/plugins.py
+++ b/hydra/core/plugins.py
@@ -181,10 +181,8 @@ class Plugins(metaclass=Singleton):
                     if module_name.startswith("_") and not module_name.startswith("__"):
                         continue
                     import_time = timer()
-                    m = importer.find_module(modname)  # type: ignore
-                    assert m is not None
                     with warnings.catch_warnings(record=True) as recorded_warnings:
-                        loaded_mod = m.load_module(modname)
+                        loaded_mod = importlib.import_module(modname)
                     import_time = timer() - import_time
                     if len(recorded_warnings) > 0:
                         sys.stderr.write(
--- a/tests/standalone_apps/discovery_test_plugin/hydra_plugins/discovery_test/import_test_plugin.py
+++ b/tests/standalone_apps/discovery_test_plugin/hydra_plugins/discovery_test/import_test_plugin.py
@@ -1,0 +1,6 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+from hydra.plugins.plugin import Plugin
+
+
+class ImportTestPlugin(Plugin):
+    ...
--- a/tests/standalone_apps/discovery_test_plugin/tests/test_discovery.py
+++ b/tests/standalone_apps/discovery_test_plugin/tests/test_discovery.py
@@ -2,6 +2,8 @@
 import os
 from pathlib import Path
 
+from hydra_plugins.discovery_test.import_test_plugin import ImportTestPlugin
+
 from hydra.core.plugins import Plugins
 from hydra.plugins.plugin import Plugin
 
@@ -25,3 +27,10 @@ def test_skipped_imports(tmpdir: Path) -> None:
     assert "HiddenTestPlugin" not in discovered_plugins
 
     assert "NotHiddenTestPlugin" in discovered_plugins
+
+
+def test_imported_plugin_is_discovered_plugin() -> None:
+    # Tests that discovered object is the same object as imported object
+    discovered_plugins = {x.__name__: x for x in Plugins.instance().discover(Plugin)}
+    assert "ImportTestPlugin" in discovered_plugins
+    assert ImportTestPlugin is discovered_plugins["ImportTestPlugin"]


Jasha10:
Rebased against `main` to fix CI.

Jasha10:
I have run into a problem that I'm not sure how to resolve.

Essentially, the problem is that our tests depend on singleton state. Meanwhile, the singleton state depends on the side-effects of importing some modules. Using `pkgutil` to import these modules makes it easy to re-run the side-effects whenever they are needed, but using `importlib` means that the modules will be imported only once (and thus the side-effects cannot be re-run on demand).

Here is a demo illustrating the issue:
```python
# config_store.py
MockConfigStore = {}
```
```python
# configs.py
from config_store import MockConfigStore

MockConfigStore["my_structured_config"] = 123
```
If we use `pkgutil` to import `configs.py`, we can repeat the import's side-effect (i.e. mutation of `MockConfigStore` global state) as many times as we like.
```python
# test_with_pkgutil.py
import pkgutil
from config_store import MockConfigStore

def restore_singletons():
    MockConfigStore.clear()

def test1():
    restore_singletons()
    pkgutil.get_loader("configs").load_module()
    assert "my_structured_config" in MockConfigStore

def test2():
    restore_singletons()
    pkgutil.get_loader("configs").load_module()
    assert "my_structured_config" in MockConfigStore

test1()
test2()
print("success")
```
```commandline
$ python test_with_pkgutil.py
success
```
However, if we use `importlib` instead of `pkgutil`, we the side-effect of importing `configs.py` only happens once:
```python
# test_with_importlib.py
import importlib
from config_store import MockConfigStore

def restore_singletons():
    MockConfigStore.clear()

def test1():
    restore_singletons()
    importlib.import_module("configs")
    assert "my_structured_config" in MockConfigStore

def test2():
    restore_singletons()
    importlib.import_module("configs")
    assert "my_structured_config" in MockConfigStore

test1()
test2()
print("success")
```
```commandline
$ python test_with_importlib.py
  File "/home/rig1/dev/hydra/test_with_importlib.py", line 23, in <module>
    test2()
  File "/home/rig1/dev/hydra/test_with_importlib.py", line 19, in test2
    assert "my_structured_config" in MockConfigStore
AssertionError
```
Note that `test1` succeeds, but `test2` fails; the side-effect of importing `configs.py` occurs only once (in `test1`).

Jasha10:
Based on our team meeting, ideas for resolving this include:
1. Not calling `hydra_restore_singletons` during our tests, i.e. allowing global state to accumulate while our test suite runs.
2. Recommending that plugin modules expose a hook, e.g. a "`hydra_hook`" function. Defining such a hook in user code would allow Hydra's Plugin discovery mechanism to re-run the module's side-effects on demand. The `hydra_hook` function would make explicit calls such as `ConfigStore().store(..., node=MyUserConfig)` and `Plugins().register(MyUserPlugin)`.

Each approach has its drawbacks.
- Idea (1) would make our tests less independent of each other; testA could modify global state in a way that affects testB. Also, idea (1) may be incompatible with user workflows that involve clearing and restoring hydra's global state (e.g. calls to `hydra.initialize`, which [resets the singleton state](https://github.com/facebookresearch/hydra/blob/main/hydra/initialize.py#L20-L31)).
- Idea (2) would make the user-facing API more verbose. Also, idea (2) would have implications for backward compatibility. To help with migrating to a `hydra_hook` regime, we could issue a warning if a discovered plugin module does not define `plugin_hook`, or if plugin-discovery side effects cause hydra's global state to be mutated in a way that cannot be repeated via a call to `user_plugin_module.plugin_hook`.

mayeroa:
Hmm it seems that the issue is more complex than expected...
What approach seems the best for you? If I understand well, the first idea is not really satisfactory since it implies to rethink the way unit tests are working (due to global state persistence).
The second approach is interesting but wouldn't it too complex from a user perspective to understand the purpose of this hook?

Jasha10:
Thanks @mayeroa.
We are leaning towards the second approach (for Hydra 1.3).

lebrice:
Hey @Jasha10 , wouldn't using https://docs.python.org/3/library/importlib.html#importlib.reload in unit tests fix this problem?

omry:
This is already handled on current main for the supported Python versions. Plugin discovery reuses modules from `sys.modules`, and the original class identity failures no longer reproduce. Closing this old draft as superseded.

PR #1919, opened 2021-12-09
Update intro.md

Anirudh257:
## Motivation

Added a video explaining the nitty-gritty details of hydra.  

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebookresearch/hydra/blob/master/CONTRIBUTING.md)?

Yes

## Test Plan

No need of testing the code. Only the link needs to be checked.

## Related Issues and PRs

No

PR #2041, opened 2022-02-18
add 30min hydra concepts video to main page

pixelb:
This is to the right of the existing 1 minute overview video.
On space constrained screens, this reflows under that video.


pixelb:
@arjancodes is the source of the video

Jasha10:
This looks good to me.
@jieru-hu, I'm wondering if you have any of the same concerns with this as you had with #1919?

PR #2048, opened 2022-02-22
web: direct to github discussions rather than zulip

pixelb:
github discussions provide better integrations
and a more cohesive support surface.


jieru-hu (approved):

Jasha10 (approved):
Great!

PR #2050, opened 2022-02-22
Implement Orion as Hydra Sweeper

Delaunay:
## Motivation

Allow users to use Orion as a sweeper.

Oríon is an asynchronous framework for black-box function optimization.
Its purpose is to serve as a meta-optimizer for machine learning models and training, as well as a flexible experimentation platform for large scale asynchronous optimization procedures.
Core design value is the minimum disruption of a researcher’s workflow. It allows fast and efficient tuning, providing minimum simple non-intrusive (not even necessary!) helper client interface for a user’s script.

Effortless to adopt, deeply customizable

* [Natively asynchronous](https://orion.readthedocs.io/en/stable/code/core.html), thus resilient and easy to parallelize
* Offers the latest established hyperparameter [algorithms](https://orion.readthedocs.io/en/stable/user/algorithms.html)
* Elegant and rich [search-space definitions](https://orion.readthedocs.io/en/stable/user/searchspace.html)
* Comprehensive [configuration](https://orion.readthedocs.io/en/stable/user/config.html) system with smart defaults
* Transparent persistence in local or remote [database](https://orion.readthedocs.io/en/stable/install/database.html)
* [Integrate seamlessly](https://orion.readthedocs.io/en/stable/plugins/base.html) your own hyper-optimization algorithms

Yes

## Test Plan

Unit tested

### DIFF
--- a/plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
+++ b/plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
@@ -1,0 +1,388 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+from contextlib import contextmanager
+from copy import deepcopy
+from dataclasses import asdict
+import logging
+from typing import List, Optional, Sequence
+import os
+
+from hydra.core import utils
+from hydra.core.override_parser.overrides_parser import OverridesParser
+from hydra.core.override_parser.types import Override
+from hydra.core.plugins import Plugins
+from hydra.plugins.launcher import Launcher
+from hydra.plugins.sweeper import Sweeper
+from hydra.types import HydraContext, TaskFunction
+from hydra.core.utils import JobReturn
+
+from omegaconf import DictConfig, OmegaConf
+
+from orion.core.utils.flatten import flatten
+from orion.client import create_experiment
+from orion.client.experiment import ExperimentClient
+from orion.core.worker.trial import Trial, AlreadyReleased
+from orion.algo.space import Space, Dimension
+from orion.core.io.space_builder import DimensionBuilder, SpaceBuilder
+from orion.core.utils.exceptions import (
+    CompletedExperiment,
+    ReservationRaceCondition,
+    WaitingForTrials,
+    BrokenExperiment,
+    InvalidResult,
+)
+
+from .config import OrionClientConf, WorkerConf, AlgorithmConf, StorageConf
+
+log = logging.getLogger(__name__)
+
+
+def as_overrides(trial, additional):
+    """Returns the trial arguments as hydra overrides"""
+    kwargs = deepcopy(additional)
+    kwargs.update(flatten(trial.params))
+    return tuple(f"{k}={v}" for k, v in kwargs.items())
+
+
+def to_objective(value):
+    """Convert a return value into an Orion objective"""
+
+    if isinstance(value, (float, int)):
+        return [dict(name="objective", type="objective", value=value)]
+
+    if isinstance(value, dict):
+        return [value]
+
+    if isinstance(value, list):
+        return value
+
+    raise InvalidResult(
+        f"Value '{value}' of type '{str(type(value))}' is not an expected return type"
+    )
+
+
+class SpaceParser:
+    """Generate an Orion space from parameters and overrides"""
+
+    def __init__(self) -> None:
+        self.base_space = dict()
+        self.overrides = dict()
+        self.arguments = dict()
+
+    def space(self) -> Space:
+        """Generate the final space after overrides that will be used for the optimization"""
+        configuration = deepcopy(self.base_space)
+        configuration.update(self.overrides)
+        return SpaceBuilder().build(configuration), self.arguments
+
+    def add_from_parametrization(self, parametrization: Optional[DictConfig]) -> None:
+        """Use the parametrization retrieved from the configuration to generate a
+        preliminary research space
+
+        """
+        for k, v in parametrization.items():
+            try:
+                dim = DimensionBuilder().build(k, v)
+                self.base_space[dim.name] = dim.get_prior_string()
+            except (TypeError, NameError):
+                # Regular argument
+                self.arguments[k] = v
+
+    def add_from_overrides(self, arguments: List[str]) -> None:
+        """Create a dictionary of overrides to modify the research space"""
+        parser = OverridesParser.create()
+        parsed = parser.parse_overrides(arguments)
+
+        for override in parsed:
+            dim = self.process_overrides(override)
+
+            if dim is None:
+                self.arguments[override.get_key_element()] = override.value()
+            else:
+                self.overrides[dim.name] = dim.get_prior_string()
+
+    def process_overrides(self, override: Override) -> Dimension:
+        """Identify the sweep overrides and build a matching dimension"""
+        values = override.value()
+        name = override.get_key_element()
+
+        def build_dim(name):
+            builder = DimensionBuilder()
+            builder.name = name
+            return builder
+
+        if override.is_choice_sweep():
+            return build_dim(name).choices(*values.list)
+
+        elif override.is_range_sweep():
+            if values.step == 1:
+                return build_dim(name).uniform(values.start, values.stop, discrete=True)
+
+            choices = [
+                v for v in range(values.start, values.stop + values.step, values.step)
+            ]
+            return build_dim(name).choices(*choices)
+
+        elif override.is_interval_sweep():
+            discrete = type(values.start) is int
+            log = "log" in values.tags
+
+            cast_type = float
+            if discrete or values.start % 1 == values.end % 1 == 0.0:
+                cast_type = int
+
+            method = build_dim(name).uniform
+            if log:
+                method = build_dim(name).loguniform
+
+            return method(
+                cast_type(values.start), cast_type(values.end), discrete=discrete
+            )
+        else:
+            try:
+                # Not sweep override but could still be orion
+                return DimensionBuilder().build(name, values)
+            except (TypeError, NameError):
+                # Not a hyperparameter space definition
+                pass
+
+
+@contextmanager
+def clientctx(client):
+    try:
+        yield client
+    finally:
+        client.close()
+
+
+class OrionSweeperImpl(Sweeper):
+    def __init__(
+        self,
+        orion: OrionClientConf,
+        worker: WorkerConf,
+        algorithm: AlgorithmConf,
+        storage: StorageConf,
+        parametrization: Optional[DictConfig],
+    ):
+        self.orion_config = orion
+        self.worker_config = worker
+        self.algo_config = algorithm
+        self.storage_config = storage
+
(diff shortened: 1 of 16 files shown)


plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
```diff
@@ -0,0 +1,335 @@
...
+        def build_dim(name):
+            builder = DimensionBuilder()
+            builder.name = name
+            return builder
+
+        if override.is_choice_sweep():
+            return build_dim(name).choices(*values.list)
+
+        elif override.is_range_sweep():
+            choices = [v for v in range(values.start, values.stop, values.step)]
```
bouthilx, line 100: If values.step == 1 many algorithms would be more efficient if we convert to `'uniform(values.start, values.stop-1, discrete=True)'`.

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
```diff
@@ -0,0 +1,335 @@
...
+            return build_dim(name).choices(*choices)
+
+        elif override.is_interval_sweep():
+            discrete = type(values.start) is int
+            log = "log" in values.tags
+
+            cast_type = float
+            if discrete or values.start % 1 == values.end % 1 == 0.0:
```
bouthilx, line 108: `interval(1.0, 2.0)` would be casted to int?

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
```diff
@@ -0,0 +1,335 @@
...
+            space=self.space,
+            algorithms={algo_type: algo_config},
+            strategy=None,
+            max_trials=self.worker_config.max_trials,
+            max_broken=self.worker_config.max_broken,
+            storage=self.storage_config,
+            branching=self.orion_config.branching,
+            max_idle_time=None,
+            heartbeat=None,
+            working_dir=None,
```
bouthilx, line 221: Supporting working_dir would be important. Algorithms like PBT rely on it. Does it cause any problem with Hydra?

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
```diff
@@ -0,0 +1,347 @@
...
+                raise e
+
+    def release_all(self) -> None:
+        """Make sure not trials remain reserved"""
+        for trial in self.pending_trials:
+            try:
+                self.client.release(trial, status="interrupted")
+            except AlreadyReleased:
+                pass
+
```
bouthilx, line 259: This could be made part of clientctx.

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
Delaunay: Yes, but that does not matter since if discrete==False the result sampled will still be floats
This is just a cosmetic cast so ``space.configuration == input``

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
Delaunay: wouldn't be best to move it to Orion own ExperimentClient ?

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
Delaunay: Hydra has its own working dir configuration, I haven't looked at it yet

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
```diff
@@ -0,0 +1,347 @@
...
+    def optimize(self, client: ExperimentClient) -> None:
+        """Run the hyperparameter search in batches"""
+        failures = []
+
+        while not self.client.is_done:
+            trials = self.sample_trials()
+
+            if len(trials) == 0 and not self.client.is_done and not self.pending_trials:
+                raise RuntimeError()
```
bouthilx, line 268: Maybe break and log.warning that the algorithm could not sample more trials but the experiment is not considered done?

Delaunay: Actually, that was linked to a bug I tracked down, this is not necessary anymore

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
```diff
@@ -0,0 +1,347 @@
...
+
+            returns = self.execute_trials(trials)
+
+            self.observe_results(trials, returns, failures)
+
+            if len(failures) > self.worker_config.max_broken:
```
bouthilx, line 274:
```suggestion
            if self.client.is_broken:
```
This would better handle resumption (if possible with Hydra).

bouthilx: But we may have no failures if user is trying to resume a broken experiment...

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
Delaunay: done

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
```diff
@@ -0,0 +1,347 @@
...
+        trials: List[Trial],
+        returns: Sequence[JobReturn],
+        failures: Sequence[JobReturn],
+    ):
+        """Record the result of each trials"""
+        for trial, result in zip(trials, returns):
+            self.pending_trials.remove(trial)
+
+            if result.status == utils.JobStatus.COMPLETED:
+                value = result.return_value
```
bouthilx, line 312: It would be great to verify whether `value` is a list, in which case it would already represent the dicts results for Oríon, otherwise we build the dict as you do below. Some users return more than just the objective.

Delaunay: done

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
Delaunay: if the client is broken that means some workers had failures, that worker will raise the exception.
In that case if failures == 0 should we just log an error and not raise ?

Delaunay: to avoid printing to many exception with no debug info

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
bouthilx: It would be good to link Oríon's working dir to it then.

bouthilx: Forget it, it is not possible since trials may be resumed leading to different jobs in Hydra, and thus different job id and different working dir. We'll need to support working_dir in Oríon in parallel to working_dir in Hydra.

plugins/hydra_orion_sweeper/hydra_plugins/hydra_orion_sweeper/_impl.py
bouthilx: I think it would make sense to raise BrokenExperiment anyhow. But the stack-trace will not show one of the exceptions from previous worker execution.

Delaunay:
CLA as signed today not sure how long it takes to propagate

Delaunay:
Hello, 

@pixelb What would be the next steps to get this merged into the main branch ?

Thanks,

pixelb:
We're considering having a contrib/ dir to serve as a staging area for various new launchers / sweepers.
Your implementation looks good, and well integrated.
We'll do the necessary to merge it to contrib/.

Thank you for your contribution!

bouthilx:
@pixelb Thanks for the feedback! What will be the difference of support/usage for sweepers in contrib compared to actual sweepers (Ax, Nevergrad & Optuna)? Will they be included in the documentation as well?

jieru-hu:
hi @Delaunay thanks very much for you contribution.

We've created [contrib](https://github.com/facebookresearch/hydra/tree/main/contrib) here for plugin contributions. We've also included a README there.
Could you update this PR and move this plugin to the contrib folder?

pls let us know if you have any question on this. thanks again.

omry:
@Delaunay, @bouthilx:

Thank you for this contribution, and apologies for the very long delay here.

At a glance, the plugin looks good and appears to have been integrated in the expected way, though I have not done a deep review. I also want to acknowledge that the previous maintainer direction in 2022 was to move this into `contrib/`.

Since then, Hydra maintenance capacity has been limited, and one recurring difficulty has been keeping official and in-repo plugins up to date with their upstream dependencies. Even when plugin authors do most of the implementation work, continued maintenance, testing, releases, and CI coverage still create ongoing cost for the Hydra maintainers.

For that reason, I do not think we should add new fully supported plugins to this repository by default. More generally, there is no fundamental reason Hydra plugins need to live in the Hydra repo. The main benefit is discoverability, and I think we can provide that without taking ownership of the plugin code.

If you, or someone else from the Orion team, still want to pursue this, I think the better path is:

1. Host the plugin in the Orion repo, or in a standalone repo if that is more appropriate. It can depend on Hydra as an external dependency. If Hydra does not expose everything needed for plugin testing or development, we can discuss improving that.

2. Publish it as a community package, for example `hydra-contrib-orion-sweeper`, so users can distinguish it from officially maintained Hydra plugins.

3. We will create a community plugins section in the Hydra website, where this plugin can be listed and documented.

Let me know what you think. To keep the PR queue actionable, I plan to close this PR in 3 months if there is no response. The offer above stands regardless; if you want to pursue the external/community-plugin path later, please comment here.

image

omry:
Thank you again for this substantial contribution, and sorry for the long period without a clear resolution.

I revisited this while auditing the direction of Hydra's HPO sweepers. In addition to the maintenance concern described above, accepting this implementation in the Hydra repository would independently establish Orion-specific choices for search-space configuration, optimization results, failure handling, batching, and persistence while the corresponding shared HPO contracts are still being designed.

I am closing this pull request. The external community-plugin path described above remains open: the Orion project or another maintainer can host and publish the plugin independently, and Hydra can provide documentation and discoverability without owning its implementation.

PR #2052, opened 2022-02-23
fixup dev version stings: 1.2.3dev4 -> 1.2.3.dev4

Jasha10:
This PR updates dev semver strings of the form "1.2.3dev4" to "1.2.3.dev4".
The motivation here is:
  - consistency with [PEP 440](https://www.python.org/dev/peps/pep-0440/), which uses a period before "dev".
  - consistency with dev versions previously published on pypi.
    (for example, on pypi [we have previously published](https://pypi.org/project/hydra-core/#history) Hydra version "1.1.0.dev7", not "1.1.0dev7").

### DIFF
--- a/hydra/__init__.py
+++ b/hydra/__init__.py
@@ -1,7 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 # Source of truth for Hydra's version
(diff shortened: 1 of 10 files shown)


jieru-hu:
nice!

PR #2056, opened 2022-02-25
remove support for deprecated hydra override syntax

Jasha10:
This PR removes deprecated support for overriding `hydra` config groups without use of the keyword `override` (see this [deprecation notice](https://hydra.cc/docs/upgrades/1.0_to_1.1/defaults_list_override/)).

Removing this deprecated behavior enables the use of custom subgroups of the top-level hydra config.

#### Example:

Suppose we have a directory tree like this:
```text
├── conf
│   └── hydra
│       └── run
│           └── my_run.yaml
└── my_app.py
```
Let's say `my_app.py` is a minimal Hydra app:
```python
# my_app.py
import hydra

@hydra.main(config_path="conf", config_name=None)
def app(cfg):
    print(cfg)

if __name__ == "__main__":
    app()
```

Before this PR, it is impossible to add  `hydra/run: my_run` to the defaults list (since `hydra/run` does not appear in the defaults list of [`HydraConf`](https://github.com/facebookresearch/hydra/blob/c707ac2e82bfad4c2b29b88c2ffe6b12b7d7be71/hydra/conf/__init__.py#L108)).
```text
$ # Before this PR:
$ python my_app.py hydra.job.chdir=true +hydra/run=my_run
/Users/jasha10/hydra.git/hydra/_internal/defaults_list.py:413: UserWarning: In _dummy_empty_config_: Invalid overriding of hydra/run:
Default list overrides requires 'override' keyword.
See https://hydra.cc/docs/next/upgrades/1.0_to_1.1/defaults_list_override for more information.

  deprecation_warning(msg)
In '_dummy_empty_config_': Could not override 'hydra/run'. No match in the defaults list.

Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.
```
```text
$ # After this PR:
$ python my_app.py hydra.job.chdir=true +hydra/run=my_run
{}
```
This enabled use-case is relevant to [discussion #2051](https://github.com/facebookresearch/hydra/discussions/2051#discussioncomment-2237754).

PR #2065, opened 2022-03-03
Support backwards compatible operation by default

pixelb:
Add the "version_base" parameter to hydra.main,
which acts as central way to select the defaults for that version.

Without a specified "version_base" the user will get a warning,
and we'll default to the currently supported minimum version defaults
(currently those of version 1.1).

Specifying a "version_base" will suffice to select appropriate defaults,
rather than requiring one explicitly set other config items like
config_path, or hydra.job.chdir.

New hydra uses should use version_base=None to get
the default set of the currently installed version.

### DIFF
--- a/examples/advanced/defaults_list_interpolation/my_app.py
+++ b/examples/advanced/defaults_list_interpolation/my_app.py
@@ -4,7 +4,7 @@ from omegaconf import DictConfig, OmegaConf
 import hydra
 
 
-@hydra.main(config_path="conf", config_name="config")
+@hydra.main(version_base=None, config_path="conf", config_name="config")
 def my_app(cfg: DictConfig) -> None:
     print(OmegaConf.to_yaml(cfg))
 
--- a/website/docs/upgrades/version_base.md
+++ b/website/docs/upgrades/version_base.md
@@ -16,4 +16,4 @@ Also in this case, a warning is issued to indicate an explicit `version_base` is
 For example for Hydra 1.2, then would imply `config_path=None` and `hydra.job.chdir=False`.
 
 3. If the `version_base` parameter is an **explicit version string** like "1.1",
-then the default appropriate to that version are used.
+then the defaults appropriate to that version are used.
--- a/website/docs/upgrades/version_base.md
+++ b/website/docs/upgrades/version_base.md
@@ -1,0 +1,19 @@
+---
+id: version_base
+title: version_base
+---
+
+Hydra since version 1.2 supports backwards compatible upgrades by default
+through the use of the `version_base` parameter to **@hydra.main()** and **hydra.initialize()**.
+
+There are three classes of values that the `version_base` parameter supports,
+given new and existing users greater control of the default behaviors to use.
+
+1. If the `version_base` parameter is **not specified**, Hydra 1.x will use defaults compatible with version 1.1.
+Also in this case, a warning is issued to indicate an explicit `version_base` is preferred.
+
+2. If the `version_base` parameter is **None**, then the defaults are chosen for the current minor Hydra version.
+For example for Hydra 1.2, then would imply `config_path=None` and `hydra.job.chdir=False`.
+
+3. If the `version_base` parameter is an **explicit version string** like "1.1",
+then the default appropriate to that version are used.
--- a/examples/advanced/ad_hoc_composition/hydra_compose_example.py
+++ b/examples/advanced/ad_hoc_composition/hydra_compose_example.py
@@ -6,6 +6,6 @@ from hydra import compose, initialize
 if __name__ == "__main__":
     # initialize the Hydra subsystem.
     # This is needed for apps that cannot have a standard @hydra.main() entry point
-    initialize(config_path="conf")
+    initialize(version_base=None, config_path="conf")
     cfg = compose("config.yaml", overrides=["db=mysql", "db.user=${oc.env:USER}"])
     print(OmegaConf.to_yaml(cfg, resolve=True))
--- a/examples/advanced/config_search_path/my_app.py
+++ b/examples/advanced/config_search_path/my_app.py
@@ -4,7 +4,7 @@ from omegaconf import DictConfig, OmegaConf
 import hydra
 
 
-@hydra.main(config_path="conf", config_name="config")
+@hydra.main(version_base=None, config_path="conf", config_name="config")
 def my_app(cfg: DictConfig) -> None:
     print(OmegaConf.to_yaml(cfg))
 
--- a/examples/advanced/hydra_app_example/hydra_app/main.py
+++ b/examples/advanced/hydra_app_example/hydra_app/main.py
@@ -14,7 +14,7 @@ def add(app_cfg: DictConfig, key1: str, key2: str) -> Any:
     return ret
 
 
-@hydra.main(config_path="conf", config_name="config")
+@hydra.main(version_base=None, config_path="conf", config_name="config")
 def main(cfg: DictConfig) -> None:
     add(cfg.app, "num1", "num2")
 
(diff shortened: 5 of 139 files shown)


Jasha10:
Let me outline an argument in favor of using hydra's [`Singleton`](https://github.com/facebookresearch/hydra/blob/1aa46fbe460c7e3d136687d9ae5cda7f98aecfbb/hydra/core/singleton.py#L8) class to encapsulate the version base (instead of using `global __version_base__`):

Remote launcher plugins need to serialize Hydra's global state so that the remote process that will run the job can use the same global state. For example, the Hydra submitit launcher uses this pattern:
  - [create a list](https://github.com/facebookresearch/hydra/blob/1aa46fbe460c7e3d136687d9ae5cda7f98aecfbb/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py#L131) of jobs to launch, [including the `Singleton` state](https://github.com/facebookresearch/hydra/blob/1aa46fbe460c7e3d136687d9ae5cda7f98aecfbb/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py#L141) as an input for these jobs
  - call the submitit executor with this list of jobs
  - in each remote process corresponding to one of the jobs, [call `Singleton.set_state`](https://github.com/facebookresearch/hydra/blob/1aa46fbe460c7e3d136687d9ae5cda7f98aecfbb/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py#L60) before composing the config for that specific job.

Since the final config is composed on the remote process, that remote process must have access to Hydra's global state. I believe that this should include the version base, as the composition process will probably depend on the version base (e.g. PR #2056 removes deprecated functionality that affects config composition, and so the version base will play a role in finalizing that PR). The `Singleton.get_state` and `Singleton.set_state` methods allow simultaneously serializing or restoring the states of all of Hydra's singleton subclasses (e.g. `ConfigStore`, `Plugins`, `GlobalHydra`, etc) using a single access point.

examples/advanced/defaults_list_interpolation/my_app.py
```diff
@@ -4,7 +4,7 @@
 import hydra
 
 
-@hydra.main(config_path="conf", config_name="config")
+@hydra.main(version_base=None, config_path="conf", config_name="config")
```
jieru-hu, line 7: Thank you! overall it looks great!
My question is do we need to show `version_base=None` in all the examples? For new users starting afresh they probably won't need to know this till later (there is fair amount to ramp up on when first starting Hydra, so IMO less is better). 
For existing users, they are likely to change this version_base at upgrade time and we will include the pointer to version_base in the upgrade guide.

pixelb: Note that an unspecified version_base currently gives a warning.
Also I was removing settings of config_path=None as that's implied with version_base=None.
Also I didn't add version_base params to most of the intro docs for the reasons you state.
I'll push an update with docs, to hopefully clarify this more.
Happy to adjust after that

website/docs/upgrades/version_base.md
```diff
@@ -0,0 +1,19 @@
+---
+id: version_base
+title: version_base description
```
Jasha10, line 3:
```suggestion
title: version_base
```

website/docs/upgrades/version_base.md
```diff
@@ -0,0 +1,19 @@
...
+---
```
Jasha10, line 4: It would be good to add `'upgrades/version_base'` and `'upgrades/1.1_to_1.2/hydra_main_config_path'` to the `website/sidebars.js` file so that these new markdown files can be found via the hydra docs sidebar.

jieru-hu:
lg! should we add some notes on version_base in the developer guide as well? This is helpful to know  for introducing breaking changes.

Jasha10 (approved):
Looks good to me!

omry:
Looking back at this change, Hydra 1.2 needed a compatibility path so applications could migrate without adopting every changed behavior at once.

The problem was not the existence of a compatibility lever, but its blast radius and prominence. Omitting `version_base` selected legacy behavior and produced a warning, while new applications and examples had to specify `version_base=None`. This made a migration mechanism something every user encountered in `hydra.main`, Hydra’s most important public API. Supporting the old 1.1/1.2 boundary also distributed compatibility checks throughout the implementation.

This should have been an advanced opt-in mechanism from the start, reserved for meaningful breaking transitions where supporting both behaviors is reasonably inexpensive.

[Issue #3237](https://github.com/facebookresearch/hydra/issues/3237) applies that correction for Hydra 1.4: omission selects current behavior, `version_base=None` is deprecated, compatibility levels below 1.3 are unsupported, and future explicit compatibility levels should only be introduced when they provide clear migration value at a reasonable maintenance cost.

This preserves the useful migration capability without making it routine boilerplate for every Hydra application.

omry:
Revisiting this PR, the `version_base` API was designed backwards.

Omitting `version_base` warned and selected a compatibility level hard-coded to 1.1. Passing `version_base=None` selected the behavior of the installed Hydra version. This meant the natural spelling—omitting an optional argument—gave users legacy behavior, while obtaining normal current behavior required an explicit and non-obvious `None`.

That legacy default remained 1.1 in both Hydra 1.2 and Hydra 1.3. As a result, every new application had to know about Hydra’s compatibility machinery merely to request the behavior of the version it had installed.

The PR itself explicitly recommended `version_base=None` for new applications. Hydra then propagated that spelling throughout tutorials, examples, tests, and API documentation. Review feedback even questioned whether `None` needed to appear everywhere; that concern was correct. The design made it unavoidable boilerplate in Hydra’s primary public API.

There was also no plan to deprecate or remove `None`. It was introduced as the documented, supported way to select current behavior. We are now left trying to remove an argument that Hydra deliberately taught users to write.

The sensible design would have been the inverse from the beginning:

- Omission selects current behavior.
- Explicit version strings opt into compatibility behavior.
- Compatibility levels are added only for meaningful breaking transitions that are practical to maintain.

A compatibility mechanism can be useful. Making legacy compatibility the default and forcing every normal user to opt out with `None` was the mistake.

omry:
Closing the loop on this PR: `version_base` never became the versioned compatibility mechanism described here.

The current implementation has no Hydra 1.3 behavior boundary. Every compatibility check is against 1.2. Aside from the value exposed through `hydra.runtime.version_base`, the public states collapse to one bit:

- Omitted or `"1.1"`: enable legacy behavior.
- `None`, `"1.2"`, or `"1.3"`: disable legacy behavior.

That bit controls two old defaults (`config_path` and `hydra.job.chdir`) and eight deprecated pre-1.2 syntax and API paths. There are no distinct 1.2 and 1.3 compatibility modes.

Conceptually, this was like changing Java’s universal entry point from:

```java
public static void main(String[] args)
```

to:

```java
public static void main(String[] args, String jvmVersionBase)
```

and then requiring every application to pass `null` to request the normal behavior of its installed JVM, while omission selected legacy behavior and produced a warning.

The result is a multi-valued public API, global singleton state, runtime metadata, and mandatory `version_base=None` boilerplate controlling what is effectively a poorly named boolean. Hydra then taught every new user to include that workaround in applications, examples, tests, and documentation.

This abstraction will be removed, not extended with additional nominal compatibility levels.

The staged removal is now:

- Hydra 1.3.5 will warn applications that omit `version_base` or explicitly select `"1.1"`. Those are the only applications relying on behavior that Hydra 1.4 will remove.
- Hydra 1.4 will delete the remaining 1.1 compatibility paths. Omission will select the surviving behavior, explicit `"1.1"` will be an error, and other explicit values will warn that the obsolete argument must be removed.
- Hydra 1.5 will remove the `version_base` parameter, singleton state, comparison machinery, and `hydra.runtime.version_base` entirely.

The removal plan is tracked in https://github.com/facebookresearch/hydra/issues/3237. The Hydra 1.3 migration warning is tracked in https://github.com/facebookresearch/hydra/issues/3334.

PR #2087, opened 2022-03-11
`_partial_` docs: add note on speedup from repeated calls

Jasha10:
Addition to `_partial_` keyword docs, based on discussion with @jieru-hu.

Is this too much information?


Jasha10:
Todo: Implement the same change on v1.1 docs.

jieru-hu (approved):
thanks! lg!

PR #2094, opened 2022-03-15
Add '-' char in ID.

gleize:
## Motivation

The override grammar doesn't handle the `'-'` character in the ID of groups even though the yaml file does not have this issue.
For example, this is perfectly acceptable in the yaml file :
```yaml
defaults:
  - /models/classifier-heads@head: old_name
```
but the override parameter `'models/classifier-heads@head=new_name'` throws an error indicating the group ID is incorrect.

Yes

## Test Plan

* Ran `python setup.py antlr`.
* Ran `pytest` (python 3.8).
* Replaced my `hydra/grammar/gen` folder by newly generated one and tested on my setup similar to the example described above.

## Related Issues and PRs

N/A

### DIFF
--- a/hydra/grammar/OverrideLexer.g4
+++ b/hydra/grammar/OverrideLexer.g4
@@ -25,7 +25,7 @@ COLON: ':';
 SLASH: '/';
 
 KEY_ID: ID -> type(ID);
-KEY_SPECIAL: (CHAR|'_'|'$') (CHAR|DIGIT|'_'|'$')*;  // same as ID but allowing $
+KEY_SPECIAL: (CHAR|'_'|'$') (CHAR|DIGIT|'_'|'-'|'$')*;  // same as ID but allowing $
 DOT_PATH: (KEY_SPECIAL | INT_UNSIGNED) ('.' (KEY_SPECIAL | INT_UNSIGNED))+;
 
 ////////////////
@@ -60,7 +60,7 @@ BOOL:
 NULL: [Nn][Uu][Ll][Ll];
 
 UNQUOTED_CHAR: [/\-\\+.$%*@?|];  // other characters allowed in unquoted strings
-ID: (CHAR|'_') (CHAR|DIGIT|'_')*;
+ID: (CHAR|'_') (CHAR|DIGIT|'_'|'-')*;
 // Note: when adding more characters to the ESC rule below, also add them to
 // the `_ESC` string in `_internal/grammar/utils.py`.
 ESC: (ESC_BACKSLASH | '\\(' | '\\)' | '\\[' | '\\]' | '\\{' | '\\}' |
--- a/hydra/grammar/OverrideParser.g4
+++ b/hydra/grammar/OverrideParser.g4
@@ -53,7 +53,7 @@ dictKeyValuePair: dictKey COLON element;
 
 primitive:
       QUOTED_VALUE                               // 'hello world', "hello world"
-    | (   ID                                     // foo_10
+    | (   ID                                     // foo-bar_10
         | NULL                                   // null, NULL
         | INT                                    // 0, 10, -20, 1_000_000
         | FLOAT                                  // 3.14, -20.0, 1e-1, -10e3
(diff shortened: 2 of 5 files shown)


Jasha10:
Thanks for the PR, @gleize.

odelalleau:
> I don't see any problem with adding '-'. @odelalleau, would you be willing to give a second opinion?

Only thing that comes to mind is that in the past we thought of potentially adding some basic arithmetic operations, so that you could do something like `${x} - 1` and then `-` would become special somehow. That being said this is very far from a fully fleshed out idea and maybe it's not actually a problem at all.

The same change should be reflected in OmegaConf's grammar. Pay attention to the comments at the top of https://github.com/omry/omegaconf/blob/master/omegaconf/grammar/OmegaConfGrammarParser.g4

Jasha10:
Ok, thanks very much Olivier!

@gleize, could you please open a PR against OmegaConf making the same modification to the grammar?
- [this line](https://github.com/omry/omegaconf/blob/1d622e7823523710ba1e0f1542330704fea21a23/omegaconf/grammar/OmegaConfGrammarLexer.g4#L69) from `OmegaConfGrammarLexer.g4` should be updated
- [this line](https://github.com/omry/omegaconf/blob/1d622e7823523710ba1e0f1542330704fea21a23/omegaconf/grammar_parser.py#L28) from `grammar_parser.py` should be updated
- Adding a few test cases to `tests/test_grammar.py` would be much appreciated.

gleize:
Thanks @Jasha10 for the pointers.
Here is the OmegaConf PR : https://github.com/omry/omegaconf/pull/880

omry:
I think ID should not support -.
IDs are accessible at runtime via `cfg.foo`.
`cfg.foo-bar` is not legal python.
In addition, overring parameters with `-` could confuse argparse which if the - is prefixing the variable.

gleize:
> I think ID should not support -. IDs are accessible at runtime via `cfg.foo`. `cfg.foo-bar` is not legal python. In addition, overring parameters with `-` could confuse argparse which if the - is prefixing the variable.

One thing I've seen in other libraries to handle the non-pythonic IDs (e.g. `cfg.foo-bar.key`) is to add an alternative access method using `__getitem__` (e.g. `cfg["foo-bar"].key`). This would enable the library to be fully compatible with the JSON/YAML formats. Right now it feels a bit weird to write syntactically correct config files that are loaded properly but cannot be used with the overriding grammar.

Regarding argparse, I'm currently using this fork and haven't had any issue with argparse. But maybe you're thinking about a very specific case where it would fail ?

odelalleau:
> I think ID should not support -. IDs are accessible at runtime via `cfg.foo`. `cfg.foo-bar` is not legal python

Why should IDs be necessarily accessible at runtime via `cfg.foo`?

> In addition, overring parameters with `-` could confuse argparse which if the - is prefixing the variable.

Note that this PR actually doesn't allow `-` as first character so this shouldn't actually be an issue.

> One thing I've seen in other libraries to handle the non-pythonic IDs (e.g. `cfg.foo-bar.key`) is to add an alternative access method using `__getitem__` (e.g. `cfg["foo-bar"].key`)

That's what OmegaConf is doing as well, `cfg["foo-bar"]` works.

> Regarding argparse, I'm currently using this fork and haven't had any issue with argparse. But maybe you're thinking about a very specific case where it would fail ?

Hydra allows some command line parameters starting with `-` (https://hydra.cc/docs/advanced/hydra-command-line-flags/) => this might mess with them if `-` was allowed as first character.

Jasha10 (approved):

Jasha10:
After discussion on https://github.com/facebookresearch/hydra/issues/2363, I'm inclined to move forward with this PR.

> Only thing that comes to mind is that in the past we thought of potentially adding some basic arithmetic operations, so that you could do something like `${x} - 1` and then `-` would become special somehow. That being said this is very far from a fully fleshed out idea and maybe it's not actually a problem at all.

Here's a cross-link to the OmegaConf feature request for arithmetic support: https://github.com/omry/omegaconf/issues/91 .

pixelb (approved):

Issue #2100, opened 2022-03-18
Hydra-submitit launcher does not log error messages to log files.


abhiskk:
# 🚀 Feature Request

Logging error messages in hydra-submitit launches to relevant log files. Currently the error messages are ignored by the hydra-submitit setup and are only piped to the terminal.

## Motivation

I have been using the hydra-submitit launcher for multi-node training and have been observing the above behavior. If the messages are logged to the relevant log files it will give us consolidation of the log messages and error message and will make it easy to track experiments. The logging of messages to terminal is also painful in the scenario when you launch hyperparameter searches with a single launch command and the messages from multiple hyperparameter search runs come to the same terminal.

For e.g. if we launch the runs we get a message like this on the terminal:

```
INFO: launch command: python train_d2_net.py -m         --config-path=../configs         --config-name=dummy.yaml         +mode=cluster         hydra.sweep.dir=/checkpoint/akadian/exps/20220316T012225/20220316T012225LOGS         use_wandb=true hydra.launcher.timeout_min=4320 D2.SOLVER.BASE_LR=0.02,0.002,0.0002,0.00002
[2022-03-16 01:23:26,351][HYDRA] Submitit 'slurm' sweep output dir : /checkpoint/akadian/exps-w3-reasoning/20220316T012225/20220316T012225LOGS
[2022-03-16 01:23:26,352][HYDRA]  #0 : +mode=cluster use_wandb=True wandb_project=lvis_debug D2.SOLVER.BASE_LR=0.02
[2022-03-16 01:23:26,358][HYDRA]  #1 : +mode=cluster use_wandb=True wandb_project=lvis_debug D2.SOLVER.BASE_LR=0.002
[2022-03-16 01:23:26,363][HYDRA]  #2 : +mode=cluster use_wandb=True wandb_project=lvis_debug D2.SOLVER.BASE_LR=0.0002
[2022-03-16 01:23:26,367][HYDRA]  #3 : +mode=cluster use_wandb=True wandb_project=lvis_debug D2.SOLVER.BASE_LR=2e-05
```

If any of these experiments crash then the log gets piped to the terminal using which we had submitted the jobs but they are not present in the relevant log files where everything else is logged, like: `.submitit/54459356/54459356_0_log.out, .submitit/54459356/54459356_0_log.err`.

For eg this error message appeared on the terminal which was used to submit the jobs, but was never piped to the relevant log files.

```
[...]

PR #2111, opened 2022-03-25
Add job return logging

jieru-hu:
I tried a fews ways to add the logging in hydra core without a callback. However didn't find a clean way to add the logging in both RUN and MULTIRUN mode without adding much more noise to the console. 

Add logging as a callback means: 1) job return will be logged together with the rest of the application log, 2) users can opt in.

related to #2100

### DIFF
--- a/website/docs/experimental/callbacks.md
+++ b/website/docs/experimental/callbacks.md
@@ -5,9 +5,6 @@ sidebar_label: Callbacks
 ---
 
 import GithubLink from "@site/src/components/GithubLink"
-import {ExampleGithubLink} from "@site/src/components/GithubLink"
-
-<ExampleGithubLink text="Examples" to="hydra/experimental"/>
 
 The <GithubLink to="hydra/experimental/callback.py">Callback interface</GithubLink> enables custom
 code to be triggered by various Hydra events.
--- a/website/docs/experimental/callbacks.md
+++ b/website/docs/experimental/callbacks.md
@@ -5,7 +5,9 @@ sidebar_label: Callbacks
 ---
 
 import GithubLink from "@site/src/components/GithubLink"
+import {ExampleGithubLink} from "@site/src/components/GithubLink"
 
+<ExampleGithubLink text="Examples" to="hydra/experimental"/>
 
 The <GithubLink to="hydra/experimental/callback.py">Callback interface</GithubLink> enables custom
 code to be triggered by various Hydra events.
--- a/hydra/experimental/log_job_return_callback.py
+++ b/hydra/experimental/log_job_return_callback.py
@@ -1,0 +1,23 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import logging
+from typing import Any
+
+from omegaconf import DictConfig
+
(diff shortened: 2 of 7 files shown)


pixelb (approved):
Is this a callback we might be able to enable by default?
I.e. what are the disadvantages of enabling it?
Does it result in more console output?

In any case the code looks good. thanks

website/docs/experimental/callbacks.md
```diff
@@ -5,7 +5,9 @@ sidebar_label: Callbacks
 ---
 
 import GithubLink from "@site/src/components/GithubLink"
+import {ExampleGithubLink} from "@site/src/components/GithubLink"
 
+<ExampleGithubLink text="Examples" to="hydra/experimental"/>
```
Jasha10, line 10: I think users might be confused when this "Examples" button redirects to the `hydra/experimental` directory; there are some things in that directory that are unrelated to callbacks.

What if we create a file `hydra/experimental/callbacks.py` that contains all the callback implementations (instead of using separate files `log_job_return_callback.py` and `pickle_job_info_callback.py`)? The `ExampleGithubLink` in the docs could then point to that specific file `callbacks.py`.
This would make the yaml files simpler too: users would write `_target_: hydra.experimental.callbacks.LogJobReturnCallback` instead of `_target_: hydra.experimental.log_job_return_callback.LogJobReturnCallback`).

Jasha10:
> Is this a callback we might be able to enable by default? I.e. what are the disadvantages of enabling it? Does it result in more console output?

It does result in more console output. Here is an example:
```commandline
$ cd examples/tutorials/basic/running_your_hydra_app/5_basic_sweep
$ python my_app.py -m db=mysql,postgresql '+hydra.callbacks.foo._target_=hydra.experimental.log_job_return_callback.LogJobReturnCallback' hydra.job.chdir=true
[2022-03-27 09:14:38,045][HYDRA] Launching 4 jobs locally
[2022-03-27 09:14:38,045][HYDRA]        #0 : db=mysql db.timeout=5
driver=mysql, timeout=5
[2022-03-27 09:14:38,141][hydra.experimental.log_job_return_callback][INFO] - Succeeded with return value: None
[2022-03-27 09:14:38,142][HYDRA]        #1 : db=mysql db.timeout=10
driver=mysql, timeout=10
[2022-03-27 09:14:38,233][hydra.experimental.log_job_return_callback][INFO] - Succeeded with return value: None
[2022-03-27 09:14:38,233][HYDRA]        #2 : db=postgresql db.timeout=5
driver=postgresql, timeout=5
[2022-03-27 09:14:38,323][hydra.experimental.log_job_return_callback][INFO] - Succeeded with return value: None
[2022-03-27 09:14:38,324][HYDRA]        #3 : db=postgresql db.timeout=10
driver=postgresql, timeout=10
[2022-03-27 09:14:38,414][hydra.experimental.log_job_return_callback][INFO] - Succeeded with return value: None
```

website/docs/experimental/callbacks.md
jieru-hu: That's a great idea! Since this would mean touching the PickleJobCallback as well, I will create that PR as a follow up. 
For now I will remove this change.

Issue #1891, opened 2021-11-15
Clean up Hydra 1.1 deprecations


jieru-hu:
We have some deprecation clean up to do for Hydra 1.2. We will need to go through them 1 by 1.

Here is a not necessarily comprehensive [list](https://github.com/facebookresearch/hydra/search?q=%23+DEPRECATED%3A+remove+in+1.2 )

PR #2136, opened 2022-04-08
remove support for _name_ and _group_ from package header

pixelb:
(when version_base >= 1.2)

Was deprecated, and slated for removal in 1.2.

Addresses issue #1891

### DIFF
--- a/hydra/core/default_element.py
+++ b/hydra/core/default_element.py
@@ -7,6 +7,7 @@ from typing import List, Optional, Pattern, Union
 from omegaconf import AnyNode, DictConfig, OmegaConf
 from omegaconf.errors import InterpolationResolutionError
 
+from hydra import version
 from hydra._internal.deprecation_warning import deprecation_warning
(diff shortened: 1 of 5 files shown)


hydra/core/default_element.py
```diff
@@ -116,19 +117,20 @@ def set_package_header(self, package_header: Optional[str]) -> None:
...
+                deprecation_warning(
+                    message=dedent(
+                        f"""\
+                        In '{path}': Usage of deprecated keyword in package header '# @package {package_header}'.
+                        See {url} for more information"""
+                    ),
+                )
+
+            if package_header == "_group_":
+                return
```
jieru-hu, line 133: hmm not sure if I missed anything here...  it seems like this PR only removes the warning but doesn't actually remove the support?

pixelb: Well the special "_group_" handling is only for < "1.2",
but you're right, I need to avoid the "_group_" handling on line 143 also

PR #2154, opened 2022-04-14
document instantiation of targets from builtins module

### LINKED ISSUE #2153: Document use of `builtins` module with `instantiate`
To use `instantiate` with a `_target_` such as `len` or `dict`, it's necessary to make reference to the `builtins` module.
```python
# instantiate({"_target_": "len"}, [1,2,3])  # this gives an InstantiationException
instantiate({"_target_": "builtins.len"}, [1,2,3])  # this works, returns the number 3
```
It would be good to document this as it has come up in some user issues (e.g. #2148).

Jasha10:
This PR documents use of instantiation targets in the `builtins` module.
Closes #2153.

### DIFF
--- a/website/docs/advanced/instantiate_objects/overview.md
+++ b/website/docs/advanced/instantiate_objects/overview.md
@@ -368,3 +368,16 @@ assert bar1.foo is bar2.foo  # the `Foo` instance is re-used here
 ```
 This does not apply if `_partial_=False`,
 in which case a new `Foo` instance would be created with each call to `instantiate`.
+
+
+### Instantiation of builtins
+
+The value of `_target_` passed to `instantiate` should be a "dotpath" pointing
+to some callable that can be looked up via a combination of `import` and `getattr`.
+If you want to target one of Python's [built-in functions](https://docs.python.org/3/library/functions.html) (such as `len` or `print` or `divmod`),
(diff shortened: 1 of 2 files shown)


website/docs/advanced/instantiate_objects/overview.md
```diff
@@ -368,3 +368,23 @@ assert bar1.foo is bar2.foo  # the `Foo` instance is re-used here
...
+from hydra.utils import instantiate
+# instantiate({"_target_": "len"}, [1,2,3])  # this gives an InstantiationException
+instantiate({"_target_": "builtins.len"}, [1,2,3])  # this works, returns the number 3
+```
+To look up the `len` built-in function in the above example,
+Hydra splits the dotpath `"builtins.len"` and employs logic equivalent to the following:
+```
+from importlib import import_module
+mod = import_module("builtins")
+target = getattr(mod, "len")
```
pixelb, line 389: This is a bit specific / uneeded.
There is enough info previously I think

pixelb (approved):

website/docs/advanced/instantiate_objects/overview.md
Jasha10: Ok, I've removed the 6 lines above.

Issue #2162, opened 2022-04-20
Upgrading Optuna from v2.x.x to v3.0.0


himkt:
Sorry but it is not a feature request.

## Summary

Thank you for integrating Optuna, hyperparameter optimization library. We, Optuna dev-team, are working on the next major version of the library Optuna v3. In Optuna v3, we will deprecate distributions used in:https://github.com/facebookresearch/hydra/blob/61ab29ac07c7ff4f87836437fb8f76166f9c549d/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py#L33-L42

We will introduce `FloatDistribution` and `IntDistribution`. So distributions in hydra_optuna_sweeper will have to be updated:

- `UniformDistribution(low, high)` => `FloatDistribution(low, high)`
- `DiscreteUniformDistribution(low, high, q)` => `FloatDistribution(low, high, step)`
- `LogUniformDistribution(low, high)` => `FloatDistribution(low, high, log=True)`
- `IntUniformDistribution(low, high)` => `IntDistribution(low, high)`
- `IntUniformDistribution(low, high, step)` => `IntDistribution(low, high, step)`
- `IntLogUniformDistribution(low, high)` => `IntDistribution(low, high, log=True)`

Now hydra optuna plugin installs the latest Optuna. Users will see the deprecation warnings in the Optuna v3 release. This PR restricts the constraint to avoid such warnings. Note that it doesn't mean Optuna v3 is imcompatible with Optuna v2. It only shows warnings and deprecated distributions are internally converted to appropriate distributions.

```
root@b9ffe3ebef50:/work/plugins/hydra_optuna_sweeper# python example/sphere.py --multirun
/work/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py:109: FutureWarning: DiscreteUniformDistribution has been deprecated in v3.0.0. This feature will be removed in v6.0.0. See https://github.com/optuna/optuna/releases/tag/v3.0.0. Use :class:`~optuna.distributions.FloatDistribution` instead.
  return DiscreteUniformDistribution(value.start, value.stop, value.step)
[I 2022-04-20 06:13:41,041] A new study created in memory with name: sphere
[2022-04-20 06:13:41,042][HYDRA] Study name: sphere
[2022-04-20 06:13:41,043][HYDRA] Storage: None
[2022-04-20 06:13:41,043][HYDRA] Sampler: TPESampler
[...]

PR #2163, opened 2022-04-20
Add version constraint for Optuna

himkt:
See https://github.com/facebookresearch/hydra/issues/2162 for more information.

## Motivation

This PR updates the version constraint of Optuna to avoid installing the next major version of Optuna.
See also 🔗 https://github.com/facebookresearch/hydra/issues/2162

Yes

## Test Plan

## Related Issues and PRs

https://github.com/facebookresearch/hydra/issues/2162

### DIFF
--- a/plugins/hydra_optuna_sweeper/setup.py
+++ b/plugins/hydra_optuna_sweeper/setup.py
@@ -27,7 +27,7 @@ setup(
     ],
     install_requires=[
         "hydra-core>=1.1.0.dev7",
-        "optuna<3.0.0",
+        "optuna>=2.10.0,<3.0.0",
     ],
     include_package_data=True,
 )
--- a/plugins/hydra_optuna_sweeper/setup.py
+++ b/plugins/hydra_optuna_sweeper/setup.py
@@ -27,7 +27,7 @@ setup(
     ],
     install_requires=[
         "hydra-core>=1.1.0.dev7",
-        "optuna>=2.10.0",
+        "optuna<3.0.0",
     ],
     include_package_data=True,
 )


Jasha10 (approved):
Thanks very much for letting us know about the upcoming Optuna API changes!
Once Optuna 3.0 is released, we can upgrade this sweeper to use the new API.

plugins/hydra_optuna_sweeper/setup.py
```diff
@@ -27,7 +27,7 @@
     ],
     install_requires=[
         "hydra-core>=1.1.0.dev7",
-        "optuna>=2.10.0",
```
jieru-hu, line 30: could you keep the "optuna>=2.10.0"  in the meanwhile?

jieru-hu (changes requested):

plugins/hydra_optuna_sweeper/setup.py
himkt: So you mean it's enough to track the issue #2162 and not necessary to constrain the version, right? It will show warning messages for users when Optuna v3 is available. Is it OK?

Jasha10: Just throwing this out there: using multiple constraints on the version should be legal:
```python
    install_requires=[
        ...
        "optuna >= 2.10.0, < 3.0.0",
        ...
```
The downside of pinning Optuna < 3.0 is that users would not be able to access new Optuna features. Jieru, is why you were concerned about the pin?

jieru-hu: sorry for the confusion, i should have provvided more context.  I meant `"optuna >= 2.10.0, < 3.0.0"`

the current impl assumes the version is >=2.10.0 so it'd be great to keep that still

himkt: Thank you @Jasha10 and @jieru-hu! Sorry for my poor understanding. 4b39011 restore the lower constraint. 🙋

jieru-hu: thank you @himkt!

PR #2166, opened 2022-04-21
Support Python3.10 on main branch

Jasha10:
- The most significant change is to `hydra/core/plugins.py`; the previous strategy used for importing plugins was causing a `DeprecationWarning` when I tested locally on Python3.10.
- Our ray launcher does not support python3.10 (as the ray library on which it depends only works with <= 3.9). The CI for ray+py3.10 is "passing" (i.e. it is green) but the ray tests are not actually running.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -271,19 +271,19 @@ workflows:
       - test_macos:
           matrix:
             parameters:
-              py_version: ["3.6", "3.7", "3.8", "3.9"]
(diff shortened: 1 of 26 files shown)


Jasha10:
The CI seems to be working now except for the nevergrad plugin tests on Python3.10 + macOs/Linux, which fails during the pip installation of numpy.

Jasha10:
I'm closing this PR in favor of #2183; this PR's plugin tests will not run (issue #2181).

Issue #692, opened 2020-06-19
[Feature Request] Support asynchronous launching


jan-matthis:
# 🚀 Feature Request

Some launchers enqueue jobs rather than executing them directly. It'd be great if Hydra supported that. 

The interim solution in the RQ launcher (#683) is to throw an exception after enqueuing, which doesn't make for a great user experience.

See also: https://github.com/facebookresearch/hydra/pull/683#discussion_r442920298

omry:
Tentative for 1.0.

omry:
Launching enhancements will come in 1.2 or 1.3.

RemyLau:
I think non-blocking launching, especially for `submitit`, could be very helpful. Seems like a naive solution for this would be simply bypassing the following two lines when the non-blocking option (say `submitit_slurm_nonblocking`) is set

https://github.com/facebookresearch/hydra/blob/61ab29ac07c7ff4f87836437fb8f76166f9c549d/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py#L145-L146

and replace them with `return None`.

Apparently, this assumes that the target function does not return anything (usually directly saving results to some files for example), which is typically how I'm setting things up.

Would this be a reasonable patch to be implemented for the non-blocking feature?

Jasha10:
@RemyLau what use-case would be enabled by removing the line you mentioned above?

RemyLau:
Hi @Jasha10, the way I've been using the `submitit_slurm` plugin is to mimic how I would use SLURM to submit a bunch of jobs to run a script with different settings via job arrays. In short, each job would save its own result, so I don't need a master worker to aggregate all the results. Hence, there's no need to block and wait for all jobs to be finshed.

### Usual setup (without Hydra)

Here's an example of my usual main script (say `src/main.py`), which consists of four major components:
1. Load configurations based on the selected option
2. Load data
3. Run some heavy computation on the data with the configurations, which usually takes hours to run
4. Save the results once the computations are done

```python
def main() -> None:
    parser = argparse.ArgumentParser()
    parser.addargument("--cfg", type=int, help="Configuration id")
    args = parser.parse_args()
    # Here I'm simplifying the configuration loading process.. in reality, it's much messier....
    cfg = load_cfg(args.cfg)

    data = load_data(cfg.data_dir)
    results = heavy_computations(data, cfg.my_params)
    save_results(cfg.result_dir, results)

if __name__ == "__main__":
[...]

PR #2171, opened 2022-04-23
Add non-blocking slurm job submission option

RemyLau:
## Motivation

Yes

## Test Plan

No special setup is required.

## Related Issues and PRs

#692

### DIFF
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence
 
 from hydra.core.singleton import Singleton
-from hydra.core.utils import JobReturn, filter_overrides, run_job, setup_globals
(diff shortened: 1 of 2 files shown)


plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/submitit_launcher.py
```diff
@@ -143,6 +143,13 @@ def launch(
             )
 
         jobs = executor.map_array(self, *zip(*job_params))
+        if self.config.hydra.launcher.submitit_nonblocking:
+            null_returns = []
+            for _ in jobs:
+                null_returns.append(JobReturn())
+                null_returns[-1].status = JobStatus.COMPLETED
```
Jasha10, line 150: I think `JobStatus.UNKNOWN` would technically be a more accurate description than `JobStatus.COMPLETED`.

omry:
Thanks for contributing this and for clearly explaining the use case.

This approach is incompatible with Hydra’s launcher design. A launcher must return the actual `JobReturn` values for the jobs it launches. Returning fabricated `COMPLETED` results while jobs are still running misrepresents their state and discards their eventual results and failures.

Supporting asynchronous submission would require a different launcher and sweeper execution contract, rather than a Submitit-specific option within the current contract. We are therefore rejecting this PR. The broader design problem is tracked separately in #3201.

PR #2172, opened 2022-04-25
add contrib folder

jieru-hu:

PR #2173, opened 2022-04-25
Skip contrib in CI

jieru-hu:
a follow up to #2172

tested by adding files that should fail CI under contrib and verify CI still passes


pixelb (approved):
thanks!

PR #2180, opened 2022-04-28
avoid running plugin tests on forked PRs

pixelb:
As per https://circleci.com/blog/managing-secrets-when-you-have-pull-requests-from-outside-contributors/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,18 @@ parameters:
     default: v1
 
 commands:
+  early_return_for_forked_pull_requests:
+    description: >-
+      If this build is from a fork, stop executing the current job and return success.
+      This is useful to avoid steps that will fail due to missing credentials.
+    steps:
+      - run:
+          name: Early return if this build is from a forked PR
+          command: |
+            if [ -n "$CIRCLE_PR_NUMBER" ]; then
+              echo "Nothing to do for forked PRs, so marking this step successful"
+              circleci step halt
+            fi
   macos:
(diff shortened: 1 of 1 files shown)


pixelb:
I've confirmed that this does not run circleci_pipeline.py on this PR based on a separate repo,
and does run it on a PR froma branch on the main repo

jieru-hu:
How about we skip ray launcher tests only? IMHO it is still valuable to run tests on other plugins for forked PR since we do see more external contributions for some of them.
if that makes sense, we can modify this script https://github.com/facebookresearch/hydra/blob/main/tools/ci/circleci_pipeline.py instead of the circle CI yaml?

pixelb:
Two issues with that I see.
- We still make ci token available in environment.
- We still make aws tokens available in environment for non ray plugins
I'm not sure we can effectively remove env vars programatically

pixelb:
github seems to be linking the wrong tests for this pr (using the local repo branch name instead).

The correct tests are: https://app.circleci.com/pipelines/github/facebookresearch/hydra?branch=pull%2F2180&filter=all

jieru-hu (approved):
thanks!
per our offline discussions, we will look into adding the rest of the plugin tests directly into `config.yml` instead of from calling the API (to avoid exposing the other tokens.)

pixelb:
> thanks! per our offline discussions, we will look into adding the rest of the plugin tests directly into `config.yml` instead of from calling the API (to avoid exposing the other tokens.)

Follow up in issue #2181

PR #2185, opened 2022-04-29
resolve ray init in ray launcher

colobas:
## Motivation

Trying to use hydra and ray, when providing the init config it doesn't get resolved to primitive types and it fails when I set, for instance, `runtime_env`. It's solved easily by calling `OmegaConf.to_container` before passing the config to `ray.init` - (which I notice is already being done for the `ray.remote` config)

Yes/No

## Test Plan

## Related Issues and PRs


Jasha10:
Hi @colobas, thanks for the contribution!
It may take us some time before we can properly review this PR -- we are currently working through some issues with our plugin CI tests, and we are also in a busy period before the Hydra 1.2 release.

In the meantime, could you please add a test case to the ray plugin tests? It should be a test that would have failed before, but that will pass now that you've updated `_launcher_util.py`.

jieru-hu:
thanks @colobas for spotting the inconsistency and creating the PR. I've updated the launcher in a recent PR along with some schema / dependency updates along with the fix in the PR.

Issue #2155, opened 2022-04-14
[Feature Request] Is this possible to loose the version of antlr4-python3-runtime


WangCHX:
# 🚀 Feature Request
Since my company is using monorepo (bazel), the version of each package is fixed. Since hydra-core fix the  `antlr4-python3-runtime` to 4.8, my other package which needs 4.9 can't be added into the monorepo. So I am wondering if hydra-core loose the `antlr4-python3-runtime` to something like `>=4.8`. 

## Motivation
same as feature request.

## Pitch
maybe make `antlr4-python3-runtime` >= 4.8 ?

Jasha10:
cc @pixelb

pixelb:
Note antlr-4.9.3 is required for python-3.6 support

PR #2192, opened 2022-05-05
update from antlr-4.8 to antlr-4.9

pixelb:
Use the latest 4.9 release (4.9.3)

Addresses issue #2155

Needs to follow omegaconf/pull/912 (master branch)


pixelb (commented):
@Jasha10 I don't know how valid my change here is TBH

Jasha10:
It looks like the CI failure happened when OmegaConf's [`AnyNode`](https://github.com/omry/omegaconf/blob/04ecbc13fa84a53d14468cc9b1836dbd92cd7431/omegaconf/nodes.py#L139-L141) class refused to wrap a `SimpleDataClass`:
```text
.nox/test_core-3-10/lib/python3.10/site-packages/omegaconf/nodes.py:139: in _validate_and_convert_impl
    raise UnsupportedValueType(
E   omegaconf.errors.UnsupportedValueType: Value 'DictConfig' is not a supported primitive type
E       full_key: a
E       object_type=SimpleDataClass
```

I don't think setting `allow_objects=True` is the best approach here. Setting `allow_objects=True` makes AnyNode more flexible (allowing any object type instead of just primitive object types)... but there is still the question of why SimpleDataClass is being wrapped in AnyNode instead of being treated as a structured config.

Jasha10:
I'm confused because the PR https://github.com/omry/omegaconf/pull/911 was supposed to be a pure refactoring without any behavior change.

pixelb:
> I'm confused because the PR [omry/omegaconf#911](https://github.com/omry/omegaconf/pull/911) was supposed to be a pure refactoring without any behavior change.

Yes this PR is just coincidental it seems. Reverting just that change, we're still left with the issue

Jasha10:
It seems that the CI failed not just for the OmegaConf dev release -- it failed for the OmegaConf 2.1 install as well.
This makes me think that the issue is on the Hydra side...

pixelb:
> ~It seems that the CI failed not just for the OmegaConf dev release -- it failed for the OmegaConf 2.1 install as well. This makes me think that the issue is on the Hydra side...~ Nevermind, I missed that you had bumped the pin on OmegaConf.

Right. I just changed the pin to 2.2.0.dev2 (from April 13), and the following is fine:
```
pip install -r requirements/dev.txt 
(hydra39) pbrady-fedora-PF1C3VX9:hydra (main *)$ pytest tests/instantiate/test_instantiate.py
```

So that narrows it down a lot

Jasha10:
I just did a git bisect to find that PR https://github.com/omry/omegaconf/pull/909 is the culprit.

Jasha10:
I'll send in a PR to revert https://github.com/omry/omegaconf/pull/909.

Jasha10 (approved):

PR #2197, opened 2022-05-09
Fix plugin CI

Jasha10:
I believe CI is failing on the `main` branch.


pixelb:
`error: On Python 3 formatting "b'abc'" with "{}" produces "b'abc'", not "abc"; use "{!r}" if this is desired behavior`

Might that be triggered by the updated reference to OmegaConf 2.2?
The plugin params being processed when this error message hits, come from OmegaConf.create(...)

Jasha10:
I suspect so -- OmegaConf 2.2 supports `bytes` keys/values while 2.1 did not.
I encountered a similar mypy error when working on the BytesNode PR https://github.com/omry/omegaconf/pull/872

Jasha10:
The mypy fix is merged. CI is still failing from an unrelated failure of rq_launcher.

PR #2205, opened 2022-05-11
remove support for legacy default list interpolation format

pixelb:
(when the version_base is >= 1.2).

The old ${defaults.N.name} format was deprecated,
and slated for removal in 1.2.

Follows on from the support of the newer format in PR #1044

Addresses issue #1891


jieru-hu:
looks good. probably need a NEWs segment?

jieru-hu (approved):

pixelb:
> looks good. probably need a NEWs segment?

Yes I'll looks at doing a NEWs for all these deprecated items in a separate change I think
thanks for the review

PR #2207, opened 2022-05-11
Add plugin tests back

### LINKED ISSUE #2181: Reinstate non aws plugin testing on forked PRs
This was disabled with https://github.com/facebookresearch/hydra/pull/2180
That was so the AWS and CircleCI tokens are not shared to jobs for forked PRs

To reenable we may use explicit circleci config to run these jobs, being cognizant of token propagation.

We probably will only run the ray plugin tests (i.e. aws tests) for non forked PRs

jieru-hu:
closes #2181

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,9 @@ jobs:
             conda activate hydra
             pip install nox dataclasses --progress-bar off
             nox -s lint test_tools test_core test_jupyter_notebooks -ts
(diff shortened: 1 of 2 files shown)


pixelb (approved):
Looks good.
Downside to extra step rather than extra jobs is somewhat slower response from main jobs,
but it's much better having these tests run by default on forked PRs
thanks!

PR #2225, opened 2022-05-23
Update Optuna Sweeper

### LINKED ISSUE #2237: [Bug] Optuna sweeper raises uncaught exception when run ends with nan value
# 🐛 Bug
## Description
I perform hp search for a deep learning model and there are situations when model diverges and starts producing nans in it's output. My code exits gracefully returning tuple of nans as metrics. This is an expected behavior. But optuna sweeper doesn't think so.

1. The problem starts here: https://github.com/facebookresearch/hydra/blob/main/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py#L377 . Here values is a tuple of NaNs and state is COMPLETED (because the experiment code exited gracefully)
2. `study.tell` raises an error here: https://github.com/optuna/optuna/blob/release-v2.10.0/optuna/study/study.py#L652
 because of the check preformed here https://github.com/optuna/optuna/blob/release-v2.10.0/optuna/study/_optimize.py#L319
3. Error is finally reraised here: https://github.com/facebookresearch/hydra/blob/main/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py#L391 ,which ultimately crashes hydra.

## Checklist

## To reproduce

```
  File "/opt/conda/lib/python3.8/site-packages/hydra_plugins/hydra_optuna_sweeper/_impl.py", line 237, in sweep
    study.tell(trial=trial, state=state, values=values)
  File "/opt/conda/lib/python3.8/site-packages/optuna/study/study.py", line 652, in tell
    raise ValueError(values_conversion_failure_message)
ValueError: Trial 0 failed, because the objective function returned nan.

  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/utils.py", line 211, in run_and_report
    return func()
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/utils.py", line 386, in <lambda>
    lambda: hydra.multirun(
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/hydra.py", line 140, in multirun
    ret = sweeper.sweep(arguments=task_overrides)
  File "/opt/conda/lib/python3.8/site-packages/hydra_plugins/hydra_optuna_sweeper/optuna_sweeper.py", line 42, in sweep
    return self.sweeper.sweep(arguments)
  File "/opt/conda/lib/python3.8/site-packages/hydra_plugins/hydra_optuna_sweeper/_impl.py", line 240, in sweep
    study.tell(trial=trial, state=state, values=values)
  File "/opt/conda/lib/python3.8/site-packages/optuna/study/study.py", line 592, in tell
    raise ValueError(
ValueError: Values were told. Values cannot be specified when state is TrialState.PRUNED or TrialState.FAIL.
```

## Expected Behavior
If a code returns NaN, then mark trial as failed and proceed without crashing.

## System information
- **Hydra Version** :  1.2
- **Python version** : 3.8.12
- **Virtual environment type and version** : [NGC Docker container](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/pytorch)
- **Operating system** : GNU/Linux

zhengyu-yang:
## Motivation
According to [optuna's doc](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html#optuna.study.Study.tell), `state==FAIL, values!=None` will raise value error.

Yes

### DIFF
--- a/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
+++ b/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
@@ -387,7 +387,7 @@ class OptunaSweeperImpl(Sweeper):
 
                 except Exception as e:
                     state = optuna.trial.TrialState.FAIL
-                    study.tell(trial=trial, state=state, values=values)
+                    study.tell(trial=trial, state=state, values=None)
                     raise e
 
             n_trials_to_go -= batch_size


jieru-hu:
thanks for the contribution @zhengyu-yang 

could you provide a bit more context on what the PR is solving?

zhengyu-yang:
To reproduce my error, you can just change this [line](https://github.com/facebookresearch/hydra/blob/88abe128dbec2f03672dacadf7c584abd156657a/plugins/hydra_optuna_sweeper/example/sphere.py#L10) to `return float('nan')`. However, after more close inspection, I think this the error is mainly caused by optuna. Although optuna's [documentation](https://optuna.readthedocs.io/en/v2.10.0/faq.html?highlight=nan#how-are-nans-returned-by-trials-handled) says "Trials that return `NaN` (`float('nan')`) are treated as failures, but they will not abort studies," optuna still raises error and aborts (see [issue 1](https://github.com/optuna/optuna/issues/3132) and [issue 2](https://github.com/optuna/optuna/issues/3205)). 

However, I think the file change I suggest is still valid. Since according to [optuna's documentation](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html#optuna.study.Study.tell), after the sweeper plugin setting the `state` to `FAIL`, the `values` must be `None` to avoid error.

jieru-hu:
related to https://github.com/facebookresearch/hydra/issues/2237

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
```diff
@@ -387,7 +387,7 @@ def sweep(self, arguments: List[str]) -> None:
 
                 except Exception as e:
                     state = optuna.trial.TrialState.FAIL
-                    study.tell(trial=trial, state=state, values=values)
+                    study.tell(trial=trial, state=state, values=None)
```
jieru-hu, line 390: could we add a check here and only set `values=None` if "nan" exists in values? `study.tell` can fail in many other ways, and could potentially also caused by other bugs so i think it'd better to be more specific here.

jieru-hu (changes requested):
thanks for the contribution. pls see my comment. 
im also happy to take over this PR in case that works better for you.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
zhengyu-yang: Sure, I can add this

zhengyu-yang: Maybe a more specific way to handle this problem prior to `optuna 3.0` is to check `NaN` right after we cast every return values to float. If `NaN` is found, we can tell `optuna` this study is `FAIL` directly. However, I think even in that case, this change is still valid.

Jasha10:
> related to #2237

I think this PR will close #2237, in fact.

omry:
Thanks for the contribution. I am closing this PR as stale: it has been inactive for several years and the Optuna Sweeper work is now being collected under the broader modernization effort in #3210.

The underlying failure-handling concern is still useful context. If you are still interested, please join the discussion in #3210 so the fix can be considered as part of the current Optuna modernization path.

PR #2233, opened 2022-05-25
Add args and kwargs keys and arguments in raise

AntreasAntoniou:
Debugging issues without the given arguments and kwarguments is needlessly complex, I am adding such details to make it easier.

## Motivation

I have found that debugging issues regarding arguments and kwarguments without any clues as to what those are is needlessly hard. I have decided to include those in my error reporting messages, and I think the wider community would benefit from that as well. 

Yes

## Test Plan

Standard/existing tests involving implemented Exceptions should suffice.

### DIFF
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -68,8 +68,9 @@ def _call_target(
                 v._set_parent(None)
     except Exception as e:
         msg = (
-            f"Error in collecting args and kwargs for '{_convert_target_to_string(_target_)}':"
-            + f"\n{repr(e)}"
+            f"Error in collecting args and kwargs for '{_convert_target_to_string(_target_)}' \n"
+            f"with args: {args}, kwargs: {kwargs}. \n"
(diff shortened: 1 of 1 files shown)


jieru-hu:
thanks for the contribution @AntreasAntoniou  - it looks like you need to update the tests as well. 
here's out developer guide in case it helps https://hydra.cc/docs/development/testing/

jieru-hu (changes requested):
see comment. thanks!

omry:
Thanks for the contribution. I agree that better diagnostics for instantiate failures would be useful, but this specific approach needs maintainer buy-in and design discussion first.

Dumping `args` and `kwargs` directly into exception messages is risky: these values can be large configs or arbitrary runtime objects, and they may include secrets, generators, tensors, datasets, clients, or objects with expensive/broken `repr()` implementations. Error reporting here needs a deliberate policy around size, redaction, and safety rather than printing the raw values by default.

I am closing this PR as stale/incomplete. A future proposal for better instantiate diagnostics should start with an issue describing the intended behavior and safety constraints before implementation.

PR #2234, opened 2022-05-26
Update Structured Config docs

Jasha10:
Update docs r.e. support for `bytes`, `Path`, `Union`.


Jasha10:
> maybe also add more info/linked on strengthened nested container?
> 
> Also needs to update the 1.2 versioned doc :)

Done!

jieru-hu (approved):
looks great, thanks!

PR #2238, opened 2022-05-26
Mypy: use builtins.str to avoid variable shadowing

Jasha10:
The CI seems to be failing.


Jasha10:
TODO after merging: cherry-pick to `1.2_branch` (and maybe `1.1_branch`?)

PR #2242, opened 2022-05-26
fix colorlog default log path

### LINKED ISSUE #2241: [Bug] Colorlog plugin generates `.log` file in cwd instead of output dir
# 🐛 Bug
I'm using hydra v1.2 with `chdir` set to false.

When I don't use colorlog plugin, the `.log` file with python logs gets generated in my output directory.

But when I attach colorlog plugin with:
```yaml
defaults:
    - override hydra/hydra_logging: colorlog
    - override hydra/job_logging: colorlog
```
The `.log` file gets generated in current working directory

## Checklist

## Expected Behavior
I would expect the `.log` file to be always saved in output directory by default.

## System information
- **Hydra Version** :  1.2
- **Python version** : 3.10
- **Operating system** : linux

jieru-hu:
closes #2241

tested by running an hydra app locally with `hydra.job.chdir=False` and verified the file got created in the output dir as expected.

### DIFF
--- a/plugins/hydra_colorlog/hydra_plugins/hydra_colorlog/__init__.py
+++ b/plugins/hydra_colorlog/hydra_plugins/hydra_colorlog/__init__.py
@@ -1,3 +1,3 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
-__version__ = "1.2.0"
+__version__ = "1.2.1"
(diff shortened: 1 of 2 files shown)


Jasha10 (approved):
Excellent, thanks!

ashleve:
Hey @jieru-hu @Jasha10 

Has this fix been released?

I am still encountering issues with the incorrect positioning of the `.log` file when using colorlog, so I suspect this fix is pending a significant release from colorlog. 

However, it's been half a year now, so perhaps it would be worth it to make a minor release in the meantime?

hiyyg:
I am also having this bug, how to resolve this?

PR #2292, opened 2022-07-14
Fix ci: update pin on `ray`

Jasha10:


Jasha10:
Updating `1.12.0` -> `1.13.0` made the tests pass on my local machine, but CI still fails.
The error we're getting on CI after the updated pin is different from the one before.

Jasha10:
Closing this in favor of #2382

PR #2296, opened 2022-07-14
Copy entire `master_config` cache to `sweep_config`.

dominicgkerr:
Partially reverts the intention of #749 (to not copy cached `env` resolver values), as that pull request was made redundant by #1653 (on account of `env` being deprecated, and replaced by the non-caching `oc.env` resolver). Expanded tests to check behaviour of arbitrarily named resolvers (and not just the "special" `now` and `oc.env` resolvers) with/without value caching.

## Motivation

I wanted to create randomly (but memorably) named output directories (say, using the `coolname` python package), instead of using `${now:}` time-stamped directories. Unfortunately, as only the values cached by the `now` resolver are copied from the `master_config` to the `sweep_config` this is impossible (or at least very, very difficult?). Without copying the cache (and the accompanying "random directory name" generated by the `master_config`), every `sweep_config` job is written into a differently named output directory. 

I initially tried using a `hydra.experimental.callback.Callback` to pass "cached" values between the launcher and sweeper configs. I used `on_run_start`/`on_multirun_start` to record the output directory name generated by the launcher, but unfortunately the `on_job_start` method is called too late (that is, just after the sweep directory is created) for me to copy the callback's "cached" value...

This pull request reverts the behaviour of #749, which I argue was a little too strict - instead it (or this pull request) could have copied the entire `master_config` cache to the `sweep_config` EXCEPT for any `"env"` values. Since the `env` resolver has been deprecated (and replaced with the non-caching `oc.env` resolver), I decided not to introduce an `if k == "env":` loop/branch (to filter our `"env"` values), and instead simply copied across the entire cache.

Yes

## Test Plan

I've updated the test to additionally check the behaviour of arbitrarily named resolvers (and not just `hydra`'s/`omegaconf`'s "special" `now` and `oc.env` resolvers) with/without value caching. Arguably the `now` and `oc.env` present in `test_sweep_config_cache` already provide sufficient coverage by themselves though.

## Related Issues and PRs

Partially reverts #749. Motivated by #1653

### DIFF
--- a/hydra/_internal/config_loader_impl.py
+++ b/hydra/_internal/config_loader_impl.py
@@ -4,9 +4,8 @@ import os
 import re
 import sys
 import warnings
-from collections import defaultdict
 from textwrap import dedent
-from typing import Any, Dict, List, MutableSequence, Optional
+from typing import Any, List, MutableSequence, Optional
 
 from omegaconf import Container, DictConfig, OmegaConf, flag_override, open_dict
 from omegaconf.errors import (
--- a/hydra/_internal/config_loader_impl.py
+++ b/hydra/_internal/config_loader_impl.py
@@ -311,13 +311,10 @@ class ConfigLoaderImpl(ConfigLoader):
             run_mode=RunMode.RUN,
         )
 
-        # Partial copy of master config cache, to ensure we get the same resolved values for timestamps
-        cache: Dict[str, Any] = defaultdict(dict, {})
-        cache_master_config = OmegaConf.get_cache(master_config)
-        for k in ["now"]:
-            if k in cache_master_config:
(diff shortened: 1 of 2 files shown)


pixelb:
This looks really good, and ready to merge.

I was reluctant to avoid a `if k == "env"` check, but OmegaConf since 2.2 does indeed no longer support the caching ${env} resolver, since https://github.com/omry/omegaconf/commit/d64ebc5b3492b606241d3e7ea4528ffae8593b27 so we can assume that as we now depend on that version of OmegaConf

pixelb (approved):

dominicgkerr:
@pixelb Brilliant! Thanks (for reviewing & approving)

Issue #2318, opened 2022-07-27
[Bug] pickle5 requires python version < 3.8


lupus83:
# 🐛 Bug
## Description

pickle5 requires python version < 3.8, thus, is not required for more recent python version.

## Checklist
- [x ] I checked on the latest version of Hydra

Jasha10:
Closed by #2319

Issue #2079, opened 2022-03-09
[Bug] Installation of hydra_ray_launcher with Python 3.8 fails due to pickle5 error.


ncvv:
# 🐛 Bug
## Description
I am having trouble installing `hydra-ray-launcher` on my Windows 10 machine using Python 3.8 due to an error while building pickle5 which is part of the plugin installation requirements (see [here](https://github.com/facebookresearch/hydra/blob/796fdf1c54eabdcd5121d3a9c8ab6af70ee41763/plugins/hydra_ray_launcher/setup.py#L33)).

## Checklist

- [x]  I checked on the latest version of Hydra.
- [x]  I created a minimal repro.

## To reproduce

```console
PS C:\Users\ncvv\src> python --version
Python 3.8.6
PS C:\Users\ncvv\src> python -m venv C:/Users/ncvv/venvs/hydra-launcher
PS C:\Users\ncvv\src> & C:/Users/ncvv/venvs/hydra-launcher/Scripts/Activate.ps1
(hydra-launcher) PS C:\Users\ncvv\src> pip install hydra-core --upgrade
Collecting hydra-core
  Using cached hydra_core-1.1.1-py3-none-any.whl (145 kB)
Collecting omegaconf==2.1.*
  Using cached omegaconf-2.1.1-py3-none-any.whl (74 kB)
Collecting importlib-resources; python_version < "3.9"
  Using cached importlib_resources-5.4.0-py3-none-any.whl (28 kB)
Processing c:\users\ncvv\appdata\local\pip\cache\wheels\c8\d0\ab\d43c02eaddc5b9004db86950802442ad9a26f279c619e28da0\antlr4_python3_runtime-4.8-py3-none-any.whl 
  Using cached PyYAML-6.0-cp38-cp38-win_amd64.whl (155 kB)
Collecting zipp>=3.1.0; python_version < "3.10"
  Using cached zipp-3.7.0-py3-none-any.whl (5.3 kB)
Installing collected packages: antlr4-python3-runtime, PyYAML, omegaconf, zipp, importlib-resources, hydra-core
Successfully installed PyYAML-6.0 antlr4-python3-runtime-4.8 hydra-core-1.1.1 importlib-resources-5.4.0 omegaconf-2.1.1 zipp-3.7.0
(hydra-launcher) PS C:\Users\ncvv\src> pip install hydra-ray-launcher --upgrade
Collecting hydra-ray-launcher
[...]

PR #2319, opened 2022-07-27
fix hydra-ray-launcher dependency for python versions >= 3.8

lupus83:
## Motivation

pickle5 does not support python versions >= 3.8. Thus, the default pickle library should be used for those versions.

Yes

## Test Plan

Previous existing tests

## Related Issues and PRs

https://github.com/facebookresearch/hydra/issues/2318#issue-1319651414
https://github.com/facebookresearch/hydra/issues/2079#issue-1163741413

### DIFF
--- a/plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
+++ b/plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core_aws.py
@@ -8,9 +8,9 @@ from typing import Any, Dict, List, Sequence
 
 import cloudpickle  # type: ignore
 
-if sys.version_info < (3, 8):
+try:
     import pickle5 as pickle  # type: ignore
-else:
+except ModuleNotFoundError:
     import pickle
 from hydra.core.singleton import Singleton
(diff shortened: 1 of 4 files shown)


pixelb:
Right. I see this is needed since https://github.com/pitrou/pickle5-backport/commit/6128b82e1b

You're missing `import sys` in these checks.
Perhaps it would be better to:

```
try:
    import pickle5 as pickle
except ModuleNotFoundError:
    import pickle
```

lupus83:
Right! Thank you for pointing that out!

I've added the import of sys, which should ensure better maintainability in the future.

pixelb:
I'd have a slight preference for the ModuleNotFoundError way above (i.e. the "Better to ask for forgiveness" pattern).

BTW the python constraint is only on pickle5 >= 0.0.12, and we're specifying pickle5 == 0.0.11, so how are you hitting this?

Also given the above, as a follow up we should be able to remove the version constraint on pickle5

lupus83:
I don't mind either. Just implemented the try-catch-version.

Unfortunately, I don't remember how I encountered that error. Though it was quite time-consuming to resolve it, which is why we should merge it nevertheless.

pixelb (approved):

PR #2330, opened 2022-08-08
hydra_ax_sweeper support for when evaluation function returns a dict

pkendrick:
## Motivation

ax supports optimisation using multiple metrics. One use case is that we may have an experiment which optimises based on one metric (`experiment.objective_name`), but than we can define additional constraints and these can be based on other metrics (`experiments.outcome_constraints`).  Multiple metric outputs are supported by returning a dictionary from the evaluation function. Currently ax-sweeper does not support this as it does not allow a dict as the `raw_data` input to `ax_client.complete_trial()`:

see _core.py line 228:
```
                assert isinstance(val, (int, float, tuple))
```

Yes

## Test Plan

I have added a new test that tests that when the evaluation function returns a dict of metrics, and that the sweeper runs and the `experiment.objective_name` and `experiments.outcome_constraints` experimental parameters are interpreted by ax.

### DIFF
--- a/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
+++ b/plugins/hydra_ax_sweeper/hydra_plugins/hydra_ax_sweeper/_core.py
@@ -221,8 +221,11 @@ class CoreAxSweeper(Sweeper):
             for idx in range(len(batch)):
                 val: Any = rets[idx].return_value
                 # Ax expects a measured value (int or float), which can optionally
-                # be given in a tuple along with the error of that measurement
-                assert isinstance(val, (int, float, tuple))
+                # be given in a tuple along with the error of that measurement.
+                # Alternatively, the task function can return a dict whose values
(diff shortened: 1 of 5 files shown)


Jasha10:
Thanks @pkendrick. We'll get back to you on this shortly.

pkendrick:
Hi @Jasha10 just wondering if there was any update on this PR? Thanks!

Jasha10 (approved):
Thanks for your patience @pkendrick. This diff looks good to me.
I've pushed some minor changes, and will merge assuming that the ax-sweeper CI passes.

(I've opened PR #2374 to trigger the CI pipeline)

Jasha10:
I've just published a [dev release](https://pypi.org/project/hydra-ax-sweeper/1.3.0.dev0/) of the hydra-ax-sweeper to pypi.
```
pip install hydra-ax-sweeper==1.3.0.dev0
```

PR #2339, opened 2022-08-13
macOS CI: bump xcode 12.0.0 -> 13.4.1

Jasha10:
Attempting to fix CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,7 +119,7 @@ jobs:
       py_version:
         type: string
     macos:
-      xcode: "12.0.0"
+      xcode: "13.4.1"
     steps:
       - macos:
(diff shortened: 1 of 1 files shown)


Jasha10:
With this diff, the CI is still broken, but it is now less broken than before.
Here's an example of the failure from before this diff: 
> Build-agent version 1.0.135211-b712450b (2022-08-08T15:42:46+0000)
Creating a dedicated VM with xcode:12.0.0 image
failed to create host: Image xcode:12.0.0 is not supported

(From https://app.circleci.com/pipelines/github/facebookresearch/hydra/14733/workflows/e824e526-f13a-4203-849f-83401e2e2d90/jobs/146111)

Now at least the macos CI *starts* running successfully.

Issue #2038, opened 2022-02-17
[Feature Request] hydra with torch.distributed.launch


18445864529:
# 🚀 Feature Request

## Motivation

I want to use hydra with `torch.distributed.launch` for multi-node multi-GPU training. The problem is the `torch.distributed.launch` module will automatically pass a `local_rank` argument to the script thus leading to `unrecognized arguments: --local_rank`. 
In this case, how should I program my customized file accordingly to accept this appended argument?
Currently, it seems command arguments in the format of `--some_argument` will be directly received by hydra.

Jasha10:
The issue is that `@hydra.main` reads from `sys.argv`. Since `torch.distributed` is using `sys.argv` too, there is a conflict.
The most flexible solution is to use the [compose API](https://hydra.cc/docs/advanced/compose_api/).

Jasha10:
Thanks for the insight @jbaczek.

That is a good point about being unable to use a custom launcher without also using a custom sweeper + multirun mode. A feature request r.e. support for custom launching via single-run mode would be most welcome.

jbaczek:
I hit another wall with calling torch distributed as a launcher. Launcher is called in the hydra's main wrapper. I wanted to pass original function to the multiprocessing context. This pickles the function to call together with arguments (btw `OmegaConf.register_new_resolver` is unpicklable because it creates local function in runtime. And this is in a singleton state) and passes it to `Popen`. But here comes the issue: Pickle serializes only metadata about the function and then restores it from a code. But at this point hydra's main wrapper already has overwritten main function, so pickled function does not match function that it tries to restore. Here is a repro:
```
import functools                  
import pickle                     
                                  
def decorate(f):                  
    @functools.wraps(f)           
    def wrapper(*args, **kwargs): 
        print(pickle.dumps(f))    
        return f(*args, **kwargs) 
    return wrapper                
                                  
@decorate                         
def main():                       
    print('A')                    
                                  
if __name__ == '__main__':        
    main()                        
```
I'm still looking into the issue. Any hints on how to solve this are welcome.

longmao-yiran:
you can use --use_env to shield` arguments: --local_rank`

jbaczek:
@Jasha10 I have implemented a working plugin for this. Would you be interested in MR?

Jasha10:
@jbaczek I'd certainly be interested to take a look at the code you've written. I can't promise that we'd merge a PR (I'd have to discuss it with the Hydra team).

[...]

PR #2361, opened 2022-08-30
add task_function kwarg to on_job_start callback

Jasha10:
This PR closes feature request #2352 opened by @jgbos, adding a `task_function` keyword argument to the `on_job_start` callback.

TODO:
- [X] add test

### DIFF
--- a/hydra/_internal/callbacks.py
+++ b/hydra/_internal/callbacks.py
@@ -4,6 +4,8 @@ from typing import TYPE_CHECKING, Any, Optional
 
 from omegaconf import DictConfig, OmegaConf
 
+from hydra.types import TaskFunction
+
 if TYPE_CHECKING:
     from hydra.core.utils import JobReturn
 
@@ -41,8 +43,15 @@ class Callbacks:
             function_name="on_multirun_end", reverse=True, config=config, **kwargs
         )
 
-    def on_job_start(self, config: DictConfig, **kwargs: Any) -> None:
-        self._notify(function_name="on_job_start", config=config, **kwargs)
+    def on_job_start(
+        self, config: DictConfig, *, task_function: TaskFunction, **kwargs: Any
+    ) -> None:
+        self._notify(
+            function_name="on_job_start",
+            config=config,
+            task_function=task_function,
+            **kwargs,
+        )
 
     def on_job_end(
         self, config: DictConfig, job_return: "JobReturn", **kwargs: Any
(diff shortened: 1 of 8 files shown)


pixelb (approved):
This looks useful and should be backwards compat with older callers.
Approving assuming tests

Jasha10:
I've added a test to confirm that the updated callback API works as expected.
That being said, pickling the `task_function` doesn't work.
The problem is that the `@hydra.main` decorator wraps the `task_function` (à la `functools.wraps`), which makes pickling fail.
When I add the line `pickle.dumps(task_function)` to the `on_job_start` method in the test code, a `PicklingError` occurs.

Steps to reproduce:
1. check out this PR branch and install hydra:
```text
git clone -b possible-improvement-to-rerun https://github.com/Jasha10/hydra jasha10_hydra
cd jasha10_hydra
pip install -e .
```
2. modify the file `tests/test_apps/app_with_callbacks/on_job_start_accepts_task_function/my_app.py` so that the `on_job_start` method calls `pickle.dumps` on the `task_function`:
```python
$ cat tests/test_apps/app_with_callbacks/on_job_start_accepts_task_function/my_app.py
...

class OnJobStartCallback(Callback):
    def on_job_start(
        self, config: DictConfig, *, task_function: TaskFunction, **kwargs: Any
    ) -> None:
        import pickle
        pickle.dumps(task_function)

...
```
3. run the test app:
```text
$ python tests/test_apps/app_with_callbacks/on_job_start_accepts_task_function/my_app.py
/Users/jasha10/dev/foo/jasha10_hydra/hydra/_internal/callbacks.py:28: UserWarning: Callback OnJobStartCallback.on_job_start raised PicklingError: Can't pickle <function my_app at 0x10853eaf0>: it's not the same object as __main__.my_app
  warnings.warn(
```

This is the same problem mentioned in [this comment](https://github.com/facebookresearch/hydra/issues/2038#issuecomment-1063935678) on the `torch.distributed.launch` feature request.
@jgbos, what do you think?

jgbos:
@Jasha10 thanks for starting this PR.  This might be a bigger challenge.  Before I went on vacation last week I got a callback implementation running but I had to do two things:

- Use `cloudpickle` instead of `pickle`
- Make sure the task function is in a loadable package (i.e., not in `__main__`)

So I know this is possible, though I'm not sure how you should proceed with testing.  I know the `submitit` team successfully pickles task functions, I'll go dig through their code.

Jasha10:
> So I know this is possible

Got it. So in your assessment, will this PR still be useful in its current form?

jgbos:
> Got it. So in your assessment, will this PR still be useful in its current form?

Yep.  The challenge is on the user end I've successfully implemented this.

Issue #2362, opened 2022-09-02
[Bug] hydra-nevergrad-sweeper has a requirement on numpy<1.20.0


ajw1980:
In hydra_nevergrad_sweeper/setup.py:
```
"numpy<1.20.0",  # remove once nevergrad is upgraded to support numpy 1.20
```
I think nevergrad supports numpy > 1.20.0 so I think this requirement should be changed.
https://github.com/facebookresearch/nevergrad/blob/main/requirements/main.txt

Jasha10:
This issue is fixed by PR #2371.
I've published nevergrad sweeper version `1.3.0.dev0` to pypi, which should work with more recent numpy versions. You can install it with `pip install hydra-nevergrad-sweeper==1.3.0.dev0`.

PR #2371, opened 2022-09-08
nevergrad sweeper: unpin numpy

Jasha10:
This PR closes issue #2362 by removing the nevergrad sweeper's pin on `numpy`.

Three Commits:
- Bump hydra-nevergrad-sweeper dev version 1.2.0 -> 1.3.0.dev0
- nevergrad-sweeper: remove pins on numpy and cma
- nevergrad-sweeper: support python3.10

I'll confirm that the nevergrad CI passes before merging.

Todo after merge:
- [ ] publish dev release of `hydra-negergrad-sweeper` to Pypi


jloveric:
pypi install still needs to be updated with this change.

odelalleau:
> pypi install still needs to be updated with this change.

Indeed, right now you need to manually specify `hydra-nevergrad-sweeper==1.3.0.dev0` to have a working install, which is not obvious (otherwise by default you get a dependency conflict error)

PR #2374, opened 2022-09-09
ax sweeper: allow task function to return dict

Jasha10:
This PR is an exact copy of diff #2330. I've opened this PR so that the plugin CI tests will run.


Jasha10:
Closing as #2330 is now merged.

PR #2376, opened 2022-09-10
fix mypy CI

Jasha10:
This PR rectifies `mypy` CI issues:
- modify the noxfile:
  - remove the `--namespace-packages` flag from the noxfile call to `mypy`
  - don't use the `-e` flag to install plugins when running lints (i.e. use a non-editable install)
- add `py.typed` markers to Hydra plugin packages, making the packages compliant with [PEP 561](https://peps.python.org/pep-0561/).


Jasha10:
Closing in favor of PR #2377, which is based on a branch of the facebookresearch/hydra repo (rather than of the jasha10/hydra repo).

PR #2377, opened 2022-09-12
mypy CI: make plugins PEP-561-compliant

Jasha10:
This PR rectifies `mypy` CI issues:
- add `py.typed` markers to Hydra plugin packages, making the them [PEP 561](https://peps.python.org/pep-0561/) compliant.
- modify the noxfile:
  - remove the `--namespace-packages` flag from the noxfile call to `mypy`
  - don't use the `-e` flag to pip-install plugins when running lints (i.e. use a non-editable install)


pixelb:
Great stuff. Down to a ray specific issue now

pixelb (approved):

PR #2388, opened 2022-09-17
Implement Boxed-wrapper for instantiation

### LINKED ISSUE #2364: [Feature Request] Preserve dataclasses as instantiate-arguments
# 🚀 Feature Request

An additional option/syntax to pass dataclasses to instantiate-calls such that they're not converted to DictConfig/Dict instances.

## Motivation

Currently, when passing an instance of type `dataclass` as an argument to an instantiate call, the instance is converted into a OmegaConf/Dict object:

```python

@dataclass
class Bar:
    val: int

@dataclass
class Root:
    bar: Bar
    baz: int

cfg = OmegaConf.create(
    {
        "_target_": "__main__.Root",
        "baz": 123,
        "_convert_": "partial"
    }
)

bar = Bar(8)
instantiate(cfg, bar=bar)
```
Output:
```
Root(bar={'val': 8}, baz=123)
```

This is [expected](https://hydra.cc/docs/advanced/instantiate_objects/overview/) behaviour and makes sense when the dataclass represents configuration.

However, we're using dataclasses for most of our classes to reduce boilerplate code and get easy immutability. Therefore, it would be great if there was an option to preserve an already instantiated dataclass.

## Pitch

I would like an option/flag that allows the user to specify that certain dataclasses should not be converted in the instantiate-call so that the above script would output

```
Root(bar=Bar(val=8), baz=123)
```

Naturally, whatever the accepted solution, we should not break current behaviour. Hence, the do-not-convert flag has to be opt-in.

I am suggesting creating a `Boxed` class inheriting from [ValueNode](https://github.com/omry/omegaconf/blob/master/omegaconf/nodes.py#L21) that makes sure a dataclass passed into `instantiate` wrapped by `Boxed(...)` is preserved. Then, whenever the value is passed into a class constructor, hydra would just need to strip the `Boxes` class. This has two advantages IMO:
- the syntax is very clear and opt-in. The new instantiate call would look sth like this:
```
instantiate(cfg, bar=Boxed(bar))
```
- all changes could be done within hydra without having to change omegaconf.

Yes, if the maintainers think this solution fits well within hydra, I am happy to contribute a PR.

dsuess:
## Motivation

Motivation for this feature is outlined in #2364 

Yes

## Test Plan

Tests for this feature have been added as `test_instantiate_preserves_dataclasses`.

## Related Issues and PRs

Resolves #2364

### DIFF
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -24,6 +24,11 @@ class _Keys(str, Enum):
     PARTIAL = "_partial_"
 
 
+def Boxed(item):
+    # TODO Fill in
+    return item
+
+
 def _is_target(x: Any) -> bool:
     if isinstance(x, dict):
         return "_target_" in x
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -4,10 +4,11 @@ import copy
 import functools
 from enum import Enum
 from textwrap import dedent
-from typing import Any, Callable, Dict, List, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
 from omegaconf import OmegaConf, SCMode
 from omegaconf._utils import is_structured_config
+from omegaconf.base import Box, DictKeyType, Metadata, Node
 
 from hydra._internal.utils import _locate
 from hydra.errors import InstantiationException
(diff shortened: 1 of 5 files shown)


omry:
Addition to OmegaConf should probably done there. OmegaConf is not designed to allow new node types to be added externally. I suspect adding those would break various functions in OmegaConf.

dsuess:
Thanks @omry for your feedback. Can you please elaborate what I should look out for/test for when adding this to omegaconf? Would this be a viable feature if we move the `Boxed` implementation to omegaconf?

omry:
Discuss this with @Jasha10 who is maintaining OmegaConf.
I am not 100% sure that this belongs there, but I can see an argument for allowing arbitrary data objects in a boxed form (that unboxes automatically as you are accessing the config).

Jasha10:
Thanks for putting together this PR, @dsuess.

> Can you please elaborate what I should look out for/test for when adding this to omegaconf?

Here is the main question:
"How should `Boxed` interact with omegaconf primitives such as `DictConfig`, `ListConfig`, and `OmegaConf.create`?"


Here's an example that builds off your original post from #2364 to show what sort of behavior I'd look for on the OmegaConf side:
```python
bar = Bar(8)

# Below: it would be nice if `OmegaConf.create` could accept `Boxed` input:
cfg2 = OmegaConf.create(
    {
        "_target_": "__main__.Root",
        "baz": 123,
        "_convert_": "partial"
        "bar": Boxed(bar)  # proposed API
    }
)


# Below: I think DictConfig.__getattr__ should return
# the unboxed `Bar` instance, not the `Boxed` instance:
assert isinstance(cfg2.bar, Bar)
```
To fully integrate `Boxed` with omegaconf, we'd have to test that e.g. the `__getattr__` and `__setattr__` methods on `DictConfig` and `ListConfig` can handle boxed objects properly. If we're going to enable such interaction between `Boxed` and the other omegaconf primitives, then `Boxed` should definitely be part of omegaconf. A Hydra diff will still be required, of course, so that `hydra.utils.instantiate` can handle the `Boxed` node type.
```python
obj = hydra.utils.instantiate(cfg2)
assert isinstance(obj.bar, Bar)  # per #2364
```

pixelb:
I do think it would be useful in omegaconf. I slightly prefer the _convert_ interface to Boxed as it's a bit more explicit.

dsuess:
Closing due to comments. I'll see if we can put sth together inside omegaconf.

Issue #2368, opened 2022-09-07
[Feature Request] Option to specify the arg `--config-path, -cp` as an absolute path


c-salmi:
# 🚀 Feature Request

Option to specify the arg `--config-path, -cp` as an absolute path.

## Motivation

We are working on a new package for [local motion planning research](https://github.com/maxspahn/localPlannerBench). There is a lot of complex configuration required in this project and I think it's a perfect candidate to integrate hydra.

The problem is, it's a package. We would like potential users to be able to install the package (through pypi) and put their configuration files in an arbitrary location. However, currently the config_path arg is relative to the Python file declaring @hydra.main(). In our application @hydra.main() is called within the package, which makes relative paths tricky.

## Pitch

**Describe the solution you'd like**
Would it be possible to check if the specified path in `--config-path` is absolute or relative, using for instance `os.path.isabs(config_path)`? If it's an absolute path, we could simply set the search_path to be equal to the config_path. Otherwise we could keep the current behavior.

**Describe alternatives you've considered**
Alternatively, it could be done using an additional arg specifically for paths relative to the current working directory (or absolute paths).

**Are you willing to open a pull request?**
Yes

pixelb:
Yes this makes sense. Thanks

udaylunawat:
I am facing a similar scenario.

How does one pass the absolute path to config_path in the hydra decorator?

When using
`@hydra.main(config_path='/content/configs/', config_name="config", version_base=None)`

raises error
```
Primary config module 'package.directory.configs' not found.
Check that it's correct and contains an __init__.py file

Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.
```

The workaround that I am currently using is

pass an argument `--config-dir /content/configs`

and using the following decorator
```
@hydra.main(config_path='.', config_name="config", version_base=None)
```

Jasha10:
Hi @udaylunawat,
Support for absolute path as an argument to `@hydra.main` has not been implemented yet.
This is related to issue https://github.com/facebookresearch/hydra/issues/2274.

PR #2395, opened 2022-09-29
support absolute --config-path (#2368)

pixelb:
This was actually already supported as
os.path.join("/blah", "/abs/config"), just returns "/abs/config", but we've made this explicit in code and docs now.

### DIFF
--- a/tests/test_hydra.py
+++ b/tests/test_hydra.py
@@ -761,7 +761,7 @@ for details.
                     eval "$(python {script} -sc uninstall=bash)"
 
                 --config-path,-cp : Overrides the config_path specified in hydra.main().
-                                    The config_path is relative to the Python file declaring @hydra.main()
+                                    The config_path is absolute or relative to the Python file declaring @hydra.main()
                 --config-name,-cn : Overrides the config_name specified in hydra.main()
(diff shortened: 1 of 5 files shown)


tests/test_hydra.py
```diff
@@ -1033,8 +1033,15 @@ def test_override_with_invalid_group_choice(
...
-@mark.parametrize("config_name", ["cfg1", "cfg2"])
+@mark.parametrize(
+    "config_path",
+    [
+        "dir1",
+        "dir2",
+        os.path.abspath("tests/test_apps/app_with_multiple_config_dirs/dir2"),
+    ],
+)
+@mark.parametrize("config_name", ["cfg1", "cfg2", "cfg2"])
```
Jasha10, line 1044:
```suggestion
@mark.parametrize("config_name", ["cfg1", "cfg2"])
```
There's duplication here.

Jasha10 (approved):
The implementation looks good to me!
A few nits:
- news fragment
- could benefit from a test case in `tests/test_config_search_path.py:test_compute_search_path_dir`

Jasha10 (approved):

PR #2406, opened 2022-10-05
remove dead code from defaults_list.py

Jasha10:
`if d.is_options()` is always True, so delete the `else` block.

```python
            if isinstance(d, GroupDefault) and d.is_options():
                # d.is_options() is always True here
                if d.is_options():  # remove this line
                    ...  # dedent this block
                else:  # remove this line
                    ...  # delete this block
            ...
```

### DIFF
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -546,26 +546,22 @@ def _create_defaults_tree_impl(
 
             if isinstance(d, GroupDefault) and d.is_options():
                 # overriding may change from options to name
-                if d.is_options():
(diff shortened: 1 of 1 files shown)


pixelb (approved):
Noticed with a tool or just in passing?

Jasha10:
> Noticed with a tool or just in passing?

Just in passing.

Jasha10:
> with a tool

Come to think of it, we do have a [coverage session](https://github.com/facebookresearch/hydra/blob/76d63c60904f4ec5534ae2180cb78b06a575e4e7/noxfile.py#L479) defined in our noxfile, though it does not run as part of our CI.

PR #2418, opened 2022-10-15
support Python3.11

skshetry:
## Motivation

When trying to import hydra in Python 3.11, it throws following error:
```python
  File "/home/user/projects/iterative/hydra/hydra/conf/__init__.py", line 75, in JobConf
    @dataclass
     ^^^^^^^^^
  File "/home/user/.pyenv/versions/3.11-dev/lib/python3.11/dataclasses.py", line 1221, in dataclass
    return wrap(cls)
           ^^^^^^^^^
  File "/home/user/.pyenv/versions/3.11-dev/lib/python3.11/dataclasses.py", line 1211, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.pyenv/versions/3.11-dev/lib/python3.11/dataclasses.py", line 959, in _process_class
    cls_fields.append(_get_field(cls, name, type, kw_only))
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.pyenv/versions/3.11-dev/lib/python3.11/dataclasses.py", line 816, in _get_field
    raise ValueError(f'mutable default {type(f.default)} for field '
ValueError: mutable default <class 'hydra.conf.JobConf.JobConfig.OverrideDirname'> for field override_dirname is not allowed: use default_factory
```

This is due to a recent change in dataclasses in Python 3.11 that disallows mutable to be set as a default value. This is done via check for hashability.

See https://github.com/python/cpython/issues/88840. Note that I am trying to add support for Python 3.11 to hydra. This seems like a major blocker.

For reproducibility, I used the following command to search for these:
```console
rg ".*: .* = [A-Z].*\(.*\)"
```

Yes

## Test Plan

If all tests pass, it should be good.

### DIFF
--- a/tests/test_hydra.py
+++ b/tests/test_hydra.py
@@ -3,6 +3,7 @@ import os
 import re
 import subprocess
 import sys
+from dataclasses import dataclass
 from logging import getLogger
 from pathlib import Path
 from textwrap import dedent
@@ -1893,3 +1894,16 @@ def test_hydra_runtime_choice_1882(tmpdir: Path) -> None:
         from_name="Expected output",
         to_name="Actual output",
     )
+
+
+@mark.skipif(sys.version_info > (3, 11), reason="mutable defaults not allowed in 3.11")
+def test_dataclass_with_assigned_mutable_default():
+    @dataclass
+    class Y:
+        pass
+
+    @dataclass
+    class X:
+        y: Y = Y()
+
+    OmegaConf.create(X)
--- a/tests/test_hydra.py
+++ b/tests/test_hydra.py
@@ -1287,28 +1287,24 @@ def test_app_with_error_exception_sanitized(tmpdir: Any, monkeypatch: Any) -> No
         f"hydra.sweep.dir={tmpdir}",
         "hydra.job.chdir=True",
     ]
-    expected = dedent(
-        """\
-        Error executing job with overrides: []
-        Traceback (most recent call last):
-          File ".*my_app.py", line 13, in my_app
-            foo(cfg)
-          File ".*my_app.py", line 8, in foo
-            cfg.foo = "bar"  # does not exist in the config
-        omegaconf.errors.ConfigAttributeError: Key 'foo' is not in struct
+    expected_regex = dedent(
+        r"""
+        Error executing job with overrides: \[\]
+        Traceback \(most recent call last\):
+          File ".*my_app\.py", line 13, in my_app
+            foo\(cfg\)
+          File ".*my_app\.py", line 8, in foo
+            cfg\.foo = "bar"  # does not exist in the config
+        omegaconf\.errors\.ConfigAttributeError: Key 'foo' is not in struct
             full_key: foo
             object_type=dict
 
-        Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace."""
-    )
+        Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace\.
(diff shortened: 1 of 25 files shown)


Jasha10:
Hi @skshetry, thanks for working on Python3.11 support!

Using `default_factory` everywhere sounds like a good option.

The only other option I can think of is to use the decorator `@dataclass(frozen=True)` instead of `@dataclass`, which would make the dataclasses immutable. I think that option is sub-optimal because OmegaConf has [special support for frozen dataclasses](https://omegaconf.readthedocs.io/en/2.2_branch/structured_config.html?highlight=frozen#frozen-classes), using them to create read-only `DictConfig` instances.

We'll eventually need to make the same change for the OmegaConf codebase as for Hydra.

Jasha10:
It might be a good idea to create some regression tests like the following (to make sure the old dataclass pattern will still be supported with python < 3.11):

```python
if python_version < (3, 11):
    @dataclass
    class X:
        y: Y = Y()
```

skshetry:
> It might be a good idea to create some regression tests like the following (to make sure that OmegaConf and Hydra can still handle old dataclass pattern when python < 3.11):
> 
> ```python
> if python_version < (3, 11):
>     @dataclass
>     class X:
>         y: Y = Y()
> ```

I am hesitant to test that, as it is `dataclasses` behaviour that we’ll be testing. Instantiating via `default_factory` or default assigned value is its responsibility. And remember, this pattern is not going to change for <3.11.

Jasha10:
The point would be to make sure that a regression doesn't occur in OmegaConf/Hydra. We want to make sure that OmegaConf/Hydra can still handle dataclasses created using the old pattern.
In particular, I'd want to prevent regression in [`omegaconf._utils.get_dataclass_data`](https://github.com/omry/omegaconf/blob/03f69634afaa619fd931e99eeeceb63a8b3e5bba/omegaconf/_utils.py#L378) method, which processes a dataclass or dataclass instance.

```python
if python_version < (3, 11):
    @dataclass
    class X:
        y: Y = Y()
    
    # Will OmegaConf choke on dataclasses defined using
    # the old pattern? Currently this works:
    cfg = OmegaConf.create(X)
    # We should test it to make sure it keeps working (for python 3.6-3.10).
```

tests/test_hydra.py
```diff
@@ -1893,3 +1894,16 @@ def test_hydra_runtime_choice_1882(tmpdir: Path) -> None:
         from_name="Expected output",
         to_name="Actual output",
     )
+
+
+@mark.skipif(sys.version_info > (3, 11), reason="mutable defaults not allowed in 3.11")
+def test_dataclass_with_assigned_mutable_default():
```
skshetry, line 1900: I was not sure where to add this test. Let me know where it fits.

Jasha10:
Thanks @skshetry. I'm working on a python3.11 update for omegaconf. I'm planning to revisit this PR after the OmegaConf support is finished.

skshetry:
ping @Jasha10

Jasha10:
Thanks @skshetry

Jasha10:
FYI I've removed `test_dataclass_with_assigned_mutable_default` as it is redundant with the tests in https://github.com/omry/omegaconf/pull/1032.

PR #2499, opened 2022-12-10
plugin tests: use CPU version of pytorch to reduce CI download time

Jasha10:
Hydra's ax-sweeper plugin requires pytorch as a dependency.
Our CI sometimes times-out when the pytorch dep is downloading.
This PR modifies noxfile to install the CPU version of pytorch if `torch` is
required for a plugin test. This way, download time is minimized.

Commits:
- noxfile: install cpu version of torch
- noxfile plugin tests: print installed pytorch version

Example CI failure: https://app.circleci.com/pipelines/github/facebookresearch/hydra/15223/workflows/6fd6ccd7-2183-4659-b4dd-d2fc95e70adb/jobs/153959
See the `test_plugins_vs_core-3.9` nox session that ran during that CI job. The nox session failed while downloading `torch`.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -116,12 +116,32 @@ def install_selected_plugins(
 
 
 def install_plugin(session: Session, install_cmd: List[str], plugin: Plugin) -> None:
+    if plugin_requires_torch(plugin):
+        install_cpu_torch(session)
     cmd = install_cmd + [plugin.abspath]
     session.run(*cmd, silent=SILENT)
     if not SILENT:
         session.run("pipdeptree", "-p", plugin.name)
     # Test that we can import all installed plugins
     session.run("python", "-c", f"import {plugin.module}")
+
+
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -124,6 +125,28 @@ def install_plugin(session: Session, install_cmd: List[str], plugin: Plugin) ->
...
+def install_cpu_torch(session: Session) -> None:
+    """
+    Install the CPU version of pytorch.
+    This is a much smaller download size than the normal version `torch` package hosted on pypi.
+    The smaller download prevents our CI jobs from timing out.
+    """
+    session.install(
+        "torch", "--extra-index-url", "https://download.pytorch.org/whl/cpu"
+    )
+
```
omry, line 148: @Jasha10, why can't Ax just install the CPU version in tests?

PR #2500, opened 2022-12-10
[minor] more verbose about a job's error

min-xu-ai:
When a job raises an exception, it seems to be swallowed. After this change, the backtrace shows up in main.log as well as the job's stdout file. See below
image

### DIFF
--- a/hydra/core/utils.py
+++ b/hydra/core/utils.py
@@ -4,6 +4,7 @@ import logging
 import os
 import re
 import sys
+import traceback
 from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import datetime
@@ -186,6 +187,9 @@ def run_job(
                 ret.return_value = task_function(task_cfg)
                 ret.status = JobStatus.COMPLETED
             except Exception as e:
+                name = "HYDRA_FULL_ERROR"
+                if name in os.environ and os.environ[name] == "1":
+                    log.info(traceback.format_exc())
                 ret.return_value = e
                 ret.status = JobStatus.FAILED
 


Jasha10 (approved):
Thanks @min-xu-ai!

hydra/core/utils.py
```diff
@@ -186,6 +187,9 @@ def run_job(
                 ret.return_value = task_function(task_cfg)
                 ret.status = JobStatus.COMPLETED
             except Exception as e:
+                name = "HYDRA_FULL_ERROR"
+                if name in os.environ and os.environ[name] == "1":
+                    log.info(traceback.format_exc())
```
omry, line 192: better `log.error` here, no?

min-xu-ai: yeah, I agree. also, it is likely crash from here, so maybe that won't be missed in most cases.

omry:
If I am not mistaken, the exception is supposed to be printed downstream (especially if HYDRA_FULL_ERROR is set).

min-xu-ai:
But somehow it didn’t in my experiments

omry:
@Jasha10, did you confirm this before accepting the patch?
Exception handling in Hydra is not simple, this feels wrong.

Jasha10:
Yes. To confirm, I modified the file `examples/tutorials/basic/running_your_hydra_app/5_basic_sweep/my_app.py` so that the `my_app` function raises a `RuntimeError`. Then I ran a basic sweep:
`HYDRA_FULL_ERROR=1 python example/my_app.py -m`
This sweep contains 4 jobs.

Here's what I observed:
- Before this diff: none of the log files has a traceback (the logs are empty). On stdout, we get a traceback for the first job, which is printed after the sweep completes.
- After this diff: all four log files contain a traceback. On stdout, we get a traceback for each of the four jobs (printed immediately while the job is running). In addition, we get one more traceback (for the first job) after the sweep completes.

hydra/core/utils.py
Jasha10: I've created followup PR https://github.com/facebookresearch/hydra/pull/2549 changing the log level from `INFO` to `ERROR`.

omry:
A couple of things:
1. Hydra typically prints a sanitized stack trace unless HYDRA_FULL_ERROR is 1. This deviates from the pattern as far as I can tell.
2. By adding print, aren't you getting a duplicate printout on non-multirun executions?

omry:
@Jasha10, this is also missing a unit test that will ensure it's not breaking in the future.
I suggest you unland this until this and my questions are addressed.

Jasha10:
Unlanding in #2556.

PR #2183, opened 2022-04-28
Support Python 3.10

Jasha10:
To work around issue #2181, this PR is based on a branch of the hydra repo (not based on a fork).
Supersedes PR #2166.

PR #2531, opened 2023-01-03
Remove usages of legacy importlib API.

bwoodsend:
## Motivation

The Python >= 3.10 compatible usages of importlib added in 277ff54b71e31531923a51459b328391d76bb5b3 can be safely applied to any version of Python >= 3.5 which covers the full range that Hydra supports.

Getting rid of the old deprecated API, asides from cleaning up, will keep Hydra compatible with PyInstaller if PyInstaller doesn't get cold feet and revert its dropping of support for the legacy API (https://github.com/pyinstaller/pyinstaller/pull/7344).

Yes

## Test Plan

Standard CI/CD testing should be sufficient.

## Related Issues and PRs

None.

### DIFF
--- a/hydra/core/plugins.py
+++ b/hydra/core/plugins.py
@@ -182,20 +182,16 @@ class Plugins(metaclass=Singleton):
                     import_time = timer()
 
                     with warnings.catch_warnings(record=True) as recorded_warnings:
-                        if sys.version_info < (3, 10):
-                            m = importer.find_module(modname)  # type: ignore
-                            assert m is not None
-                            loaded_mod = m.load_module(modname)
+                        spec = importer.find_spec(modname, None)
+                        assert spec is not None
+                        if modname in sys.modules:
+                            loaded_mod = sys.modules[modname]
                         else:
-                            spec = importer.find_spec(modname)
-                            assert spec is not None
-                            if modname in sys.modules:
-                                loaded_mod = sys.modules[modname]
-                            else:
-                                loaded_mod = importlib.util.module_from_spec(spec)
-                            if loaded_mod is not None:
-                                spec.loader.exec_module(loaded_mod)
-                                sys.modules[modname] = loaded_mod
+                            loaded_mod = importlib.util.module_from_spec(spec)
+                        if loaded_mod is not None:
+                            sys.modules[modname] = loaded_mod
+                            assert spec.loader
+                            spec.loader.exec_module(loaded_mod)
 
                     import_time = timer() - import_time
                     if len(recorded_warnings) > 0:


bwoodsend:
Remaining failures look sufficiently irrelevant to call background noise.

bwoodsend:
@Jasha10 ?

Jasha10:
Thanks for the ping @bwoodsend! We should be able to merge this shortly.

omry:
Any concerns about compatibility with supported legacy Python versions?

bwoodsend:
No, because current hydra doesn't support Python 3.5.

hydra/core/plugins.py
```diff
@@ -184,20 +184,16 @@ def _scan_all_plugins(
...
-                            else:
-                                loaded_mod = importlib.util.module_from_spec(spec)
-                            if loaded_mod is not None:
-                                spec.loader.exec_module(loaded_mod)
-                                sys.modules[modname] = loaded_mod
+                            loaded_mod = importlib.util.module_from_spec(spec)
+                        if loaded_mod is not None:
+                            sys.modules[modname] = loaded_mod
+                            assert spec.loader
+                            spec.loader.exec_module(loaded_mod)
```
Jasha10, line 196: Shouldn't the `exec_module` line come before the `sys.modules[modname] = loaded_mod` line?
I'm worried that the current line order might leave `sys.modules` in an inconsistent state, with an improperly initialized module, if `exec_module` raises an exception.

Jasha10 (approved):
Thanks again @bwoodsend.
One relatively minor nit / question:

hydra/core/plugins.py
bwoodsend: It has to be that way around so that if the plugin has any kind of circular imports, the module is findable by its submodules. See the bottom of [this example](https://docs.python.org/3/library/importlib.html#approximating-importlib-import-module).

It does sound sensible for failed imports to be removed. The docs don't mention anything about removing them but a quick experimentation shows that `import` and `importlib.import_module()` adds the module at the start and then removes it again on failure.
```python
# broken_plugin.py
import sys
print(sys.modules["broken_plugin"])
assert 0
```
```python
>>> import broken_plugin
<module 'broken_plugin' from '/tmp/broken_plugin.py'>
  File "<stdin>", line 1, in <module>
  File "/tmp/broken_plugin.py", line 3, in <module>
    assert 0
AssertionError

>>> import sys
>>> "broken_plugin" in sys.modules
False
```
Same story if I swap `import` for `importlib.import_module()`.

I can add a `sys.modules.pop(modname, None)` to the `except` block if you prefer?

omry:
> No, because current hydra doesn't support Python 3.5.

This code is running on 3.6 to 3.9, wasn't it?
```python
                        if sys.version_info < (3, 10):
                            m = importer.find_module(modname)  # type: ignore
```

hydra/core/plugins.py
Jasha10: Hmm, I didn't think about submodules.
I think cleaning up broken submodules could be difficult. It's easy to do `pop(modname, None)`, but `pop(submodule_name, None)` is more challenging. Maybe it's better keep the code simple and let the user worry about cleaning up the broken plugins...

Jasha10:
> This code is running on 3.6 to 3.9, wasn't it?

Yes. I introduced that `if`-block in PR #2183 because the code was working for 3.6-3.9 but it did not work for python3.10.

omry:
> > This code is running on 3.6 to 3.9, wasn't it?
> 
> Yes. I introduced that `if < (3, 10)`-block in PR #2183 because the code was working for 3.6-3.9 but it did not work for python3.10.

Gotcha. I did seem unfamiliar :)

PR #2549, opened 2023-01-14
Use log.exception in `run_job` failure case

Jasha10:
This is a followup to PR #2500, which introduced logging for exceptions raised during a job.
This PR increases the log level from `INFO` to `ERROR`.

### DIFF
--- a/hydra/core/utils.py
+++ b/hydra/core/utils.py
@@ -4,7 +4,6 @@ import logging
 import os
 import re
 import sys
-import traceback
 from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import datetime
@@ -189,7 +188,7 @@ def run_job(
             except Exception as e:
                 name = "HYDRA_FULL_ERROR"
                 if name in os.environ and os.environ[name] == "1":
-                    log.info(traceback.format_exc())
+                    log.exception("Got error while running job")
                 ret.return_value = e
                 ret.status = JobStatus.FAILED
 


hydra/core/utils.py
```diff
@@ -189,7 +188,7 @@ def run_job(
             except Exception as e:
                 name = "HYDRA_FULL_ERROR"
                 if name in os.environ and os.environ[name] == "1":
-                    log.info(traceback.format_exc())
+                    log.exception("Got error while running job")
```
Jasha10, line 191: Using `log.exception` takes care of printing the traceback.

Jasha10:
Should we consider changing to log the exception in all cases, even if `HYDRA_FULL_ERROR` is not set? Currently, if `HYDRA_FULL_ERROR` is not set, multirun sweeps could have silent failure (see [my comment on #2500](https://github.com/facebookresearch/hydra/pull/2500#issuecomment-1382643619)).

min-xu-ai:
> Should we consider changing to log the exception in all cases, even if `HYDRA_FULL_ERROR` is not set? Currently, if `HYDRA_FULL_ERROR` is not set, multirun sweeps could have silent job failure (see [my comment on #2500](https://github.com/facebookresearch/hydra/pull/2500#issuecomment-1382643619)).

👍

omry:
Thanks for the follow-up PR. I confirmed that the underlying problem is still real: failed jobs in a basic multirun can leave empty per-job logs, even though Hydra eventually reports the first failure on the console.

This PR itself is stale now because it was a follow-up to #2500, and #2500 was reverted by #2556. The code this PR modifies no longer exists on current `main`, so this is not the right implementation path anymore.

I opened #3215 to track the remaining bug for Hydra 1.4. I am closing this PR in favor of that issue.

Issue #2564, opened 2023-01-31
[Bug] relative config_path resolution is not consistent


simpkins:
# 🐛 Bug
## Description
The config_path parameter to hydra.main() and hydra.experimental.initialize() must be a path name, or a relative module name.

If a relative module name is passed in, hydra does not behave consistently about what this module name is relative to.  If any of HYDRA_MAIN_MODULE, FB_PAR_MAIN_MODULE, or FB_XAR_MAIN_MODULE is set, the argument is treated relative to the value of that environment variable.  Otherwise hydra tries to determine the module that contains the task function, and the config_path is treated relative to that module.

This results in different behavior based on environment variables.  In our code base this has resulted in hundreds of locations where users are manually checking these environment variables before calling hydra.main(), and passing in different config paths based on the value of these environment variables.  (or, in a handful of cases, unsetting all of these variables before calling hydra.)  It seems better to simply fix hydra to behave consistently rather than having hundreds of call sites working around the hydra behavior.

Note that in our situation this mainly matters for FB_XAR_MAIN_MODULE and FB_PAR_MAIN_MODULE, since these environment variables are set conditionally depending on the current build mode, so hydra behaves differently in different build modes.

To minimize the impact of any code depending on the current behavior, it would probably be easiest to provide some way for callers to pass in a top-level module path, and prevent hydra from treating it as relative.  e.g., we could perhaps allow config_path to start with `pkg://` to indicate that what follows is a top-level module name.

#2395 added support for absolute paths, but there is no mechanism to pass in non-relative modules that I can tell.

## Checklist
- [X] I checked on the latest version of Hydra
- [X] I created a minimal repro (See [this](https://stackoverflow.com/help/minimal-reproducible-example) for tips).

## Expected Behavior
Provide a consistent behavior for config_path that does not depend on build mode.

## System information
- **Hydra Version** :  1.3.1
- **Python version** : 3.8.6

omry:
The automatic detection of the main module is a tricky part of Hydra. 
It was not possible in fbcode for some reason (at least, I was not able to support it in fbcode in the same way). This led me to add support for FB_PAR_MAIN_MODULE and FB_XAR_MAIN_MODULE (HYDRA_MAIN_MODULE is there for a non-fbcode specific solution, but it's not documented, and I don't believe is being used anywhere outside of tests).

Off the top of my head, those environment variables are short-circuiting the detection of the main module and are always treated as absolute modules.

[...]

PR #2565, opened 2023-01-31
Allow specifying a non-relative module name in config_path

simpkins:
## Motivation

When constructing the config search path, allow callers to directly pass in a non-relative module name starting with `pkg://`.  This allows callers to bypass the relative name lookup, and always get consistent behavior regardless of the current environment variables at runtime.

Yes

## Test Plan

Added a new entry in test_config_search_path.py

## Related Issues and PRs

#2564

### DIFF
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -123,9 +123,13 @@ def compute_search_path_dir(
     calling_module: Optional[str],
     config_path: Optional[str],
 ) -> Optional[str]:
-    if config_path is not None and os.path.isabs(config_path):
-        search_path_dir = config_path
-    elif calling_file is not None:
+    if config_path is not None:
+        if os.path.isabs(config_path):
+            return config_path
+        if config_path.startswith("pkg://"):
+            return config_path
+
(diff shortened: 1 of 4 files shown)


omry:
Hi Adam.
Can you add a test covering this? I believe there should be some test testing this directly already.

omry:
Nice, thanks.
I will let @Jasha10 accept and merge. I don't know what is the status of the CI.

simpkins:
The CI failures look like they are because the Netlify config is using an old Ubuntu 16.04 image, which is no longer supported.  I'll ping some other folks about trying to get that config fixed.

omry:
Can you ping the open source team (OSS support IIRC) about it?
I believe the website preview functionality is managed by them.
If they can't fix it I think we should probably just disable the website preview from CI.

simpkins:
The CI test failures look like pre-existing problems occurring on main.  I sent out #2576 to auto-apply some of the black formatting fixes that are currently causing the tests to fail.

Jasha10 (approved):
Thanks very much @simpkins!

PR #2577, opened 2023-02-07
Obey `_recursive_` parameter on non-target nodes

michaelhball:
## Motivation

When recursively instantiating a config, the _recursive_ parameter is ignored unless the config is a target node (i.e. has a `_target_` parameter). It would be nice if the `_recursive_` flag were considered for non-target nodes as well so that a recursive flow of instantiation can be "broken" (i.e. all nodes from a certain level in the config "tree" remain as `DictConfig` objects). The reason why there is an upstream `_recursive_=True` is because the parent node has other children who we do want to recursively instantiate. As a concrete example, consider the following set of configs and corresponding classes to be instantiated (I tried to keep them as minimal as possible)

```python
@dataclass
class MyArchitecture:
    _target_: str = ".../Architecture"


@dataclass
class MyLRScheduler:
    _target_: str = ".../LRScheduler"


@dataclass
class MyOptimizer:
    _recursive_: bool = False
    scheduler: MyLRScheduler = field(default_factory=MyLRScheduler)


@dataclass
class MyNN:
    _target_: str = ".../NN"
    _recursive_: bool = True

    architecture: MyArchitecture = field(default_factory=MyArchitecture)
    optimizer: MyOptimizer = field(default_factory=MyOptimizer)


@dataclass
class MyConfig:
    nn: MyNN = field(default_factory=MyNN)


cs = ConfigStore.instance()
cs.store(name="config", node=MyConfig)


class LRScheduler:
    ...


class Architecture:
    ...


class NN:
    def __init__(self, architecture: Architecture, optimizer: MyOptimizer) -> None:
        # NB: the `optimizer` expected here is a DictConfig (duck-typed as MyOptimizer)
        self.architecture = architecture
        self.optimizer = optimizer


@hydra.main(config_name="config")
def my_app(cfg: MyConfig) -> None:
    instantiated_config: NN = instantiate(cfg.nn)
    print(f"Scheduler was instantiated: {not isinstance(instantiated_config.optimizer.scheduler, DictConfig)}")

if __name__ == "__main__":
    my_app()
```

The output of the print statement above is `"Scheduler was instantiated: True"`, while the idea behind this PR is that it should be False (due to the setting of the `_recursive_=False` parameter in the `MyOptimizer` config).

Yes

## Test Plan

Added an additional test in `tests/instantiate/test_instantiate` and verified that no existing tests are broken. As far as I can tell, none of the failing CI tests are due to this PR.

## Related Issues and PRs

### DIFF
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -354,18 +354,24 @@ def instantiate_node(
             ):
                 dict_items = {}
                 for key, value in node.items():
-                    # list items inherits recursive flag from the containing dict.
-                    dict_items[key] = instantiate_node(
-                        value, convert=convert, recursive=recursive
-                    )
+                    if recursive:
+                        # list items inherits recursive flag from the containing dict.
+                        dict_items[key] = instantiate_node(
+                            value, convert=convert, recursive=recursive
+                        )
+                    else:
+                        dict_items[key] = value
                 return dict_items
             else:
                 # Otherwise use DictConfig and resolve interpolations lazily.
                 cfg = OmegaConf.create({}, flags={"allow_objects": True})
                 for key, value in node.items():
-                    cfg[key] = instantiate_node(
-                        value, convert=convert, recursive=recursive
-                    )
+                    if recursive:
+                        cfg[key] = instantiate_node(
+                            value, convert=convert, recursive=recursive
+                        )
+                    else:
+                        cfg[key] = value
                 cfg._set_parent(node)
                 cfg._metadata.object_type = node._metadata.object_type
                 if convert == ConvertMode.OBJECT:
--- a/tests/instantiate/__init__.py
+++ b/tests/instantiate/__init__.py
@@ -420,6 +420,11 @@ class NestedConf:
     b: Any = field(default_factory=lambda: User(name="b", age=2))
 
 
+@dataclass
+class NestedConfNoTarget:
+    a: Any = field(default_factory=lambda: SimpleClassDefaultPrimitiveConf)
+
+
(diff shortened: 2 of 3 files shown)


jdrudolph:
@michaelhball I've found a workaround for this issue:
```python
@dataclass
class MyOptimizer:
    _target_: str = "MyOptimizer"  # self-referential target to break recursive instantiation
    _recursive_: bool = False
    scheduler: MyLRScheduler = field(default_factory=MyLRScheduler)
```

would love to see this PR merged though!

Zacharias030:
Could we please have this PR considered? Tagging people with a release bit for help @omry @Jasha10 @jieru-hu @pixelb!

omry:
Thanks for the PR. I agree that the request is reasonable: non-target config nodes are currently transparent to recursive instantiation, and `_recursive_: false` on such a node should plausibly act as a stop for that subtree, matching the behavior people already expect from target nodes.

This PR is not ready to merge as-is. Please update it with:

- A focused regression test for the motivating shape: target node -> non-target child with `_recursive_: false` -> target grandchild. The assertion should show that the child remains a `DictConfig` and the grandchild is not instantiated.
- Documentation for the `_recursive_` behavior in `instantiate()`, including the distinction between target nodes and non-target container nodes.
- A news fragment for the user-visible behavior change.
- A rebase/update against current `main`.

If these fixes arrive in time and the final behavior looks clean, I can consider this for Hydra 1.4. Otherwise I will defer it to a future release.

omry:
Thanks for the original contribution and for identifying this gap. The requested behavior is now present on `main` through #3302, and #3310 adds explicit regression coverage for the motivating target → non-target → target shape plus the corresponding 1.4 feature news fragment. Closing this PR as superseded.

PR #2677, opened 2023-06-08
Fix some tests

shagunsodhani:
## Motivation

Fixes #2676 

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebookresearch/hydra/blob/master/CONTRIBUTING.md)?

Yes

## Test Plan

CI

## Related Issues and PRs

#2676

PR #2689, opened 2023-06-20
Use syntax updated with Python 3.7 and avoid unnecessary list creations

bryant1410:
I ran the following, which updates the code to syntax to Python >= 3.7:

```bash
pyupgrade $(find . -name "*.py" -type f) --py37-plus
```

This includes, for example, the Python-2 style of calling `super` and using f-strings where possible. I reviewed these changes manually.

Also, I looked for unnecessary list comprehension creations, where a generator expression could have been created instead, thus saving memory and compute (though small if the lists were small).

### DIFF
--- a/hydra/core/config_store.py
+++ b/hydra/core/config_store.py
@@ -133,10 +133,10 @@ class ConfigStore(metaclass=Singleton):
     def list(self, path: str) -> List[str]:
         d = self._open(path)
         if d is None:
-            raise IOError(f"Path not found {path}")
+            raise OSError(f"Path not found {path}")
 
         if not isinstance(d, dict):
-            raise IOError(f"Path points to a file : {path}")
+            raise OSError(f"Path points to a file : {path}")
 
         return sorted(d.keys())
 
--- a/hydra/test_utils/test_utils.py
+++ b/hydra/test_utils/test_utils.py
@@ -221,7 +221,7 @@ def find_parent_dir_containing(
         cur = os.path.relpath(os.path.join(cur, ".."))
         max_up = max_up - 1
     if max_up == 0:
-        raise IOError(f"Could not find {target} in parents of {os.getcwd()}")
+        raise OSError(f"Could not find {target} in parents of {os.getcwd()}")
     return cur
 
 
@@ -333,7 +333,7 @@ if __name__ == "__main__":
                 modified_env.update(env_override)
         subprocess.check_call(cmd, env=modified_env)
 
-        with open(output_file, "r") as f:
+        with open(output_file) as f:
             file_str = f.read()
             output = str.splitlines(file_str)
 
--- a/tests/instantiate/__init__.py
+++ b/tests/instantiate/__init__.py
@@ -39,7 +39,7 @@ def partial_equal(obj1: Any, obj2: Any) -> bool:
     if isinstance(obj1, list):
         if len(obj1) != len(obj2):
             return False
-        return all(partial_equal(o1, o2) for o1, o2 in zip(obj1))
+        return all(partial_equal(o1, o2) for o1, o2 in zip(obj1, obj2))
     if not (isinstance(obj1, partial) and isinstance(obj2, partial)):
         return False
     return all(
--- a/tests/instantiate/__init__.py
+++ b/tests/instantiate/__init__.py
@@ -39,7 +39,7 @@ def partial_equal(obj1: Any, obj2: Any) -> bool:
     if isinstance(obj1, list):
         if len(obj1) != len(obj2):
             return False
-        return all([partial_equal(obj1[i], obj2[i]) for i in range(len(obj1))])
+        return all(partial_equal(o1, o2) for o1, o2 in zip(obj1))
     if not (isinstance(obj1, partial) and isinstance(obj2, partial)):
         return False
     return all(
--- a/tools/configen/configen/utils.py
+++ b/tools/configen/configen/utils.py
@@ -59,7 +59,7 @@ def type_str(t: Any) -> str:
 
 def is_tuple_annotation(type_: Any) -> bool:
     origin = getattr(type_, "__origin__", None)
-    return origin is tuple  # pragma: no cover
+    return origin is tuple
 
 
 def convert_imports(imports: Set[Any], string_imports: Iterable[str]) -> List[str]:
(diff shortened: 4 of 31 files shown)


hydra/core/config_store.py
```diff
@@ -133,10 +133,10 @@ def get_type(self, path: str) -> ObjectType:
     def list(self, path: str) -> List[str]:
         d = self._open(path)
         if d is None:
-            raise IOError(f"Path not found {path}")
+            raise OSError(f"Path not found {path}")
 
         if not isinstance(d, dict):
-            raise IOError(f"Path points to a file : {path}")
+            raise OSError(f"Path points to a file : {path}")
```
Jasha10, line 139: Why is this change necessary?

hydra/test_utils/test_utils.py
```diff
@@ -221,7 +221,7 @@ def find_parent_dir_containing(
         cur = os.path.relpath(os.path.join(cur, ".."))
         max_up = max_up - 1
     if max_up == 0:
-        raise IOError(f"Could not find {target} in parents of {os.getcwd()}")
+        raise OSError(f"Could not find {target} in parents of {os.getcwd()}")
```
Jasha10, line 224: Same question.

tools/configen/configen/utils.py
```diff
@@ -59,10 +59,7 @@ def type_str(t: Any) -> str:
 
 def is_tuple_annotation(type_: Any) -> bool:
     origin = getattr(type_, "__origin__", None)
-    if sys.version_info < (3, 7, 0):
-        return origin is Tuple or type_ is Tuple  # pragma: no cover
-    else:
-        return origin is tuple  # pragma: no cover
+    return origin is tuple  # pragma: no cover
```
Jasha10, line 62: I think `# pragma: no cover` comment here can be removed.
The comment was only there to prevent our coverage linter from complaining when the line was not hit by the python3.6 test suite.

tools/configen/tests/test_modules/future_annotations.py
```diff
@@ -26,12 +26,12 @@ class ExampleClass:
...
-        passthrough_list: List[LibraryClass],
-        dataclass_val: List[User],
-        def_value: List[str] = [],
+        lst: list[str],
+        passthrough_list: list[LibraryClass],
+        dataclass_val: list[User],
+        def_value: list[str] = [],
         default_str="Bond, James Bond",
-        none_str: Optional[str] = None,
+        none_str: str | None = None,
```
Jasha10, line 34: This change is not compatible with python3.7.

Python3.7 doesn't support subscripting the builtin `list`, nor does it support using the bar `|` for union types.

Jasha10 (changes requested):
Thanks @bryant1410.
Overall it looks good, with a few issues:

tools/configen/tests/test_modules/future_annotations.py
bryant1410: It does for typing (but e.g. not type aliases) when using `from __future__ import annotations` as at the beggining of this file.

bryant1410: This is because the usage of that annotation makes the types be treated as strings and they are evaluated later.

hydra/core/config_store.py
bryant1410: It's not strictly necessary because they are semantically the same, but it's the recommended canonical name now. See https://stackoverflow.com/a/29347946/1165181

bryant1410:
BTW, the tests are failing for another reason, right?

Jasha10:
> BTW, the tests are failing for another reason, right?

I think so. The failing test should be fixed by PR #2677, which was recently merged into `main`. Could you please rebase this PR against the `main` branch?

tools/configen/tests/test_modules/future_annotations.py
Jasha10: I wonder, can we get away with removing the `from __future__ import annotations` line in this file?

hydra/core/config_store.py
Jasha10:
```python
>>> IOError is OSError
True
```
I see. Thanks.

bryant1410:
> I think so. The failing test should be fixed by PR #2677, which was recently merged into `main`. Could you please rebase this PR against the `main` branch?

Okay! I just merged instead of rebasing, hope it's fine. You can always rebase everything at the end, and I think it should remove the merge-type commits.

tools/configen/tests/test_modules/future_annotations.py
bryant1410: It's called `future_annotations`, so sounds like no? 😄

bryant1410:
BTW, I see the latest in `main` is still failing. You sure the tests are gonna work?

tests/instantiate/__init__.py
```diff
@@ -39,7 +39,7 @@ def partial_equal(obj1: Any, obj2: Any) -> bool:
     if isinstance(obj1, list):
         if len(obj1) != len(obj2):
             return False
-        return all([partial_equal(obj1[i], obj2[i]) for i in range(len(obj1))])
+        return all(partial_equal(o1, o2) for o1, o2 in zip(obj1))
```
Jasha10, line 42: There's an error in this update to the list comprehension.
Should be something like this:
```python
        return all(partial_equal(o1, o2) for o1, o2 in zip(obj1, obj2))
```

Jasha10 (changes requested):
Hmm, no. It seems there's an issue with the diff. See comment below:

tools/configen/tests/test_modules/future_annotations.py
Jasha10: Oh, great point.

tests/instantiate/__init__.py
bryant1410: Oh, thank you! For context, I did that one manually, not with the tool.

tools/configen/tests/test_modules/future_annotations.py
Jasha10: To keep consistency with the rest of the codebase, I still think there's benefit to using `List` instead of `list`. We don't need to convert `List -> list` just because the pyupgrade tool said so.

Jasha10:
Currently three CI jobs are failing on this branch:
* lint-3.10: seems unrelated to this diff
* test_tools-3.10: related
* test_core-3.10: seems unrelated to this diff (see #2692)

I believe the test_tools failure is caused by this diff's change to `tools/configen/tests/test_modules/future_annotations.py`.

Jasha10 (approved):
Looks good. Thanks @bryant1410!

shagunsodhani (approved):

PR #2714, opened 2023-07-14
Rename references of `master` into `main`

bryant1410:
I've checked all links to make sure the change makes sense and that they work fine.


shagunsodhani (approved):
Thank you!

PR #2715, opened 2023-07-14
Update to use `actions/checkout@v3` in `deploy_website.yml`

bryant1410:


shagunsodhani:
Could you explain the difference between the 2 options @bryant1410 ?

bryant1410:
I saw a warning in the logs here that v2 uses a version of node that's gonna be deprecated or something

PR #2733, opened 2023-08-09
Upgrade `antlr` to 4.11

jlopezpena:
## Motivation

PR following up on [this discussion](https://github.com/facebookresearch/hydra/issues/2699) regarding updating `antlr4` binary and runtime to a more recent version.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebookresearch/hydra/blob/main/CONTRIBUTING.md)?

Yes

## Test Plan

Usual tests should suffice, if there is anything broken with the new `antlr` it should fail at processing the grammars at build time

## Related Issues and PRs

In order to merge this PR a new version of `omegaconf` that includes [the corresponding update there](https://github.com/omry/omegaconf/pull/1114) should be released

PR #2861, opened 2024-02-29
win CI upgrade jre

Jasha10:


Jasha10:
CI seems to be passing on windows, so I'll incorporate this change into PR #2733.

Issue #2664, opened 2023-05-16
[Not sure if bug/feature] Where does submitit error output go?


Ubadub:
Hi,

I am using the submitit plugin to run a `--multirun` sweep. Some of my jobs errored- based on on the logging messages my code is printing to the log file in the sweep subdir, I can tell they are exiting prematurely; however, at the bottom of the log file, submitit nevertheless claims that the `Job completed successfully`.

I'm opening this issue to ask where I can find the actual stacktrace, and also why the log message erroneously claims the job completed successfully.

I do see two files: `.submitit/[JOBNAME_JOBNUMBER]/[JOBNAME_JOBNUMER]_log.out` and `.submitit/[JOBNAME_JOBNUMBER]/[JOBNAME_JOBNUMER]_log.err`. The former appears to be identical to the main log file that gets placed in the sweep subdir after the run completes. The latter, confusingly, does appear to contain some error messages, but only ones produced by third party libraries (in particular, I am using the HuggingFace `transformers` library, which prints certain warning messages that are appearing in the stderr file). I am used to seeing this output when running a job directly in the console, but it's unclear to me why that output is appearing there but not anything else that I would normally expect to see in stderr. My working hypothesis is that that the error file contains messages that are printed directly to `sys.stderr`, but doesn't contain error messages that are the result of raised exceptions.

When all the runs complete, I do see a small, relatively unhelpful error message that usually just consists of the error message only, without a stack trace or a line number. And even if many jobs fail, only one error message appears to be produced.

Is this customizable behavior? Is there some file or setting I'm missing? How can I recover a full, useable stack trace, like what I would have seen had I run the command without `--multirun` and without the submitit launcher?

odelalleau:
Yeah, it's an annoying behavior with the submitit launcher. Here's the recipe I'm using myself:
```python
import traceback

@hydra.main(...)
def main(cfg: DictConfig):
    try:
        run()
    except Exception:
        traceback.print_exc(file=sys.stderr)
        raise
```
==> you'll now see errors in the stderr log

Ubadub:
> Yeah, it's an annoying behavior with the submitit launcher. Here's the recipe I'm using myself:
> 
> ```python
> import traceback
> 
> @hydra.main(...)
> def main(cfg: DictConfig):
>     try:
>         run(cfg)
>     except Exception:
>         traceback.print_exc(file=sys.stderr)
>         raise
> ```
> 
> ==> you'll now see errors in the stderr log

[...]

PR #2863, opened 2024-03-01
Embed original traceback in Hydra's exception report.

gleize:
## Motivation

When Hydra runs jobs (multirun mode), and the job crashes, *only* the exception gets reported and raised (but not the original traceback). This is missing for proper / faster debugging of the original error.
This PR solves that missing feature.

No

## Test Plan

Tested with the submitit plugin, which pickles the exception information properly.


kirillbobyrev:
Hi! Is it possible to merge this so that SAM 3D Objects setup is slightly less convoluted?

omry (changes requested):
Thanks for working on this. This PR is addressing the problem tracked in https://github.com/facebookresearch/hydra/issues/2664, but it needs more work before it can be merged.

The core issue is real: failed remote/multirun jobs can lose the useful original traceback after the `JobReturn` crosses a serialization boundary. However, this implementation is not complete enough as-is:

- Please link this PR explicitly to #2664.
- Preserve the existing `JobReturn.return_value` behavior as much as possible. Replacing the stored exception with a `TracebackException` and then raising `JobException` changes the exception type and can break callers that expect the original exception object/type.
- The traceback representation needs to survive the serialization boundaries used by launchers. In particular, a stdlib `pickle` round-trip of `JobReturn` containing `traceback.TracebackException(*sys.exc_info())` currently fails with `TypeError: cannot pickle code objects`.
- Add tests that reproduce the lost traceback after serialization and verify the fixed behavior. The tests should cover the relevant launcher boundary, not only the local in-process path.
- Add a news fragment for the user-visible behavior change.

If this does not get completed before the upcoming 1.4 release, I will handle this fix separately.

omry:
Thanks for working on this, and apologies that the PR has remained open for so long.

The underlying problem is real and remains tracked in #2664. However, this implementation changes the exception type exposed through `JobReturn`, and the stored `TracebackException` does not reliably survive the serialization boundary used by remote launchers. The branch also now conflicts with `main`, so this would require a new implementation rather than a straightforward update.

I'm closing this PR as not planned rather than asking you to rework a two-year-old branch. We will continue tracking the fix in #2664 and address it separately with serialization and launcher coverage.

Thanks again—this PR helped clarify the failure mode and the requirements for a complete fix.

PR #2880, opened 2024-04-02
Adding the argument `target_wrapper` to `hydra.utils.instantiate` to support recursive type checking

JulesGM:
## Motivation

[beartype](https://github.com/beartype/beartype) and other libraries offer runtime typechecking with function and class decorators, like so:

```python
from beartype import beartype

@beartype
@dataclass
class ExampleDataclass:
    name: str
    unit_price: float
    quantity_on_hand: int = 0

inst = ExampleDataclass("a name", 123.32, 23)
# Is ok

inst = ExampleDataclass(123, None, 23)
# Breaks

```

In this PR, we suggest adding an argument to `hydra.utils.instantiate` that is called `target_wrapper`, that receives a callable as an argument. The idea is that callable is then called on `_target_` before it itself is called. This allows to decorate all of a configuration's `_target_`s all at once. (It's possible that the argument should be called `target_decorator`)

One of the ways this can be used is with something like `beartype`, where runtime type-checking to the portion of the code being instantiated with `_target_` where annotations are present, with a single line of non-invasive code, which would be  really incredibly helpful for us.

It would look like this:

```python
hydra.utils.instantiate(cfg, target_wrapper=beartype.beartype)
```

See the `test` example lower in the post for a more full example.

One could imagine other types of wrappers that people could want to do, for profiling for example, where folks could wrap every class type with a profiler before the instantiation happens.

One could also add recursive support for `OmegaConf.structured` with the following, which is very cool:

```python
import dataclasses
import attrs


def apply_structured(_target_):
    if dataclasses.is_dataclass(_target_) or attrs.has(_target_):
        def factory(*args, **kwargs):
            return OmegaConf.structured(_target_(*args, **kwargs))
   
    return _target_
```

and then 

```python
hydra.utils.instantiate(cfg, target_wrapper=apply_structured)
```

Yes

## Test Plan

Here is a standalone test with `beartype`:

```python
from pathlib import Path
import builtins
import beartype
import dataclasses
import hydra
import tempfile

CHECKER = beartype.beartype

@dataclasses.dataclass
class DataclassExample:
    argument_integer_example: int


config_example_good = """
_target_: test.DataclassExample
argument_integer_example: 123
"""
config_name_good = "config_name_good"


config_example_error = """
_target_: test.DataclassExample
argument_integer_example: 123aa
"""
config_name_error = "config_name_bad"


def doer(config_name, file_content, should_fail):
    with tempfile.TemporaryDirectory() as temp_dir:
        config_path_good = Path(temp_dir) / f"{config_name}.yaml"
        config_path_good.write_text(file_content)
    
        with hydra.initialize_config_dir(version_base=None, config_dir=temp_dir):
            cfg = hydra.compose(config_name=config_name)

            try:
                instantiated = hydra.utils.instantiate(
                    cfg, 
                    target_wrapper=CHECKER,
                )
                print(instantiated)
                if should_fail:
                    raise RuntimeError(f"`{config_name}` was supposed to fail but didn't: {e}")

            except hydra.errors.InstantiationException as e:
                raise RuntimeError(f"`{config_name}` was not supposed to fail: {e}")


def test_compatible_config():
    doer(config_name_good, config_example_good, should_fail=False)
        

def test_incompatible_config():
    doer(config_name_error, config_example_error, should_fail=True)


if __name__ == "__main__":
    test_compatible_config()
    test_incompatible_config()
```

## Related Issues and PRs

### DIFF
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -145,7 +145,7 @@ def _resolve_target(
     return target
 
 
-def instantiate(config: Any, *args: Any, **kwargs: Any) -> Any:
+def instantiate(config: Any, target_wrapper: Optional[Callable[..., Any]], *args: Any, **kwargs: Any) -> Any:
     """
     :param config: An config object describing what to call and what params to use.
                    In addition to the parameters, the config must contain:
@@ -168,6 +168,7 @@ def instantiate(config: Any, *args: Any, **kwargs: Any) -> Any:
                                   are converted to dicts / lists too.
                    _partial_: If True, return functools.partial wrapped method or object
                               False by default. Configure per target.
+    :param target_wrapper: Optional callable wrap _target_ with before it itself is called.
     :param args: Optional positional parameters pass-through
     :param kwargs: Optional named parameters to override
                    parameters in the config object. Parameters not present
@@ -224,7 +225,7 @@ def instantiate(config: Any, *args: Any, **kwargs: Any) -> Any:
         _partial_ = config.pop(_Keys.PARTIAL, False)
 
         return instantiate_node(
-            config, *args, recursive=_recursive_, convert=_convert_, partial=_partial_
+            config, *args, recursive=_recursive_, convert=_convert_, partial=_partial_, target_wrapper=target_wrapper,
         )
     elif OmegaConf.is_list(config):
         # Finalize config (convert targets to strings, merge with kwargs)
@@ -247,7 +248,7 @@ def instantiate(config: Any, *args: Any, **kwargs: Any) -> Any:
             )
 
         return instantiate_node(
-            config, *args, recursive=_recursive_, convert=_convert_, partial=_partial_
+            config, *args, recursive=_recursive_, convert=_convert_, partial=_partial_, target_wrapper=target_wrapper,
         )
     else:
         raise InstantiationException(
@@ -281,6 +282,7 @@ def instantiate_node(
     convert: Union[str, ConvertMode] = ConvertMode.NONE,
     recursive: bool = True,
     partial: bool = False,
+    target_wrapper: Optional[Callable[..., Any]]
 ) -> Any:
     # Return None if config is None
     if node is None or (OmegaConf.is_config(node) and node._is_none()):
@@ -314,7 +316,7 @@ def instantiate_node(
     # If OmegaConf list, create new list of instances if recursive
     if OmegaConf.is_list(node):
         items = [
-            instantiate_node(item, convert=convert, recursive=recursive)
+            instantiate_node(item, convert=convert, recursive=recursive, target_wrapper=target_wrapper)
             for item in node._iter_ex(resolve=True)
         ]
 
@@ -331,6 +333,9 @@ def instantiate_node(
         exclude_keys = set({"_target_", "_convert_", "_recursive_", "_partial_"})
         if _is_target(node):
             _target_ = _resolve_target(node.get(_Keys.TARGET), full_key)
+            if target_wrapper:
+                _target_ = target_wrapper(_target_)
+            
             kwargs = {}
             is_partial = node.get("_partial_", False) or partial
             for key in node.keys():
@@ -340,7 +345,7 @@ def instantiate_node(
                     value = node[key]
                     if recursive:
                         value = instantiate_node(
-                            value, convert=convert, recursive=recursive
+                            value, convert=convert, recursive=recursive, target_wrapper=target_wrapper,
                         )
                     kwargs[key] = _convert_node(value, convert)
 
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -4,7 +4,7 @@ import copy
 import functools
 from enum import Enum
 from textwrap import dedent
-from typing import Any, Callable, Dict, List, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
 from omegaconf import OmegaConf, SCMode
 from omegaconf._utils import is_structured_config
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -145,7 +145,7 @@ def _resolve_target(
     return target
 
 
-def instantiate(config: Any, target_wrapper: Optional[Callable[..., Any]], *args: Any, **kwargs: Any) -> Any:
+def instantiate(config: Any, *args: Any, target_wrapper: Optional[Callable[..., Any]] = None, **kwargs: Any) -> Any:
     """
     :param config: An config object describing what to call and what params to use.
                    In addition to the parameters, the config must contain:
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -282,7 +282,7 @@ def instantiate_node(
     convert: Union[str, ConvertMode] = ConvertMode.NONE,
     recursive: bool = True,
     partial: bool = False,
-    target_wrapper: Optional[Callable[..., Any]]
+    target_wrapper: Optional[Callable[..., Any]] = None,
 ) -> Any:
     # Return None if config is None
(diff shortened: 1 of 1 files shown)


JulesGM:
@leycec you might find this interesting.

odelalleau:
Makes sense to me. But it seems to me (from a distance -- I'm not super familiar with the intantiation code) that it could be implemented as a `_target_wrapper_` argument that could be either provided in the config or when calling `instantiate()`. If that's the case, it would better fit the current API (IMO).

JulesGM:
@Jasha10 you might be interested in this.

Jasha10:
@JulesGM are you familiar with the hydra-zen project, and specifically with https://mit-ll-responsible-ai.github.io/hydra-zen/generated/hydra_zen.instantiate.html? The `hydra_zen.instantiate` function takes a `_target_wrapper_` argument that seems similar.

leycec:
**Luv it.** Thanks so much for pinging me on, @JulesGM. I have now learned many things. I have learned about [the appropriately named OmegaConf](https://omegaconf.readthedocs.io), which I now realize I have been waiting for my entire life. I have also learned that @patrick-kidger of [`jaxtyping`](https://docs.kidger.site/jaxtyping/api) fame is still right about everything.

Interestingly, `jaxtyping` offers a [similar runtime type-checker-agnostic `@jaxtyping.jaxtyped(typechecker=beartype.beartype)` API](https://docs.kidger.site/jaxtyping/api/runtime-type-checking/) to that of the proposed `hydra.utils.instantiate(cfg, target_wrapper=beartype.beartype)` API:

```python
# Import both the annotation and the `jaxtyped` decorator from `jaxtyping`
from jaxtyping import Array, Float, jaxtyped

# Use your favourite typechecker: usually one of the two lines below.
from typeguard import typechecked as typechecker
from beartype import beartype as typechecker

# Type-check a function
@jaxtyped(typechecker=typechecker)
def batch_outer_product(x: Float[Array, "b c1"],
                        y: Float[Array, "b c2"]
                      ) -> Float[Array, "b c1 c2"]:
    return x[:, :, None] * y[:, None, :]
```

So for so good. You're in good company, @JulesGM. But I've been wondering... can we eventually simplify and streamline the process of selecting competing runtime type-checkers *or* is literally every Python package ever going to now define its own non-orthogonal proprietary API for selecting competing runtime type-checkers? It's the latter, isn't it? I'm sighing. You can almost feel the hot fetid breath I'm exhaling. :face_exhaling: 

If we accept the current status quo and do nothing,<sup>← *what will happen, guaranteed*</sup> then user headaches explode combinatorially. Currently, users have to manually notify *every* Python package of their preferred runtime type-checker with an API unique to that package. Instead, users should be able to trivially, publicly, and globally notify *every* Python package of their preferred runtime type-checker all-at-once simultaneously with just a single Python statement. Instead, we're now facing the exact opposite scenario.

*Introducing...*

## `anytype`: Utopia Never Seemed So Far Away

I'd make `anytype`. But I can barely make @beartype. The core conceit is simple, though: it's [QtPy](https://github.com/spyder-ide/qtpy), but for runtime type-checking. In a nutshell, `anytype` would:

* Be a **thin high-level abstraction layer** over lower-level runtime type-checkers.
* Only support features **commonly supported** by *all* runtime type-checkers.
* Provide a **uniform API** for performing runtime type-checking.
* Provide a **configuration API** for selecting between supported runtime type-checkers.
* **Internally delegate** all runtime type-checking to the currently configured type-checker.

Downstream packages like Hydra and `jaxtyping` would then simply import `anytype` and use that high-level API *without* concern for the currently configured type-checker. For example, the `@anytype.anytype` decorator would be a shim for either the `@beartype.beartype` *or* comparable `@typeguard.typechecked` decorators: e.g.,

```python
from anytype import anytype  # <-- type-checker-agnostic decorator for the win

@anytype  # <-- you just won the internet. congrats
@dataclass
class ExampleDataclass:
    name: str
    unit_price: float
    quantity_on_hand: int = 0

inst = ExampleDataclass("a name", 123.32, 23)
# Is ok

inst = ExampleDataclass(123, None, 23)
# Breaks
```

Downstream users and apps would then also import and configure `anytype` to use their preferred runtime type-checker. For example, the public `anytype.configure()` function might accept a `checker` parameter whose value is an `AnytypeChecker` enumeration member identifying the desired runtime type-checker: e.g.,

```python
# Probably in the "{your_app}.__init__" subpackage to ensure this happens early:
from anytype import AnytypeChecker, configure
configure(checker=AnytypeChecker.beartype)  # <-- feel the hot claws as they rake your codebase
```

Of course, nobody has time, energy, inclination, money, or sufficient goodwill towards humanity. Nobody will ever do that. I am nobody, too. Still, a utopian dreamer dreams. If not here on GitHub, then where? Nordic Gods above, where!?


<sup>i am very tired and must now lie on a door</sup>

## Would You Like to More About Managed Democracy and `anytype`?

If so, [hit this feature request at the @beartype issue tracker](https://github.com/beartype/beartype/issues/355). Let's collaborate and listen. Let someone else do this for all of us.

rsokl:
Heh this is crazy timing! I just added this feature in hydra-zen [literally four days ago](https://github.com/mit-ll-responsible-ai/hydra-zen/pull/666). It is extremely powerful; e.g. [adding a pydantic parsing layer](https://mit-ll-responsible-ai.github.io/hydra-zen/changes.html#rc1-2024-04-01) (and adding beartype support would be straightforward as well!)

> that it could be implemented as a _target_wrapper_ argument that could be either provided in the config or when calling instantiate()

I think it should live in `instantiate`. Otherwise this grows the complexity of configs in a substantial and probably not very useful way.

With this feature, it would be very nice to be able to completely disable Hydra/omegaconf's type-checking features and just let this validation layer be responsible for everything.

rsokl:
Btw @JulesGM  you should check out hydra-zen sometime 😄  It is really nice for ML workflows and eliminates a lot of manual labor and boilerplate from using Hydra

odelalleau:
> > that it could be implemented as a _target_wrapper_ argument that could be either provided in the config or when calling instantiate()
> 
> I think it should live in `instantiate`. Otherwise this grows the complexity of configs in a substantial way, and probably without much upside.

Just to be clear, my suggestion would provide the additional flexibility of specifying the wrapper in the config but this wouldn't be mandatory: you could still provide it as a kwarg when calling `instantiate()`, just like you can override any other config setting.

rsokl:
I understand, I just think that adding this to the config abstraction gives a degree of complexity that leads users down design paths that they ought not take. You also get weird things when you have nested configs that each have a target wrapper...People will start asking for non-recursive wrappers, etc

JulesGM:
We are planning to use the ConfigStore approach to do schema and most of the type checking, but it feels very annoying (and increase the maintenance work as the parameters change) to have to write down dataclasses for things that have a `_target_` with valid type annotations.

We feel like having a basic version of _target_wrapper_ in hydra-core would be a pretty good place to start with to cover that, without having to go all the way with the inclusion of hydra-zen. The debate of whether this can be included in the configs can maybe be done further down the road as usage evolves?

rsokl:
> without having to go all the way with the inclusion of hydra-zen

fwiw hydra-zen's only installation dependencies are hydra-core and typing-extensions, so it is very lightweight. And all you would need to do is use `hydra_zen.instantiate` in place of `hydra.utils.instantiate` -- that's it; you don't need to change any of your other Hydra code. 

I'm not opposed at all to having this be added to hydra-core, just want to point out that this is also a trivially easy path forward. I just need to officially release v0.13.0, which I can probably do this week.

Jasha10:
> all you would need to do is use `hydra_zen.instantiate` in place of `hydra.utils.instantiate`

@JulesGM would that work for you?

rsokl:
hydra-zen v0.13.0 was released yesterday: https://mit-ll-responsible-ai.github.io/hydra-zen/changes.html#v0-13-0

omry:
Thanks for the PR. I agree that type validation for `instantiate()` arguments is a useful direction, but I do not want to add it through this `target_wrapper` API as-is.

I opened #3203 to track the feature/design discussion directly. That is the preferred avenue for adding this capability. Please discuss the API shape and semantics there before sending a new PR.

A compliant solution should have an agreed design first, and then include tests, documentation, and a news fragment if it changes user-visible behavior.

PR #2909, opened 2024-06-04
remove pytest-snail from dev requirements file

omry:
fixes #2908

PR #2916, opened 2024-06-19
Debugging macos CI failures

Jasha10:
Based on PR #2909.

Currently looking at homebrew-related MacOS CI failure https://app.circleci.com/pipelines/github/facebookresearch/hydra/15662/workflows/d3e1e4ea-301a-4b33-b899-61f2867925ea/jobs/162011

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,7 @@ commands:
       - run:
           name: Preparing environment - Brew
           command: |
+            set -x
             brew --version
             brew untap Homebrew/homebrew-cask-versions
             brew update
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,6 +125,7 @@ jobs:
         type: string
     macos:
       xcode: "15.3.0"
+    resource_class: macos.m1.medium.gen1
     steps:
       - macos:
           py_version: << parameters.py_version >>
@@ -203,6 +204,7 @@ jobs:
         type: string
     macos:
       xcode: "13.4.1"
+    resource_class: macos.m1.medium.gen1
     steps:
       - macos:
           py_version: << parameters.py_version >>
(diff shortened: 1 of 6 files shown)


Jasha10:
There's currently a "brownout" on circleCI's [macOS Intel Support](https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718). According to that link, circleCI Intel resources will be sunset on June 28, 2024. This is causing the current failures of all macOS jobs.

omry:
Oh, this is starting to make sense.
So the cash was cashing intel binaries and they were fetched on M1 hardware?

omry:
About Nevergrad:
I am actually aware of an issue with nevergrad and the latest numpy.
They fixed it in https://github.com/facebookresearch/nevergrad/pull/1613.
Until they release a new version compatible with the latest numpy I think we can just pin numpy.

Jasha10:
> So the cash was cashing intel binaries and they were fetched on M1 hardware?

Yes 😄

Jasha10:
> About Nevergrad:
> ...
> Until they release a new version compatible with the latest numpy I think we can just pin numpy.

Ok, sounds good!

crassirostris:
@omry @Jasha10 do you need help? I'm happy to help to make Hydra work with the newer versions of antlr

Jasha10:
@crassirostris yes please! If you can finish the PR to get the CI green, that would be great.
But I didn't think that antlr was the problem...?

omry:
Closing this draft as superseded by subsequent CI work. The CircleCI macOS setup being debugged here has since been removed in favor of GitHub Actions, and the remaining relevant changes have either landed independently or been replaced by newer implementations.

Thanks for investigating the failures and documenting the root cause.

PR #2928, opened 2024-07-24
Use latest configstore singleton instead of storing it

jesszzzz:
## Motivation

Instead of storing the `ConfigStore.instance()` on `StructuredConfigStore`, refetch it every time. This ensures that if somehow the `ConfigStore` singleton is duplicated, we're getting the most recent one instead of an old copy (when Hydra loads configurations it makes a deep copy of the repository and deep copies of its singletons generate a new instance of the singleton, which causes issues with lazy imports)

Yes

## Test Plan

Fairly safe change, CI passes. That being said, if anyone is accessing `StructuredConfigSource.store` (which seems unlikely), they should switch to using `ConfigStore.instance()`

## Related Issues and PRs

### DIFF
--- a/hydra/_internal/core_plugins/structured_config_source.py
+++ b/hydra/_internal/core_plugins/structured_config_source.py
@@ -9,12 +9,10 @@ from hydra.plugins.config_source import ConfigResult, ConfigSource
 
 
 class StructuredConfigSource(ConfigSource):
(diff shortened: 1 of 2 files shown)


omry:
This is acceptable but please add a test and a news fragment file.

andywag (approved):
lgtm

Issue #2929, opened 2024-07-26
[Feature Request] Add json parsing functions to convert


jesszzzz:
# 🚀 Feature Request

Add a json_str parser function that takes an object and converts it into a JSON string representation of the object
     

## Motivation

**Is your feature request related to a problem? Please describe.**

Some of our config values are JSON strings that are unstructured to maintain flexibility, eg.
```
@dataclass
SomeConfig:
     metadata: str  # this should be a json-formatted string)
     # other attributes
```

Overriding `metadata` is pretty difficult since there are so many quotes in JSON strings that need to escaped. Being able to specify the object and have the hydra parser convert it into a JSON-formatted string would make this much more convenient.

## Pitch

Enable an override function like `++metadata=json_str({a: 1, b: {c: ['d', 'e']}}` that would set the `metadata` value to the JSON representation of the object (in this case, `'{"a": 1, "b": {"c": ["d", "e"]}}'`)

**Describe alternatives you've considered**
Having `metadata` be `Any` and doing the conversion at runtime, would prefer to do the json conversion as early as possible to reduce having to do it every time the field is accessed.
Passing in a correctly formatted json string, this is just more tedious

**Are you willing to open a pull request?** Yes

omry:
Have you considered using OmegaConf interpolation to convert a node to JSON on access?
In the example below, structured_metadata is the metadata in a dict format (for example) and your own function would take that node and conver it to json:
```python
@dataclass
SomeConfig:
     metadata: str  = "${my_json_converter:${.structured_metadata}}"
     structured_metadata: Optional[Any] = None
     # other attributes

```

jesszzzz:
It's a cool idea, but relies on the tool author to set this up correctly, and the tool consumer to know/find this relationship between `metadata` and `structured_metadata`.  Having a json_str parsing function would give the tool consumer a more consistent experience.

omry:
My position is that it's not useful enough to be included.

PR #2930, opened 2024-07-26
Add json_str parsing function to overrides syntax

jesszzzz:
## Motivation

Adding a json_str function to simplify passing a JSON-formatted string as an override, for example `++metadata=json_str({a: 1, b: {c: ['d', 'e']}}` would set the metadata value to the JSON representation of the object (in this case, '{"a": 1, "b": {"c": ["d", "e"]}}'). This is much simpler to write than constructing the JSON string, with all its quotes, so that it will be correctly parsed in the override string.

Yes

## Test Plan

Updated the tests and they all pass

## Related Issues and PRs

This addresses issue https://github.com/facebookresearch/hydra/issues/2929

### DIFF
--- a/hydra/_internal/grammar/grammar_functions.py
+++ b/hydra/_internal/grammar/grammar_functions.py
@@ -1,5 +1,6 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 import builtins
+import json
 import random
 from copy import copy
 from typing import Any, Callable, Dict, List, Optional, Union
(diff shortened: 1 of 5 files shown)


andywag (commented):
I didn't see a nested test example ie {"a" : {"b" : "c"}} . I didn't see how it is supported in the code as well but might be missing something int the conversion step.

manav-a (approved):
Looks good in general but maybe add some more test cases with complex jsons

omry:
Thanks for contributing these changes.

Tagging @jesszzzz, who authored both changes, @manav-a, who approved this pull request, and @andywag, who approved https://github.com/facebookresearch/hydra/pull/2934, because this changes the planned direction for these APIs.

I revisited `json_str(...)` from this pull request together with `to_hydra_override_value_str()` from https://github.com/facebookresearch/hydra/pull/2934 while preparing Hydra 1.4. These APIs support related configuration transport use cases: `json_str(...)` produces JSON text from an override value, while `to_hydra_override_value_str()` expresses a programmatically generated Python value as Hydra override syntax.

The use cases are valid, but I do not think transport and application-specific serialization should become part of Hydra's global override language or its public utility API. Together these APIs create additional compatibility surfaces around quoting, escaping, and type handling. The escaping fixes in https://github.com/facebookresearch/hydra/pull/2974 illustrate that maintenance burden.

Hydra already supports dictionary and list values directly through the [`dictContainer` and `listContainer` productions in its override grammar](https://github.com/facebookresearch/hydra/blob/main/hydra/grammar/OverrideParser.g4#L45-L50). For example, an override can combine nested dictionaries and lists:

```shell
python app.py 'metadata={name:example,tags:[one,two],nested:{values:[1,2,3]}}'
```

The underlying use cases can be handled through existing extension points:

- Pass structured values directly using Hydra's dictionary and list syntax.
- Pass an explicitly quoted JSON string when the field requires JSON text.
- Use an application-owned [OmegaConf custom resolver](https://omegaconf.readthedocs.io/en/latest/custom_resolvers.html) when structured CLI input needs application-specific conversion.
- Use a `ConfigSource` plugin for large or reusable configurations. Meta already had a Configurator-backed config source that allowed Hydra to load configurations from Configurator instead of the filesystem or a Python package.

`json_str(...)` and `to_hydra_override_value_str()` have been present on `main` since August 2024. The subsequent escaping fixes suggest that they had already been adopted internally at Meta. Given how long they have existed, I need to assume that internal users may depend on them.

I therefore plan to retain both APIs in Hydra 1.4 with deprecation warnings, remove them from the documentation, and remove them completely in Hydra 1.5.

Issue #2977, opened 2024-10-29
[Feature Request] Add a generation strategy config to ax-sweeper plugin


lwelzel:
# 🚀 Feature Request

Add the possibility to configure the generation strategy of the sweeper in the ax-sweeper plugin.

## Motivation

I want to specify a specific generation strategy for the sweep executed by the AxSweeper. This is important when I want to use the sweeper to generate samples in a specific way, e.g. only on a Sobol grid. Defining a specific generation strategy is necessary when the default is not good, e.g. when running large sweeps that do not need to optimize an objective, or when the automatically chosen strategy is not good.
Right now the generation strategy is automatically chosen by the AxClient. From what I understand it is not possible to change this using a hydra config file.

This is also related to this feature request for setting max_parallelism (issue #1837) since max_parallelism is defined in the generation steps of a generation strategy. The ax-sweeper is also now several versions behind ax-platform and botorch so that it would be nice to also update the overall ax-sweeper interface (#2777).

## Pitch

I want to be able to define a generation strategy in a config file, e.g. like:
```python
defaults:
  - override hydra/sweeper: ax
hydra:
  sweep:
    dir: ${hydra.job.name}_${now:%Y-%m-%d_%H-%M-%S}
    subdir: jobnr_${hydra.job.num}_jobid_${hydra.job.id}
  sweeper:
    ax_config:
      experiment:
        name: "example"
      generation_strategy: # new generation_strategy config as part of ax_config
        steps: # can define several steps on the generation strategy
                   # I should be able to set the kwargs of ax.modelbridge.generation_strategy.GenerationStep in this config
        - model: SOBOL # new (list of) generation_step config as part of generation_strategy config
          num_trials: 5
          max_parallelism: 1  # the max_parallelism for a step can be set
        - model: GPEI
          num_trials: 2
        name: default_strategy
      max_trials: 7 # this should automatically be set to the sum of the num_trials of the steps
      params:
        ...
```

**Describe alternatives you've considered**
The other sweepers don't support the same features as Ax/BOTorch so that I would need to write something that replaces the ax-sweeper plugin.

**Are you willing to open a pull request?**
I have implemented a possible solution (PR #2978), however, I have not added tests for the feature yet since I dont have experience with defining automated tests.

omry:
[...]

PR #2978, opened 2024-10-29
Ax sweeper generation strategy feature

lwelzel:
Add the possibility to configure the generation strategy of the sweeper in the ax-sweeper plugin.

## Motivation

I want to be able to define a generation strategy with specific steps in the config for the ax-sweeper. See issue #2977

### Have you read the Contributing Guidelines
Yes

## Test Plan

The tests could be similar to the current tests for the ax-sweeper. I am not sure how this works with the old versions of the ax-platform and botorch since using their more recent versions lead to some errors because the ax-sweeper config ExperimentConfig dataclass was misconfigured (ExperimentConfig.minimize and ExperimentConfig.objective_name).

## Related Issues and PRs
Feature request: #2977


omry:
Thank you for the PR and for writing up the generation-strategy use case.

I am going to close this PR for now. The Ax Sweeper plugin has since moved to newer Ax versions, so this patch is stale against the current Ax API. It is also not ready as an implementation: it needs API/design review, tests, documentation, and a news fragment.

I would like Ax feature coverage to be improved systematically rather than piecemeal. I attached the linked feature issue #2977 under #3177, which is collecting concrete use cases for newer Ax Sweeper capabilities, including generation strategy configuration. If you are still interested in this feature, please chime in there with the use case and the config shape you would want.

PR #2985, opened 2024-11-15
Upgrade react-live version to remove dependency on dom-iterator

hstonec:
## Motivation

We got reported that dom-iterator:1.0.0 is a vulnerable dependency. However, it is already the latest version and the only package depends on it is react-live. So upgrading react-live to remove dom-iterator from the dependency.

Yes

## Test Plan
```
yarn
yarn start
```
Then verified the website works as expected.

## Related Issues and PRs
N/A


Sanjay-Ganeshan (approved):
lgtm

PR #3006, opened 2025-01-08
Update default_element.py

andywag:
Force deep copy when package overrides are used

## Motivation

Yes/No

## Test Plan

## Related Issues and PRs


omry:
Thank you for the PR.

I am going to close this for now. The PR does not include a concrete issue, repro, or failing test showing the package-override problem it is trying to solve, so the motivation is speculative from the maintainer side.

The proposed fix is also not acceptable as written because it mutates OmegaConf global state while constructing a defaults-list element. If there is a real package-override aliasing or mutation bug here, please open an issue with a minimal reproduction and expected behavior first. That would make it possible to evaluate a localized fix in the config composition path.

PR #3019, opened 2025-02-03
Fix int full key when deep copying in instantiate

jesszzzz:
## Motivation

Was seeing an error like this when the full config is a list:
```
hydra/_internal/instantiate/_instantiate2.py", line 161, in _deep_copy_full_config
    full_key = full_key[: -len(str(index))] + f"[{index}]"
TypeError: 'int' object is not subscriptable
```

This fixes it by typecasting to a string, and adds a test to avoid future instances

Yes

## Test Plan

new unittest passes

## Related Issues and PRs

### DIFF
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -151,8 +151,9 @@ def _deep_copy_full_config(subconfig: Any) -> Any:
         return copy.deepcopy(subconfig)
 
     full_key = subconfig._get_full_key(None)
-    if not full_key:
+    if full_key == "" or full_key is None:  # Do not exit early if full_key is 0
         return copy.deepcopy(subconfig)
+    full_key = str(full_key)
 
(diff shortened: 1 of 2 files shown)


hydra/_internal/instantiate/_instantiate2.py
```diff
@@ -151,8 +151,9 @@ def _deep_copy_full_config(subconfig: Any) -> Any:
         return copy.deepcopy(subconfig)
 
     full_key = subconfig._get_full_key(None)
-    if not full_key:
+    if full_key == "" or full_key is None:
```
tonykao8080, line 154: I think the new condition is the same as `not full_key`. is there a specific scenario you cannot use `not full_key`?

jesszzzz: Oh yeah that's another bug I forgot to mention in the summary, if it's the first item of a list, `full_key` is `0` so we can't use `not full_key`

tonykao8080 (approved):
lgtm

PR #3022, opened 2025-02-05
[Hydra] Improve error message in parse_overrides

wu0607:
Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

## Motivation

Yes/No

## Test Plan

## Related Issues and PRs

### DIFF
--- a/hydra/core/override_parser/overrides_parser.py
+++ b/hydra/core/override_parser/overrides_parser.py
@@ -2,11 +2,6 @@
 import sys
 from typing import Any, List, Optional
 
-from omegaconf.vendor.antlr4.error.Errors import (
-    LexerNoViableAltException,
-    RecognitionException,
-)
-
 from hydra._internal.grammar import grammar_functions
 from hydra._internal.grammar.functions import Functions
 from hydra.core.config_loader import ConfigLoader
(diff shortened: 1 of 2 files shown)


hydra/core/override_parser/overrides_parser.py
```diff
@@ -98,7 +98,8 @@ def parse_overrides(self, overrides: List[str]) -> List[Override]:
                     msg = f"Error parsing override '{override}'" f"\n{e}"
                 raise OverrideParseException(
                     override=override,
-                    message=f"{msg}"
+                    message=f"Error when parsing index: {idx}, string: {override} out of {overrides}."
```
jesszzzz, line 101: nit: `msg` already contains `override` so this can be simplified to just `Error when parsing index: {idx} out of {overrides}.`

tests/test_hydra_cli_errors.py
```diff
@@ -67,26 +74,36 @@
...
+) -> None:
     monkeypatch.chdir("tests/test_apps/app_without_config/")
     if isinstance(override, str):
         override = [override]
     cmd = ["my_app.py", "hydra.sweep.dir=" + str(tmpdir)] + override
     ret = normalize_newlines(run_with_error(cmd))
     assert (
-        re.search("^" + re.escape(normalize_newlines(expected.strip())), ret)
-        is not None
+        expected_substring_1.strip() in ret and expected_substring_2.strip() in ret
```
jesszzzz, line 95: not a blocker but could consider making the parameter `expected_substrings: List[str]` and here checking `all(s.strip() in ret for s in expected_substrings)`

hydra/core/override_parser/overrides_parser.py
wu0607: some exception types don't have `override` included, there I would keep the current implementation

tests/test_hydra_cli_errors.py
wu0607: @jesszzzz good suggestion! updated the implementation

Issue #1393, opened 2021-02-11
[Feature Request] support singleton pattern in recursive instantiation


omry:
The following config can in theory be used to instantiate one object
```yaml

container:
  _tareget_: ...
  object1:
    _tareget_: ...
    _singleton_: true

  object2: ${.object1}
```

```python
c = instantiate(cfg.container)
assert c.object1 is c.object2
```

This will will probably require changes in OmegaConf.

rsokl:
This would be very powerful. It would effectively enable Hydra to be used for dependency injection (but orchestrated via configs).

aliutkus:
This would really be an awesome feature.

Do I understand that it's gonna be implemented for future hydra versions ?


thanks a lot for the cool work

aliutkus:
As a note, I would like this to work:

```
misc:
  a:
     _target_: mymodule.A
     _singleton_: true
     foo: 1

b: 
   _target_: mymodule.B
   bar: 2
   a: ${misc.a}

c:
   _target_: mymodule.C
   baz: 3
   a: ${misc.a}
```

i.e., the reference would not need to happen inside the same object to be instantiated. This may be very useful for cases where the same object (say some datapipeline) is to be used by models or other things, while keeping some structure to the config

martinj96:
Hey! Any updates on the timeline concerning this feature? Thanks!

npuichigo:
Is there any plan or milestone to land this feature?

Ubadub:
Hi, I noticed that this was initially part of the Hydra 1.3.0 milestone but was removed. It seems like a very useful feature to have. Are there any updates on when we can expect this feature?

npuichigo:
@omry

yuvalkirstain:
@omry will be amazing to have this feature :) any ways to go around it for now?

Jasha10:
I'll comment to record one potential issue I see with the proposed API.

It looks like calling `resolve` before `instantiate` will result in `object1` and `object2` being different objects.

```python
from omegaconf import OmegaConf

cfgA = OmegaConf.create("""
object1:
 _target_: MyTarget
 _singleton_: true
object2: ${object1}
""")

cfgB = cfgA.copy()
OmegaConf.resolve(cfgB)

assert OmegaConf.to_yaml(cfgB) == """\
object1:
 _target_: MyTarget
  _singleton_: true
object2:
 _target_: MyTarget
  _singleton_: true
"""
```

[...]

PR #3025, opened 2025-02-07
Add reference tag to allow getting a varaible by reference

vladoovtcharov:
## Motivation

Adds the ability to reference a variable without making a copy, similar to that proposed in https://github.com/facebookresearch/hydra/issues/1393

Instead of _singleton_: true being added to the object, instead anywhere that wants to be instantiated as a reference to that variable can use the syntax:
`
foo:
  bar:
    _target_: "myclass"
    x: 4.0
reference_variable:
  _reference_: "foo.bar"
`
Then reference variable and foo.bar will refer to the same object after instantiate.

By registering a custom resolver the syntax ends up looking even nicer:

`
foo:
  bar:
    _target_: "myclass"
    x: 4.0
reference_variable: ${ref:foo.bar}
`

Yes

## Test Plan

Add test cases, especially for almost circular cases, e.g.
`
foo:
  a: ${ref:bar.y}
  b: 
    _target_: "ClassA"

bar:
  x: ${ref:foo.b}
  y:
    _target_: "ClassB"
`
Also a bit worried about performance/memory usage with the proposed implementation but not sure about the best way to test that.

## Related Issues and PRs

A different approach to [PR#3103](https://github.com/facebookresearch/hydra/pull/3013) but solves similar problem

### DIFF
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -22,6 +22,29 @@ class _Keys(str, Enum):
     RECURSIVE = "_recursive_"
     ARGS = "_args_"
     PARTIAL = "_partial_"
+    REFERENCE = "_reference_"
+
+
+class TaggedValue:
+    def __init__(self, traverse_node=None, value=None):
+        self.traverse_node = traverse_node
+        self.value = value
+
+    @property
+    def is_list(self):
+        return isinstance(self.traverse_node, list)
+
+    @property
+    def is_dict(self):
+        return isinstance(self.traverse_node, dict)
+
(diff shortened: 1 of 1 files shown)


omry:
Thank you for the PR. This is an interesting idea, and the `ref` syntax may be a useful direction for the broader shared-object / instantiate identity problem.

I am going to close this PR for now because this API has not had maintainer feedback or buy-in yet. Beyond the missing tests and news fragment, there are also fundamental design questions still open, including initialization order, how references interact with normal OmegaConf interpolation, and cycle detection / error semantics.

Please continue the design discussion on #1393 before opening a new implementation PR. This will not make it into the upcoming 1.4 release.

PR #3037, opened 2025-03-06
Update docusaurus to stop using prismjs

jesszzzz:
## Motivation

Prismjs is found to have a security issue (https://github.com/facebookresearch/hydra/security/dependabot/102).
It's being pulled in by the version of docusaurus we were using. This upgrades docusaurus to a newer version that doesn't use prismjs.

Along with this, we needed to
* Update the docusaurus config to add the apikey field to algolia (not entirely sure it'll work but we'll know better once this lands) and move the google analytics field
* Only include the FBOnly sidebar key when it has elements
* Remove the version prefix from versioned sidebars

Also did some stuff to make the tests pass
* Added types for setuptools so mypy stopped complaining
* Updated the github actions/cache version since the previous one has been killed

Yes

## Test Plan

`yarn; yarn build; yarn start`
Opened the dev website and checked it looked right, pages loaded etc
CI/CD

## Related Issues and PRs

### DIFF
--- a/.github/actions/macos/action.yml
+++ b/.github/actions/macos/action.yml
@@ -6,7 +6,7 @@ runs:
   using: composite
   steps:
   - name: restore_cache
-    uses: actions/cache@v3.3.2
+    uses: actions/cache@v4.2.0
     with:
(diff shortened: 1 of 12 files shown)


tonykao8080 (approved):
looks good

andywag (approved):

Issue #3062, opened 2025-05-19
[Feature Request] _once_ behavior for instantiate


swamidass:
# 🚀 Feature Request

I'd like a way to specify in a config that a tree should be instantiated **_once and only once_**, so that multiple identical objects aren't created when instantiate is called multiple times on the same config, or on a config with interpolations that copy a target to multiple parts of the tree.

## Motivation

**Is your feature request related to a problem? Please describe.**

Sometimes we want to ensure a subtree of the config is instantiated only ONCE. Making unnecessary copies, in these cases, can hammer resources or even break code.

Take this example:

```yaml
dataset:
  _target_: lightly.data.LightlyDataset
  root: data_dir
loader:
  _target_: torch.utils.data.DataLoader
  dataset: ${..dataset}  
  sampler:
    _target_: torch.utils.data.RandomSampler
    data_source: ${..dataset}.
```

Instantiating this config will, in fact, instantiate three separate instances of 'dataset', which is undesirable, because each one takes up memory. Even worse, calling instantiate a second time will create 3 more instances!

## Pitch

**Describe the solution you'd like**

I'd like this to be a built in feature of hydra (requiring just a minor version bump), using this syntax instead:

```yaml
dataset:
  _target_: lightly.data.LightlyDataset
  _once_: true                     # instantiate will only create this tree once, reusing it when needed.
  _once_key_: unique_key # OPTIONAL: use this key, instead of the default hash of the yaml
  root: data_dir
loader:
  _target_: torch.utils.data.DataLoader
  dataset: ${..dataset}  
  sampler:
    _target_: torch.utils.data.RandomSampler
    data_source: ${..dataset}
```

**Describe alternatives you've considered**

There are two inferior a work arounds. 

***factory code***

You could do boilerplate everywhere it was important.

```python

_made_dataset = None
def make_dataset(*args, **kwargs):
  global _made_dataset
  if _made_dataset is None:
    _made_dataset = ...
  return _made_dataset

```

It works. and it might be possible to turn into a decorator to reuse. 

But there are major downslides. This requires a lot of non-obvious boiler plate wrapping, and obfuscates the actual object from the conf file where we might want it to be clear.

***hackish plugin***

Currently there isn't a plugin hook for instantiation. You can, however, make a hackish plugin that will do this.

```python

_storage = {}

def once(
    _arg_: DictConfig
):
    arg = _arg_
[...]

PR #3063, opened 2025-05-19
_once_ feature and documentation to instantiate

swamidass:
## Motivation

Sometimes we want to ensure a subtree of the config is instantiated only ONCE. Making unnecessary copies, in these cases, can hammer resources or even break code.

Take this example:

```yaml
dataset:
  _target_: lightly.data.LightlyDataset
  root: data_dir
loader:
  _target_: torch.utils.data.DataLoader
  dataset: ${..dataset}  
  sampler:
    _target_: torch.utils.data.RandomSampler
    data_source: ${..dataset}.
```

Instantiating this config will, in fact, instantiate three separate instances of 'dataset', which is undesirable, because each one takes up memory. Even worse, calling instantiate a second time will create 3 more instances!

## Pitch

I'd like this to be a built in feature of hydra (requiring just a minor version bump), using this syntax instead:

```yaml
dataset:
  _target_: lightly.data.LightlyDataset
  _once_: true                     # instantiate will only create this tree once, reusing it when needed.
  _key_: unique_key # OPTIONAL: use this key, instead of the default hash of the yaml
  root: data_dir
loader:
  _target_: torch.utils.data.DataLoader
  dataset: ${..dataset}  
  sampler:
    _target_: torch.utils.data.RandomSampler
    data_source: ${..dataset}
```

This change is fully backwards compatible, and this PR passes all tests. It can be introduced initially beta in a patch release, or in a minor release. 

Yes

## Test Plan

No special set up. Several tests added to tests/instantiate.py that verify correct behaivior. No existing tests fail. No existing behavior changes.

## Related Issues and PRs

See feature request here: #3062

### DIFF
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -268,13 +268,8 @@ def instantiate(
         _recursive_ = config.pop(_Keys.RECURSIVE, True)
         _convert_ = config.pop(_Keys.CONVERT, ConvertMode.NONE)
         _partial_ = config.pop(_Keys.PARTIAL, False)
-        _once_ = config.pop(_Keys.ONCE, False)
-        _key_ = config.pop(_Keys.KEY, None)
 
-        return instantiate_node(
-            config, *args, recursive=_recursive_, convert=_convert_, partial=_partial_,
-            once=_once_, once_key=_key_
-        )
+        return instantiate_node(config, *args, recursive=_recursive_, convert=_convert_, partial=_partial_)
     elif OmegaConf.is_list(config):
         # Finalize config (convert targets to strings, merge with kwargs)
         # Create copy to avoid mutating original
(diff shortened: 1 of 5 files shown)


swamidass:
For what it's worth, I've made the functionality available in a separately installable project:

https://github.com/swamidass/hydra-once

omry:
Thank you for the PR and for putting together a concrete implementation.

The use case is valid, but I do not think this is the right API shape to merge. The direction I want to keep focused is the singleton/reference design in #1393: https://github.com/facebookresearch/hydra/issues/1393

This PR introduces a separate `_once_` / `_key_` API and global instantiate cache semantics, which overlaps with that broader design problem. I would rather avoid adding another partial API here before we settle the singleton behavior properly.

You are welcome to propose a design in #1393. After we discuss and agree on the design direction there, a PR implementing that agreed shape would be much easier to review.

swamidass:
Great. Seems that its fairly straight forward to adapt this implementation to your prefered syntax. Would you accept the pull request if that was done?

PR #3074, opened 2025-07-31
Add security warning

jesszzzz:
## Motivation

There is a potential security issue if the config or overrides are dictated by users since then they can inject system calls and run arbitrary commands/python code. This PR adds a blocklist that, by default, prevents certain system calls from being called through the config. It can be bypassed by adding specific allowed modules to HYDRA_INSTANTIATE_ALLOWLIST_OVERRIDE. It also adds such warnings to the in-code documentation and wiki.

Yes

## Test Plan

unit tests

## Related Issues and PRs


odelalleau:
I think making Hydra/OmegaConf more secure would be nice, though I'm not sure trying to "patch" it is realistic since there are many "holes" in the design (which clearly wasn't aimed towards security).

But that being said -- how confident are you in your blocklist being exhaustive for core Python modules? Is it coming from some existing resource?

PR #3081, opened 2025-08-25
Add "best_trial": best_trial.number to results_to_serialize in _impl.py

evansdoe:
## Motivation

Adding the `best_trial_number` to the final `optimization_results.yaml` file provides several key benefits for analysis, debugging, and automation.

***

##### Traceability and Debugging

The **trial number** serves as a unique identifier that directly links the summary results to the specific, detailed logs for that single run. Hydra saves each trial's output (like logs, model checkpoints, and plots) in a numbered subdirectory. Knowing the `best_trial_number` tells you exactly which folder to inspect to reproduce or debug the best-performing run without having to manually search for it.

***

##### Analysis of the Optimization Process

Knowing *when* the best result occurred during the hyperparameter search offers valuable insights:
* An **early** best trial (e.g., trial number 5 out of 100) might suggest the search space was good, but the optimizer quickly stagnated.
* A **late** best trial (e.g., trial number 98 out of 100) suggests the search was still finding improvements and that more trials might yield even better results.

***

##### Simplified Post-Processing

For automated workflows, having a direct identifier is crucial. A script can parse the results file, grab the `best_trial_number`, and use it to reliably retrieve the correct model artifact from the corresponding trial directory for deployment, further evaluation, or archiving. This makes the entire MLOps pipeline more robust and less complex.

Yes

## Test Plan

No. All you need is to run an experiment and check if the `best_trial_number` is added to the `optimization_results.yaml`. If this is, then that is exactly the purpose of this PR.

## Related Issues and PRs

No.

### DIFF
--- a/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
+++ b/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py
@@ -411,6 +411,7 @@ class OptunaSweeperImpl(Sweeper):
             best_trial = study.best_trial
             results_to_serialize = {
                 "name": "optuna",
+                "best_trial": best_trial.number,
                 "best_params": best_trial.params,
                 "best_value": best_trial.value,
             }


omry (changes requested):
Thanks for the contribution. I agree this could be useful metadata, but I do not want to merge it in the current shape.

Requested changes:

- Please use a clearer serialized key name, `best_trial_number`, instead of `best_trial`. The value being written is the trial number, not the trial object.
- Please add or update Optuna sweeper test coverage for `optimization_results.yaml` so this field is checked alongside `best_params` and `best_value`.
- Please update the Optuna sweeper docs section that shows `optimization_results.yaml` so the documented output matches the new serialized data.
- Please add a plugin news fragment for this user-visible Optuna sweeper output change.

omry:
Thanks for the contribution, and sorry for changing direction after my earlier positive response.

I revisited the broader HPO output design and found that this belongs in the existing standardization work tracked in https://github.com/facebookresearch/hydra/issues/468. I still think exposing useful optimization result metadata is worthwhile, but I do not want to add an Optuna-specific field to the current output format before deciding the common schema across HPO sweepers.

I am closing this PR in favor of that broader work. Thank you again for proposing it.

PR #3107, opened 2025-12-20
make multirun behaves more like run

### LINKED ISSUE #3106: [Bug] Basic sweeper would behavior wrong with value type dict
# 🐛 Bug
## Description

```python
# python main.py -m +a={x:1},{x:2} +a={y:10},{y:20}

import hydra
@hydra.main(version_base="1.3", config_path=None, config_name=None)
def main(cfg) -> None:
    print(cfg)

if __name__ == "__main__":
    main()
```
would only run with
```
[2025-12-20 18:18:03,320][HYDRA] Launching 2 jobs locally
[2025-12-20 18:18:03,320][HYDRA]        #0 : +a={y:10}
{'a': {'y': 10}}
[2025-12-20 18:18:03,348][HYDRA]        #1 : +a={y:20}
{'a': {'y': 20}}
```

It should be run 4 jobs
```
[2025-12-20 18:18:03,320][HYDRA] Launching 4 jobs locally
[2025-12-20 18:18:03,320][HYDRA]        #0 : +a={x:1} +a={y:10}
{'a': {'x': 1, 'y': 10}}
[2025-12-20 18:18:03,348][HYDRA]        #1 : +a={x:1} +a={y:20}
{'a': {'x': 1, 'y': 20}}
[2025-12-20 18:18:03,373][HYDRA]        #2 : +a={x:2} +a={y:10}
{'a': {'x': 2, 'y': 10}}
[2025-12-20 18:18:03,399][HYDRA]        #3 : +a={x:2} +a={y:20}
{'a': {'x': 2, 'y': 20}}
```

## Checklist

## To reproduce

see above

## Expected Behavior

Works the same way without `-m`

## System information
- **Hydra Version** :  
- **Python version** : 
- **Operating system** : 

## Additional context

https://github.com/facebookresearch/hydra/blob/1ac07c7c001f73b2a6cdbc4c2ad700287cdf592c/hydra/_internal/core_plugins/basic_sweeper.py#L111
The basic_sweeper would only take last key into account.

clouds56:
## Motivation

Before this PR, `BasicSweeper.split_arguments` would keep only last override for the same key.
in case `a=1,2,3 a=10,20,30` it works well, but in case `a={x:10} a={y:20}` it would have different behavior between `MULTI_RUN` and `RUN`.

This PR introduce `BasicSweeper.simplify_overrides` that would try its best to remove duplicated key override.
1. it would automatically test if key is a group
2. for group it would keep last override (that is `OverrideType.CHANGE`)
3. for non-group, it would check if the value is dict or primitive `(str, int, bool, float, list)`
4. dict and primitive is exclusive, only one will wins
5. if value type is changed to dict, it would keep all dict till end
6. if value type is changed to primitive, it would only keep last value

Yes

## Test Plan

## Related Issues and PRs

fixes #3106 

### DIFF
--- a/hydra/_internal/core_plugins/basic_sweeper.py
+++ b/hydra/_internal/core_plugins/basic_sweeper.py
@@ -108,12 +108,15 @@ class BasicSweeper(Sweeper):
 
         def check_primitive(x: Any) -> bool:
             return isinstance(x, (int, float, bool)) or isinstance(x, QuotedString)
+
         for i, override in enumerate(overrides):
             if override.is_sweep_override():
                 if override.is_discrete_sweep():
                     key = override.get_key_element()
                     is_primitive[i] = all(override.sweep_iterator(check_primitive))
-                    is_dict[i] = any(override.sweep_iterator(lambda x: isinstance(x, dict)))
+                    is_dict[i] = any(
+                        override.sweep_iterator(lambda x: isinstance(x, dict))
+                    )
                     if is_primitive[i]:
                         last_primitive[key] = i
                     if is_dict[i]:
@@ -129,7 +132,9 @@ class BasicSweeper(Sweeper):
 
         for i, override in enumerate(overrides):
             key = override.get_key_element()
-            if is_primitive.get(i, False) and (last_primitive.get(key, -1) != i or last_dict.get(key, -1) > i):
+            if is_primitive.get(i, False) and (
+                last_primitive.get(key, -1) != i or last_dict.get(key, -1) > i
+            ):
                 continue
             if is_dict.get(i, False) and last_primitive.get(key, -1) > i:
                 continue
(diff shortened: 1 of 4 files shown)


clouds56:
I have a question that, in order to detect if the key is group, we need config_loader, I'm not sure if we should use loader from `Override` or from `BasicSweeper`.

omry:
Thanks for investigating this. After rereading #3106, I agree that the current run/multirun behavior is inconsistent. The specific example can already be expressed with explicit leaf overrides (`+a.x=1,2 +a.y=10,20`), however, and the proposed `simplify_overrides` heuristics are too complex and brittle for this bug.

We do not plan to support repeated overrides of the same target. Instead, Hydra will reject duplicate override targets with a clear error as a minor breaking change; tracked in #3283.

Closing this PR as not planned.

PR #3017, opened 2025-01-29
submitit launcher async support

SimarKareer:
## Motivation

After launching a job via the hydra_submitit launcher, I noticed that the command line stayed active, waiting for the job to finish.  I prefer for my submission command to finish executing after the job is launched.  I've included a very hacky solution here, but would need someone more familiar with the codebase to properly implement this change.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebookresearch/hydra/blob/main/CONTRIBUTING.md)?

Yes

## Test Plan
Ideally we should include some flag (async vs sync), and ensure both modes still work correctly.  Sorry, if I had more time I'd do this, but figured I'd make a PR in case others also wanted this feature.

## Related Issues and PRs
https://github.com/facebookresearch/hydra/issues/2479

PR #3128, opened 2026-02-06
feat(submitit): add wait_for_completion option for fire-and-forget job submission

ListIndexOutOfRange:
## Summary

Adds a `wait_for_completion` config option to the submitit launcher. When set to `False`, jobs are submitted and the launcher returns immediately without waiting for results.

## Motivation

This is useful for long-running jobs (e.g., model training) when we don't want the submitting process to block.

## Usage

```yaml
hydra:
  launcher:
    wait_for_completion: false
```

## Example output
```
[HYDRA] Submitted job 16136162
[HYDRA]   stdout: /path/to/.submitit/16136162/16136162_0_log.out
[HYDRA]   stderr: /path/to/.submitit/16136162/16136162_0_log.err
```

## Design consideration

It seems like this feature was requested a few times and my approach is suspiciously simple, so I guess I'm missing something.

For now, I wanted this PR to be non-invasive. It does not require modifications to Hydra core. The thing is, as noted in #2479, "Hydra's BasicSweeper collects the returned values from each job in the sweep". Thus, I chose to return placeholders values, such as  `JobStatus.COMPLETED`. A more invasive, perhaps cleaner way, of doing things would be to create a new JobStatus, like SUBMITTED or INPROGRESS ?

Yes

## Test Plan

I guess a simple test with `wait_for_completion=True` is enough. I'll add it later if we agree on the usefulness and design.

## Related Issues and PRs

- #2479: feature request 
- #2965: closed, unmerged
- #3017: closed, unmerged

### DIFF
--- a/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
+++ b/plugins/hydra_submitit_launcher/hydra_plugins/hydra_submitit_launcher/config.py
@@ -27,6 +27,12 @@ class BaseQueueConf:
     name: str = "${hydra.job.name}"
     # redirect stderr to stdout
     stderr_to_stdout: bool = False
+    # _________________________________ NEW PARAMETER _________________________________ #
+    # Whether to wait for job completion before returning
+    # If False, jobs are submitted and the launcher returns immediately
+    # with JobStatus.UNKNOWN
+    wait_for_completion: bool = True
+    # _________________________________________________________________________________ #
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


omry:
Thanks for working on this. The submit-and-exit use case is valid, but Hydra's current launcher and sweeper interfaces are synchronous and expect completed `JobReturn` objects. Returning submitted jobs as `COMPLETED` would lose actual results and failure propagation.

Proper support requires the broader async execution redesign tracked in https://github.com/facebookresearch/hydra/issues/3201. I am closing this PR in favor of that work.

PR #3133, opened 2026-02-15
fix overrides dropped when --multirun flag appears between them

### LINKED ISSUE #3053: [Bug] --multirun yields error on specific command line flag ordering
# 🐛 Bug
## Description
Using the multirun flag leads to an error on specific orderings of the command line flags. I am not entirely sure whether this is a user error or not, but I could not discern the underlying rule of why the error appears in some cases, but not in others.

## To reproduce

test.py:
```python
# from https://github.com/facebookresearch/hydra/tree/main/plugins/hydra_submitit_launcher/example
import logging
import os
import time

import hydra
import submitit

log = logging.getLogger(__name__)


@hydra.main(version_base=None, config_path=".", config_name="config")
    env = submitit.JobEnvironment()
    log.info(f"Process ID {os.getpid()} executing task {cfg.task}, with {env}")
    time.sleep(1)


if __name__ == "__main__":
    my_app()
```

config.yaml:
```
# from https://github.com/facebookresearch/hydra/tree/main/plugins/hydra_submitit_launcher/example
defaults:
  - override hydra/launcher: submitit_slurm

task: 1
```

```bash
 (reward-redistribution) srambica@cluster:~/reward-redistribution$ python test.py task=1 --multirun hydra/launcher=submitit_slurm
usage: test.py [--help] [--hydra-help] [--version] [--cfg {job,hydra,all}] [--resolve] [--package PACKAGE] [--run] [--multirun] [--shell-completion]
               [--config-path CONFIG_PATH] [--config-name CONFIG_NAME] [--config-dir CONFIG_DIR] [--experimental-rerun EXPERIMENTAL_RERUN]
               [--info [{all,config,defaults,defaults-tree,plugins,searchpath}]]
               [overrides ...]
test.py: error: unrecognized arguments: hydra/launcher=submitit_slurm
```

```bash
(reward-redistribution) srambica@cluster:~/reward-redistribution$ python test.py hydra/launcher=submitit_slurm --multirun task=1
usage: test.py [--help] [--hydra-help] [--version] [--cfg {job,hydra,all}] [--resolve] [--package PACKAGE] [--run] [--multirun] [--shell-completion]
               [--config-path CONFIG_PATH] [--config-name CONFIG_NAME] [--config-dir CONFIG_DIR] [--experimental-rerun EXPERIMENTAL_RERUN]
               [--info [{all,config,defaults,defaults-tree,plugins,searchpath}]]
               [overrides ...]
test.py: error: unrecognized arguments: task=1
```

## Expected Behavior
Notably this error does not occur in the following cases, confirming that the ordering is the reason for the error:
```bash
python test.py --multirun hydra/launcher=submitit_slurm task=1
```

```bash
python test.py task=1 hydra/launcher=submitit_slurm --multirun
```

## System information
- **Hydra Version** : 
hydra-core                1.3.2
hydra-submitit-launcher   1.2.0
- **Python version** : 3.10.16

Mr-Neutr0n:
When `--multirun` (or any optional flag) is placed between override arguments on the command line, argparse only sees the overrides before the flag and rejects the rest as unrecognized. For example:

```
python app.py task=1 --multirun db=mysql
# error: unrecognized arguments: db=mysql
```

This happens because `nargs="*"` positional args stop consuming tokens as soon as argparse encounters an optional flag. The overrides after the flag never get picked up.

Fixed by switching from `parse_args()` to `parse_known_args()` and merging any remaining tokens back into the overrides list. All orderings now work:

```
python app.py --multirun task=1 db=mysql    # works (already worked)
python app.py task=1 db=mysql --multirun    # works (already worked)
python app.py task=1 --multirun db=mysql    # works (was broken)
```

Fixes #3053

### DIFF
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -619,7 +619,29 @@ def get_args_parser() -> argparse.ArgumentParser:
 
 
 def get_args(args: Optional[Sequence[str]] = None) -> Any:
-    return get_args_parser().parse_args(args=args)
+    return _parse_args(get_args_parser(), args)
+
+
+def _parse_args(
+    parser: argparse.ArgumentParser, args: Optional[Sequence[str]] = None
+) -> argparse.Namespace:
+    """Parse arguments, handling overrides that appear after optional flags.
+
+    Python's argparse does not correctly collect positional arguments
+    (nargs="*") when an optional flag like ``--multirun`` appears between
+    them.  For example::
+
+        python app.py task=1 --multirun db=mysql
+
+    would only see ``task=1`` as an override, leaving ``db=mysql`` as
+    unrecognised.  This helper uses ``parse_known_args`` and appends the
+    remaining tokens to the ``overrides`` list so that flag ordering no
+    longer matters.
+    """
+    parsed, remaining = parser.parse_known_args(args=args)
+    if remaining:
+        parsed.overrides.extend(remaining)
+    return parsed
 
 
 def get_column_widths(matrix: List[List[str]]) -> List[int]:
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -619,29 +619,7 @@ def get_args_parser() -> argparse.ArgumentParser:
 
 
 def get_args(args: Optional[Sequence[str]] = None) -> Any:
-    return _parse_args(get_args_parser(), args)
-
-
-def _parse_args(
-    parser: argparse.ArgumentParser, args: Optional[Sequence[str]] = None
-) -> argparse.Namespace:
-    """Parse arguments, handling overrides that appear after optional flags.
-
-    Python's argparse does not correctly collect positional arguments
-    (nargs="*") when an optional flag like ``--multirun`` appears between
-    them.  For example::
-
-        python app.py task=1 --multirun db=mysql
-
-    would only see ``task=1`` as an override, leaving ``db=mysql`` as
-    unrecognised.  This helper uses ``parse_known_args`` and appends the
-    remaining tokens to the ``overrides`` list so that flag ordering no
-    longer matters.
-    """
-    parsed, remaining = parser.parse_known_args(args=args)
-    if remaining:
-        parsed.overrides.extend(remaining)
-    return parsed
+    return get_args_parser().parse_intermixed_args(args=args)
 
 
 def get_column_widths(matrix: List[List[str]]) -> List[int]:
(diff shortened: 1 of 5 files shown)


Mr-Neutr0n:
Hi! Quick status: this still applies cleanly to the latest main as of today, and the PR is still mergeable. Re-requesting a review. If this looks good now, I would be happy to see it merged; if the repo has moved on, I am just as happy to close it.

omry:
Thanks I'll get to it.

omry (dismissed):
Thanks for the fix. I agree this is a real CLI bug, and supporting overrides on both sides of `--multirun` is the right behavior.

Before merging, could you please make a few updates?

1. Add a news fragment for the user-visible bugfix, e.g. `news/3053.bugfix`.

2. Add an end-to-end regression test through an actual `@hydra.main` app invocation. The current parser-level test is useful, but the reported failure happens through the normal app path. It would be good to cover a command like:

   ```bash
   python app.py task=1 --multirun db=mysql
   ```

3. Please preserve the existing unknown-option behavior. The current implementation switches from `parse_args()` to `parse_known_args()` and appends all remaining tokens to `overrides`. That fixes misplaced overrides, but it also means an unrelated unknown CLI option can make it into the overrides list and probably fail later, instead of failing immediately as an unrecognized argument. For example:

   ```bash
   python app.py task=1 --bad-flag db=mysql
   ```

   In the current PR, `--bad-flag` can be appended to `overrides`. This should continue to fail as an unknown CLI option during argument parsing. Unknown CLI options should fail predictably; only valid Hydra overrides that appear after known flags should be accepted.

Once those are covered, this looks like a good 1.4.0 bugfix candidate.

dilshukbansal-lgtm:
Dabb
Oh

omry:
Thanks @Mr-Neutr0n for the contribution. Your original fix and tests identified the argument ordering issue and established the expected behavior.

I took over this PR to complete the remaining review items. The final stack uses `argparse.parse_intermixed_args()` so configuration overrides can appear on both sides of Hydra flags while unknown options are still rejected. I also added regression coverage through a real `@hydra.main` invocation and added the required news fragment.

The completed stack is now pushed. I will wait for CI before merging.

Mr-Neutr0n:
Thanks @omry, and for finishing it off properly — `parse_intermixed_args()` is the right call and handles the both-sides case my version didn't, and the `@hydra.main` regression test is better coverage than what I had. Appreciate you taking it over rather than bouncing it back.

PR #3141, opened 2026-03-29
Replace pkg_resources with importlib.metadata for Python 3.13+ compatibility

### LINKED ISSUE #3131: [Bug] Setuptools has removed pkg_resources
# 🐛 Bug
## Description
pkg_resources module no longer exists in setuptools. Can be reproduced with Python 3.13 or newer and setuptools v82. The parse_requirements function can be imported from pip internal to remove this dependency:
```python
from pip._internal.req import parse_requirements
```

## Checklist

## To reproduce

Python 3.13 withg setuptools v82. Then run pip install using the git repo as source.

```
ModuleNotFoundError: No module named 'pkg_resources'
```

## Expected Behavior
Hydra can be installed

## System information
- **Hydra Version** :  Latest git commit
- **Python version** : 3.13.12 (also tested with 3.14.2)
- **Virtual environment type and version** : uv

joaquinhuigomez:
## Motivation

Python 3.13+ and setuptools >= v82 no longer ship `pkg_resources`, causing `ModuleNotFoundError` when installing Hydra from source. This PR removes all `pkg_resources` usage in favor of stdlib alternatives.

## Changes

- **setup.py**: Replace `pkg_resources.parse_requirements()` with direct line-by-line reading of `requirements.txt`. The requirements file is simple enough that full PEP 508 parsing is unnecessary.
- **ray launcher tests**: Replace `pkg_resources.get_distribution(lib).version` with `importlib.metadata.version(lib)` (available since Python 3.8).
- **noxfile.py**: Remove the `pkg_resources` deprecation warning filter, since the import no longer exists.

## Why not `pip._internal`?

The issue description suggests `from pip._internal.req import parse_requirements` as a fix. This is a private API with no stability guarantees -- pip explicitly warns against importing from `pip._internal`. The stdlib `importlib.metadata` module is the correct replacement.

## Test plan

- Verified `requirements/requirements.txt` parses correctly with the new approach (no complex markers or URL-based requirements that would need full PEP 508 parsing).
- `importlib.metadata.version()` is a drop-in replacement for `pkg_resources.get_distribution().version` and has been available since Python 3.8.

Closes #3131

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -504,7 +504,6 @@ def test_core(session: Session) -> None:
             session,
             "build_helpers",
             "tests",
-            "-W ignore:pkg_resources is deprecated as an API:DeprecationWarning",
             *session.posargs,
         )
     else:
--- a/plugins/hydra_ray_launcher/tests/test_ray_aws_launcher.py
+++ b/plugins/hydra_ray_launcher/tests/test_ray_aws_launcher.py
@@ -10,7 +10,7 @@ from pathlib import Path
 from typing import Any, Dict, Generator, Optional
 
 import boto3  # type: ignore
-import pkg_resources
+from importlib.metadata import version as get_pkg_version
 from botocore.exceptions import (  # type: ignore
     ClientError,
     NoCredentialsError,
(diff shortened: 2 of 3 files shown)


omry:
Closing as part of an automatic cleanup sweep.

This PR appears to be superseded: it addresses `pkg_resources` removal for #3131, but that issue was already resolved on `main` by #3149.

If this is an error, or if this PR still contains unique work that should be preserved, please comment and we can reopen or follow up.

PR #3196, opened 2026-05-31
ci: declare least-privilege workflow-level contents: read

arpitjain099:
This PR adds a workflow-level `permissions: contents: read` to 2 workflow(s) that currently have no `permissions:` block (and therefore get the default broad read-write token). Each affected workflow was inspected and only reads repository contents; no publish/release/push/comment paths, so the change is non-functional in steady state and just shrinks the blast radius.

GitHub's documented Actions security recommendation. Happy to split per-file or adjust naming if preferred.

### DIFF
--- a/.github/workflows/core_tests.yml
+++ b/.github/workflows/core_tests.yml
@@ -8,6 +8,10 @@ on:
     branches:
       - main
       - '*_branch'
+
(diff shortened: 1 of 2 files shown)


omry:
Thanks!

Issue #3210, opened 2026-06-11
Modernize Optuna Sweeper


omry:
## Motivation

Hydra is being maintained again, and the bundled Optuna Sweeper needs a broader modernization pass rather than a sequence of disconnected stale PRs.

The current plugin still depends on `optuna>=2.10.0,<3.0.0` and `sqlalchemy~=1.3.0`. Optuna has moved on to newer distribution APIs and newer major versions, and there are several old or abandoned Optuna-related issues and pull requests that should be triaged together.

I would like to get the Hydra Optuna plugin up to speed and would be happy to coordinate with the Optuna team on the upgrade path.

## Scope

This umbrella issue should cover at least:

- Upgrade the plugin dependency target from Optuna 2.x to a modern supported Optuna version.
- Replace deprecated Optuna distribution APIs with the current APIs.
- Revisit the SQLAlchemy pin and storage compatibility.
- Decide what compatibility range the plugin should support.
- Review stale Optuna-related PRs and either fold them into the modernization plan or close them with a clear reason.
- Refresh tests, examples, documentation, and news fragments for the final accepted changes.

## Related work

- #2162: Optuna v2.x to v3.0 API transition reported by the Optuna team.
- #3046: Optuna 4.2.1 upgrade PR.
- #3007: closed request for Optuna 3.x and 4.x support.
- #2562: SQLAlchemy deprecation/pinning issue.
- #2647: resume existing Optuna study behavior.
- #3081: expose best trial number in serialized results.
- #1679: resume study and avoid OOM with Optuna plugin.
- #1463: fault tolerance option for Optuna Sweeper.
- #2237: uncaught exception when Optuna run ends with NaN.
- #2293: modifying trials when using a sweeper.

## Coordination

Thanks @himkt for the original heads-up in #2162. 
Is it possible to get guidance or help from the Optuna team on the recommended modernization path for the Hydra plugin? I would be happy to work with your team on this. It would be even better if your team can lead the modernization. Happy to support from the Hydra side.

omry:
Completed by https://github.com/facebookresearch/hydra/pull/3211, which upgrades the Optuna Sweeper to Optuna 4.9 and modernizes the associated APIs, dependencies, samplers, tests, and documentation.

Thanks @himkt for leading this work.

PR #2995, opened 2024-12-01
Add gpsampler

michaelriedl:
## Motivation

I wanted to bring the Optuna plugin up to date with the newest version. I made the necessary changes to make the plugin compatible. I also added in the GPSampler even though it is still experimental since some people may want to use it.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebookresearch/hydra/blob/main/CONTRIBUTING.md)?

Yes.

## Test Plan

I modified the existing tests in the plugin folder. I have tested that they pass on my WSL setup.

## Related Issues and PRs
Issues:
#2562
#2162 

PRs:
#2360

PR #3211, opened 2026-06-13
Use Optuna v4 in Optuna Sweeper

himkt:
## Motivation

Use a newer versions of Optuna (v4) to use new sampling algorithms and more latest APIs.

Yes

Behavior is not change from users' perspective. But this PR will let users install additional dependencies (cmaes, torch, and scipy). Would it be better to make them as optional dependencies?

## Test Plan

Just run on a local environment and check the result on CI.

## Related Issues and PRs

- https://github.com/facebookresearch/hydra/issues/3210

I added my patch on https://github.com/facebookresearch/hydra/pull/2995 by @michaelriedl to include his contributions on the update.


himkt:
Successfully ran the example (https://hydra.cc/docs/plugins/optuna_sweeper/) on my local environment.

plugins/hydra_optuna_sweeper/setup.py
```diff
@@ -29,8 +29,10 @@
     python_requires=">=3.10",
     install_requires=[
         "hydra-core>=1.1.0.dev7",
-        "optuna>=2.10.0,<3.0.0",
-        "sqlalchemy~=1.3.0",  # TODO: Unpin when upgrading to optuna v3.0
+        "optuna>=4.7.0,<5.0.0",
+        "cmaes>=0.12.0",  # Required for `optuna.samplers.CmaEsSampler`
+        "torch",  # Required for `optuna.samplers.GPSampler`
+        "scipy",  # Required for `optuna.samplers.GPSampler`
```
himkt: Not sure if it is acceptable to add them to requirements (it might be OK since it is a plugin, rather than hydra core itself?).

pytest.ini
```diff
@@ -14,6 +14,8 @@ filterwarnings =
   error
   ; SQLAlchemy imported by old Optuna on Python 3.12+
   ignore:datetime.datetime.utcfromtimestamp\(\) is deprecated.*:DeprecationWarning:sqlalchemy.sql.sqltypes
+  ; Optuna's sqlite connections are finalized by the GC; ignore the ResourceWarning so it isn't escalated to an unraisable error on Python 3.12+
```
himkt: Python 3.13+ started to warn when a connection to SQLite3 DB is not closed (🔗 https://github.com/lektor/lektor/issues/1227, it is the issue of a third-party project though).

This potential problem already happens with old versions of Python and it might not be a critical immediately (https://docs.python.org/3.14/library/exceptions.html#ResourceWarning).

It is difficult to fix at this time, is it possible to add this warning filter for a workaround until a future upstream (Optuna-side) fix is introduced...? 

https://github.com/facebookresearch/hydra/actions/runs/27461106520/job/81174946488

himkt:
@omry I think tests that I should make happy are now working well (with one concern https://github.com/facebookresearch/hydra/pull/3211#discussion_r3407771442).

Can I ask your review and opinion for the change?

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
```diff
@@ -43,14 +43,9 @@ class TPESamplerConfig(SamplerConfig):
...
     seed: Optional[int] = None
 
-    consider_prior: bool = True
-    prior_weight: float = 1.0
-    consider_magic_clip: bool = True
-    consider_endpoints: bool = False
     n_startup_trials: int = 10
     n_ei_candidates: int = 24
     multivariate: bool = False
-    warn_independent_sampling: bool = True
```
omry, line 53: Can you explain the config changes?
those will change existing user configs so I want to make sure we are being deliberate.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
```diff
@@ -100,20 +95,35 @@ class NSGAIISamplerConfig(SamplerConfig):
 
 
 @dataclass
-class MOTPESamplerConfig(SamplerConfig):
+class GPSamplerConfig(SamplerConfig):
     """
-    https://optuna.readthedocs.io/en/stable/reference/generated/optuna.samplers.MOTPESampler.html
+    https://optuna.readthedocs.io/en/stable/reference/samplers/generated/optuna.samplers.GPSampler.html
```
omry, line 118: is this the same sampler just renamed, or are you changing the default sampler?

plugins/hydra_optuna_sweeper/example/sphere.py
```diff
@@ -1,6 +1,11 @@
+import warnings
+
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 import hydra
 from omegaconf import DictConfig
+from optuna.exceptions import ExperimentalWarning
+
+warnings.filterwarnings("ignore", category=ExperimentalWarning)
```
omry: why is the default example using experimental features?

plugins/hydra_optuna_sweeper/NEWS.md
```diff
@@ -1,3 +1,17 @@
...
+
+### Features
+
+- Updated to be compatible with Optuna 4.7.0+
+- Added GPSampler support for Gaussian Process based optimization ([Optuna GPSampler](https://optuna.readthedocs.io/en/stable/reference/samplers/generated/optuna.samplers.GPSampler.html))
+- Added QMCSampler support for Quasi Monte Carlo based optimization ([Optuna QMCSampler](https://optuna.readthedocs.io/en/stable/reference/samplers/generated/optuna.samplers.QMCSampler.html))
+
+### API Change (Renames, deprecations and removals)
+
+- Removed the `consider_prior`, `prior_weight`, `consider_magic_clip`, `consider_endpoints`, and `warn_independent_sampling` options from the TPE sampler config, as they are deprecated in Optuna 4.x
```
omry: if they are just deprecated, we should not remove them but deprecate them.
What are users supposed to use instead?

This is might be an opportunity to try [oc.deprecated](https://omegaconf.readthedocs.io/en/latest/custom_resolvers.html#oc-deprecated). But I think its only a fit if we have another key to point people at.

plugins/hydra_optuna_sweeper/NEWS.md
```diff
@@ -1,3 +1,17 @@
...
+- Removed MOTPESampler support, as it has been removed in Optuna 4.x
+
```
omry: Gotcha. this is a big change.
Maybe we should try to detect people trying to use the old one and give them a better error than "Not found".

plugins/hydra_optuna_sweeper/setup.py
omry: Give me more context here.
Is this optional for Optuna? which components wants those?
I strongly prefer not pull those in by default.

website/docs/plugins/optuna_sweeper.md
```diff
@@ -109,6 +104,19 @@ best_value: 0.0
...
+When you use these samplers, you might want to suppress the warnings from Optuna.
+You can do this by adding the following code to your script:
+
+```python
+import warnings
+from optuna.exceptions import ExperimentalWarning
+
+warnings.filterwarnings("ignore", category=ExperimentalWarning)
+```
+
```
omry: It would be better if we default to the non experimental sampler.
documented workaround for experimental is probably not the business of the plugin (unless to chooses to make experimental the default)

pytest.ini
omry: ok. can you check if some other ignores here can be removed? 
The one with hydra changing working directory seems stale for sure.

plugins/hydra_optuna_sweeper/setup.py
himkt: Yes. Optuna has the minimal dependencies by default (https://github.com/optuna/optuna/blob/4147736eec770914630dfae6af4cf497ae7c7866/pyproject.toml#L33-L41).

When we use `GPSampler`, we need to install `torch` and `scipy`.
https://optuna.readthedocs.io/en/stable/reference/samplers/generated/optuna.samplers.GPSampler.html#optuna.samplers.GPSampler

When we use `CmaEsSampler`, we need to install `cmaes`.
https://optuna.readthedocs.io/en/stable/reference/samplers/generated/optuna.samplers.CmaEsSampler.html#optuna-samplers-cmaessampler

There are several options here I think:

- Add install dependencies as I do in the current PR
- Add an instruction to let users install the packages on demand
- Limit the scope of this PR to modify the existing codebase to work with the latest version of Optuna (I personally don't want to exclude the GPSampler since #2995 was motivated by the introduction of this sampler)

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
himkt: Yeah I should have added the reason behind this change, sorry....

These arguments are now deprecated in the following releases.

- https://github.com/optuna/optuna/releases/tag/v4.3.0
- https://github.com/optuna/optuna/releases/tag/v4.9.0

With these parameters, Optuna will raise warnings. They are scheduled to be dropped so I also remove. However, it's literally possible to specify in the API so I want to your opinion how to deal with them.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
himkt: `MOTPESampler` was removed in the Optuna v4.0.0.
https://github.com/optuna/optuna/releases/tag/v4.0.0

That is now unified to `TPESampler` (works both on a single or multiple objectives).

plugins/hydra_optuna_sweeper/example/sphere.py
himkt: You're right, it is not needed always. The warning will be appeared when users change a sampler to use on https://github.com/himkt/hydra/blob/3263e1b6487042e0f628003af4c58a132fa9fe87/plugins/hydra_optuna_sweeper/example/conf/config.yaml#L3.

Dropped in 6caadf7.

pytest.ini
himkt: OK, let me work on the another PR!

plugins/hydra_optuna_sweeper/NEWS.md
himkt: Noticed you already answered my question https://github.com/facebookresearch/hydra/pull/3211#discussion_r3440674059. I'll revise the PR to use that.

plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
himkt: :link: https://github.com/facebookresearch/hydra/pull/3211#discussion_r3440731591

plugins/hydra_optuna_sweeper/setup.py
omry: Yeah, I think runtime checks are probably a better idea. users who need to use a particular subset would get an error telling them what are the deps. but we would not force the plugin to install the superset or fragment it into a million little pieces.

Hydra's testing itself, if covering those - would need to install those deps which is not ideal - but let's see how expensive the tests are. for Pytorch - please use the cpu version. I think there is one more sweeper that uses pytorch, maybe Ax. align with how it pulls in torch.

pytest.ini
omry: what other PR?

website/docs/plugins/optuna_sweeper.md
omry: missed this one?

plugins/hydra_optuna_sweeper/setup.py
himkt: Hmm, it looks ax sweeper install torch on `install_requires`:

https://github.com/facebookresearch/hydra/blob/71f9c9f7ca7e39df6d9f3c61897e6b02d7076f33/plugins/hydra_ax_sweeper/setup.py#L32

And it will install the wheel built for GPU (CUDA 13.0) according to the PyTorch documentation https://pytorch.org/get-started/locally/. :thinking: 

Anyway, it's OK for Optuna (not needed for testing). I dropped them and let users install manually 53e2e9e, it it not aligned with the AX sweeper though (since we do not need that for testing, it can be dropped).

pytest.ini
himkt: I meant I'd create another PR to do, but now I changed my mind; I'll do that in this PR as well.

website/docs/plugins/optuna_sweeper.md
himkt: Sorry I missed it. This is just a note for users who see the experimental warning when they choose experimental samplers on their workload. Current Optuna sweeper uses TPE, which is the non-default sampler. So the warning won't appear unless they choose by their hand. And I agree with you that this is not the business of the plugin. Just dropped in cae733f.

plugins/hydra_optuna_sweeper/NEWS.md
himkt: These parameters are not needed to be specified ("as the default settings are optimal for most use cases" in the release note of Optuna https://github.com/optuna/optuna/releases/tag/v4.9.0).

But also these parameters have default values now and I'm not sure how to do an appropriate deprecation handling... (i.e. users have no way to remove warnings by modifying configurations I think)

Can you please give me an example to show warnings for deprecated keywords with default values for a resource in hydra?

plugins/hydra_optuna_sweeper/NEWS.md
himkt: Tried to do that in cae733f.

```diff
diff --git i/plugins/hydra_optuna_sweeper/example/multi-objective-conf/config.yaml w/plugins/hydra_optuna_sweeper/example/multi-objective-conf/config.yaml
index d4cc4f2d74..aec5616c82 100644
--- i/plugins/hydra_optuna_sweeper/example/multi-objective-conf/config.yaml
+++ w/plugins/hydra_optuna_sweeper/example/multi-objective-conf/config.yaml
@@ -1,11 +1,9 @@
 defaults:
   - override hydra/sweeper: optuna
-  - override hydra/sweeper/sampler: nsgaii
+  - override hydra/sweeper/sampler: motpe

 hydra:
   sweeper:
-    sampler:
-      seed: 123
     direction: [minimize, minimize]
     study_name: multi-objective
     storage: null
```

```
> uv run python example/multi-objective.py --multirun
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.12`.
Error in call to target 'hydra_plugins.hydra_optuna_sweeper.config._motpe_removed':
ValueError("The 'motpe' sampler was removed in Optuna 4.0. Use 'hydra/sweeper/sampler=tpe' instead. TPESampler now handles multi-objective optimization.")
full_key: hydra.sweeper.sampler

```

Note that `--cfg hydra -p hydra.sweeper` does not raise an exception at that time since the target become `hydra_plugins.hydra_optuna_sweeper.config._motpe_removed`:

<details>

```
> uv run python example/multi-objective.py hydra/sweeper=optuna --cfg hydra -p hydra.sweeper
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.12`.
# @package hydra.sweeper
sampler:
  _target_: hydra_plugins.hydra_optuna_sweeper.config._motpe_removed
_target_: hydra_plugins.hydra_optuna_sweeper.optuna_sweeper.OptunaSweeper
direction:
- minimize
- minimize
storage: null
study_name: multi-objective
n_trials: 20
n_jobs: 1
max_failure_rate: 0.0
search_space: null
params:
  x: range(0, 5, step=0.5)
  'y': range(0, 3, step=0.5)
custom_search_space: null
```

</details>

pytest.ini
himkt: The change directly warning still seems to happens.
https://github.com/facebookresearch/hydra/actions/runs/27855789391/job/82442952424?pr=3211

plugins/hydra_optuna_sweeper/setup.py
omry: Thanks, dropping the heavy dependencies from the default install is the right direction.

For Optuna, I do not want `torch`, `scipy`, or `cmaes` in the default plugin dependencies. They should be optional runtime dependencies for the specific samplers that need them.

For tests/CI, it is fine to follow the Ax pattern and install CPU PyTorch in the plugin test environment, e.g. via the PyTorch CPU wheel index. But that should stay test-only.

Please add clear runtime errors for optional samplers when their dependencies are missing. For example, using `GPSampler` without the needed deps should tell the user to install `torch` and `scipy`; using `CmaEsSampler` without `cmaes` should tell the user to install `cmaes`.

pytest.ini
omry: If the sqlite `ResourceWarning` is still unavoidable with current Optuna, keeping a targeted warning suppression is okay.

Please make the comment precise: this is for Optuna sqlite storage cleanup under Hydra warnings-as-errors test policy. Also, if there are remaining questions about whether this should be fixed in Hydra or Optuna, please file a follow-up issue under the Optuna modernization tracker: #3210.

The stale/unrelated warning filters should still be removed if they are no longer needed.

plugins/hydra_optuna_sweeper/NEWS.md
omry: I think the compatibility path here is to keep these fields in Hydra config for now, but stop passing them to Optuna by default.

Can you change the deprecated fields to a sentinel value, probably `None`, and strip them before instantiating `TPESampler` when they are unset? If a user explicitly sets one of these deprecated fields, Hydra should emit a clear warning or error saying the option is no longer effective with Optuna 4 and will be removed in the next major Hydra version.

One implementation detail: because the sampler is recursively instantiated before `OptunaSweeper.__init__` receives it, this filtering needs to happen before the nested sampler target is called. Filtering later in `OptunaSweeper` or `OptunaSweeperImpl` would be too late. One reasonable shape is a small Hydra-side wrapper target for the TPE sampler config that strips `None` deprecated args and then calls `optuna.samplers.TPESampler`.

Also note that removing fields from a structured config requires opening it first, e.g. with `open_dict`, otherwise `pop`/`del` on structured `DictConfig` will fail.

The important bit is that default Hydra usage should not pass these deprecated arguments to Optuna and should not warn.

website/docs/plugins/optuna_sweeper.md
omry: Thanks, dropping that note sounds right. The plugin docs should not carry workaround text for experimental samplers unless Hydra itself defaults to one.

plugins/hydra_optuna_sweeper/NEWS.md
omry: This direction sounds good.

`motpe` was the old Optuna multi-objective TPE sampler, and in Optuna 4 that functionality moved into `TPESampler`. Keeping the `motpe` config group registered only to produce a clear migration error is useful.

Please make sure this is covered by a test, and keep examples on supported samplers. The examples should not intentionally use `motpe` if that now fails.

omry:
Hi @himkt, just checking in on this.

Do you need any help with the remaining review feedback? The main open items are around optional sampler dependencies/runtime errors, the TPE deprecated-field compatibility path, the targeted warning filter, and the `motpe` migration-error test.

I am marking this as awaiting response for now, but happy to continue helping here.

himkt:
@omry I'm so sorry it took me so many days to reply while I got stuck on how to respond to your feedback. I want to discuss the direction for runtime behaviors of optional dependencies and runtime deprecation check.

> optional sampler dependencies/runtime errors

Isn't it possible to defer raising runtime errors to Optuna...? I'm not so confident whether we won't change the sampler backend for `CmaEsSampler` and `GPSampler` in the future. If we have a runtime warning in hydra and Optuna changes the backend, error message would become nonsense.

> the TPE deprecated-field compatibility path

It is almost the same as previous comment. How about delegating the warning message to Optuna for deprecated arguments? Now these deprecated parameters are replaced with the sentinels, and users will see warning messages raised inside Optuna.

https://github.com/facebookresearch/hydra/pull/3211/changes#diff-35bc09021922daf7005966ef790182d0f61fad0c9f04e280d51576ee9228916bL46-R61

https://github.com/optuna/optuna/pull/6635/changes#diff-2cc3a4088f95a10e806b68655d375f6f7f5fe2de161a0182c16168e378be8065R378-R399

> the targeted warning filter

Sorry I might not understand your intention... but I updated warning related in ade1b2a.

> the motpe migration-error test

I'm not sure if I understand your intention as well. Does 0e79f50 meet your intention?

omry:
@himkt, can you join the Zulip chat? we can chat about it there. (not right now though).
You can see the link the main readme.

himkt:
OK, I joined and created the topic.

omry:
Thanks for working through this.

A few remaining things I think we should tighten up:

1. The new TPE config shape uses `None` for deprecated Optuna arguments so that Optuna can decide whether to warn and what default to apply. That seems like the right direction, but this only works as intended with newer Optuna.

   In Optuna 4.7.0, `TPESampler` still has concrete defaults for these arguments, and passing `None` changes what Hydra passes to Optuna rather than behaving like "argument omitted". The `None` sentinel behavior appears in Optuna 4.9.0.

   How do we want to handle Optuna versions before 4.9 here? Options I see are:

   - require `optuna>=4.9.0`, or
   - keep compatibility with `>=4.7.0` by not passing these `None` values through.

   I slightly prefer raising the minimum if we want to delegate this behavior to Optuna.

2. `deterministic_objective` needs a type annotation. Without the annotation, dataclasses will not treat it as a field, so it will not be part of the Hydra/OmegaConf config and users will not be able to override it normally.

   I think this should be:

   ```python
   deterministic_objective: bool = False
   ```

3. Please add a plugin news fragment under `plugins/hydra_optuna_sweeper/news/` instead of editing `NEWS.md` directly. `NEWS.md` is generated during the release process, and the current edit also lands under a stale release heading/date.

himkt:
Thank you for the review @omry! Added commits:

- cf08ce9 increases the minimum version of Optuna
- a2eafea adds the type annotation for `deterministic_objective` (I've simply forgotten...)
- 8e1ac07 creates the news

omry:
Thanks, this is much closer.

A few remaining cleanup items:

1. CI lint is still red. Please fix the `lint` job before this can merge.

2. Please revert the direct edit to `plugins/hydra_optuna_sweeper/NEWS.md`. The new fragment under `plugins/hydra_optuna_sweeper/news/3211.feature` is the right place for this; `NEWS.md` is generated during release.

   Also, please keep the news fragment to a single concise entry. I am not sure multiline fragments render the way we want in `towncrier`. If this needs to describe both new sampler support and API/removal behavior, it may be better to split it into separate fragments, for example `3211.feature` and `3211.api_change`.

3. Can you strengthen the `motpe` test to exercise the actual Hydra config path, not only direct instantiation of `MOTPESamplerConfig`?

   The behavior we care about is that an existing user override like:

   ```text
   hydra/sweeper=optuna hydra/sweeper/sampler=motpe
   ```

   still resolves the old `motpe` config group and then fails with the new migration error:

   ```text
   The 'motpe' sampler was removed in Optuna 4.0. Use 'hydra/sweeper/sampler=tpe' instead.
   ```

   Directly instantiating `MOTPESamplerConfig` proves that the helper target raises, but it does not prove that the old config group remains registered and reachable through normal Hydra composition. That registration is the main compatibility path here.

himkt:
- I've forgotten to edit NEWS.md completely... Fixed in ecca466
- fa6e817 (+ e71d7b5) adds the test for motpe with .yaml

I just noticed that the lint CI failure also happens on the current main branch https://github.com/facebookresearch/hydra/actions/runs/28603558276/job/84820960904. How should I deal with?

omry:
Thanks, this is looking good. I checked the red website deploy job: the Docusaurus build succeeds, and the failure is only the PR workflow trying to push `gh-pages` without `WEBSITE_GITHUB_TOKEN`. That should be fine once this lands on `main`.

The remaining red lint check should be fixed by rebasing on current `main`.

One small content request: please collapse the news fragment to a single concise line. Suggested wording:

`Support Optuna 4.9.+, add GP/QMC samplers, and remove MOTPE support.`

After that, I think this is ready.

himkt:
Did in a4786cf (is it also OK to remove the mentions for deprecated arguments, right?).

omry:
Awesome, thank you.
Great work!

PR #3232, opened 2026-06-22
Reject swept configs that override multirun controller

### LINKED ISSUE #3231: Guard against Hydra launcher overrides inside swept job configs
## Problem

Hydra currently allows a swept job config to contain a defaults-list override for controller-level Hydra config, for example:

```yaml
defaults:
  - override /hydra/launcher: joblib
```

When this config is selected as part of a multirun sweep, the override is applied only while composing the individual job config. By that point the multirun controller has already selected the launcher, so the override is too late to affect the actual launcher used for the sweep.

This can make the composed job config appear to select `hydra/launcher=joblib` while the multirun still runs with the already-selected launcher, such as `basic`.

## Expected behavior

Hydra should not silently accept controller-level Hydra defaults-list overrides from swept job configs when they cannot affect the multirun controller. It should fail with a clear error explaining that launcher/sweeper selection for multirun must happen at the controller level, for example from the primary config or command line.

## Notes

This came up while triaging #2946. Hydra already rejects directly sweeping over Hydra overrides in some cases, but this is a related guardrail gap: a config selected by the sweep can still try to override controller-level Hydra choices too late.

This is separate from #851, which tracks the broader feature request to use configured launchers for regular runs without requiring `-m`.

sjh9714:
## Motivation

Fixes #3231.

A config selected only inside a sweep can currently override `hydra/launcher` in its defaults list. The composed job config then reports the new launcher even though the multirun controller already selected and is using the original launcher. This adds a guard that rejects late per-job changes to controller-level `hydra/launcher` and `hydra/sweeper` choices with a clear composition error.

Yes

#3231

## Test Plan

- `python3 -m venv .venv`
- `.venv/bin/python -m pip install --upgrade pip`
- `.venv/bin/pip install -r requirements/dev.txt`
- `.venv/bin/pip install -e .`
- `.venv/bin/pytest tests/test_hydra.py::test_multirun_rejects_swept_config_hydra_launcher_override -q` (reproduced the pre-fix failure)
- `.venv/bin/pytest tests/test_hydra.py::test_multirun_rejects_swept_config_hydra_launcher_override tests/test_hydra.py::test_app_with_sweep_cfg__override_to_basic_launcher -q`
- `NOX_PYTHON_VERSIONS=3.14 .venv/bin/nox -s test_core -- tests/test_hydra.py::test_multirun_rejects_swept_config_hydra_launcher_override tests/test_hydra.py::test_app_with_sweep_cfg__override_to_basic_launcher`
- `.venv/bin/black --check hydra/_internal/config_loader_impl.py tests/test_hydra.py tests/test_apps/sweep_hydra_launcher_override`
- `.venv/bin/isort --check --diff hydra/_internal/config_loader_impl.py tests/test_hydra.py`
- `.venv/bin/flake8 hydra/_internal/config_loader_impl.py tests/test_hydra.py`
- `git diff --check`
- `git diff --cached --check`

## Related Issues and PRs

Fixes #3231.

AI disclosure: I used Codex to inspect the multirun config-loading path, add the regression fixture, implement the guard, and run the checks listed above. I reviewed the final diff and test output before opening this PR.

### DIFF
--- a/hydra/_internal/config_loader_impl.py
+++ b/hydra/_internal/config_loader_impl.py
@@ -326,6 +326,10 @@ class ConfigLoaderImpl(ConfigLoader):
             overrides=overrides,
             run_mode=RunMode.RUN,
         )
+        ConfigLoaderImpl._ensure_no_sweep_config_controller_override(
+            master_config=master_config,
+            sweep_config=sweep_config,
+        )
 
         # Copy old config cache to ensure we get the same resolved values (for things
         # like timestamps etc). Since `oc.env` does not cache environment variables
@@ -333,6 +337,24 @@ class ConfigLoaderImpl(ConfigLoader):
         OmegaConf.copy_cache(from_config=master_config, to_config=sweep_config)
 
         return sweep_config
+
+    @staticmethod
+    def _ensure_no_sweep_config_controller_override(
+        master_config: DictConfig, sweep_config: DictConfig
+    ) -> None:
+        controller_groups = ("hydra/launcher", "hydra/sweeper")
+        master_choices = master_config.hydra.runtime.choices
+        sweep_choices = sweep_config.hydra.runtime.choices
+        for group in controller_groups:
+            master_choice = master_choices.get(group)
+            sweep_choice = sweep_choices.get(group)
+            if master_choice != sweep_choice:
+                raise ConfigCompositionException(
+                    f"'{group}' must be selected before the sweep starts, "
+                    f"but a swept job config changed it from "
+                    f"'{master_choice}' to '{sweep_choice}'. Select it in "
+                    "the primary config or from the command line."
+                )
 
     def get_search_path(self) -> ConfigSearchPath:
         return self.config_search_path
--- a/news/3231.bugfix
+++ b/news/3231.bugfix
@@ -1,0 +1,1 @@
+Reject swept configs that override the multirun launcher or sweeper.
--- a/tests/test_apps/sweep_hydra_launcher_override/__init__.py
+++ b/tests/test_apps/sweep_hydra_launcher_override/__init__.py
@@ -1,0 +1,1 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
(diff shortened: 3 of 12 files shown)


omry:
Thanks for the prompt help on this!

Overall diff looks good. One missing thing is regression coverage for the `hydra/sweeper` case.

There is already coverage for rejecting direct command-line sweeps over Hydra config, and for selecting a custom sweeper before the sweep starts, but this PR adds a different guard: rejecting sweep-level Hydra choices that are changed later by a swept job config, after the sweep has already selected them.

Since the implementation rejects both `hydra/launcher` and `hydra/sweeper`, please exercise both cases.

sjh9714:
Added the `hydra/sweeper` regression case alongside the existing `hydra/launcher` case.

Verified with:
- `.venv/bin/python -m pytest tests/test_hydra.py -k "swept_config_hydra_controller_override" -q`
- `.venv/bin/python -m pytest tests/test_hydra.py -q`
- `git diff --check`

omry:
Thanks, the added `hydra/sweeper` coverage is the right direction and the implementation still looks good.

One test hygiene issue: the new regression can leave behind `None/multirun.yaml` in the repo root. I reproduced that locally after running the focused test.

I think this is because `hydra_sweep_runner` is being called without an explicit `temp_dir`. The fixture stores `str(temp_dir)`, so `None` becomes the literal sweep directory `"None"`. In this failure path the basic sweeper writes `multirun.yaml` before the expected composition error is raised, and because the exception happens during `__enter__`, the context manager cleanup does not run.

Please pass an explicit temporary sweep directory in this test, for example by adding `tmpdir`/`tmp_path` and passing it as `temp_dir=...` to `hydra_sweep_runner(...)`.

sjh9714:
Fixed the test hygiene issue in `433c4b3` by passing `tmpdir` as `temp_dir` to `hydra_sweep_runner(...)` for this regression.

Verified with:
- `rm -rf None && .venv/bin/python -m pytest tests/test_hydra.py -k "swept_config_hydra_controller_override" -q && test ! -e None`
- `.venv/bin/python -m pytest tests/test_hydra.py -q && git diff --check && test ! -e None`

omry:
Thanks for the quick follow-up. I verified the updated PR locally, including the focused regression, the full `tests/test_hydra.py` file, and the cleanup check for the stray `None/` directory. This looks good now.

Issue #3202, opened 2026-06-08
Disallow parent traversal in Defaults List config groups


omry:
## Problem

Defaults List config groups are documented as paths relative to the containing config, or absolute paths when prefixed with `/`.

Parent traversal such as `../group` or `../../group` in a config group path is not a supported public contract, but it can currently semi-work in some paths. This makes Defaults List identity difficult to reason about and difficult or impossible to override cleanly from the command line.

Motivating example: #3104 used a nested Defaults List entry like:

```yaml
- ../../level12@out: level12_ini
```

The equivalent supported form should use an absolute config group:

```yaml
- /level1/level12@out: level12_ini
```

## Proposal

Disallow parent traversal in Defaults List config group paths.

Hydra should reject entries using `../` traversal in config groups with a clear error message explaining that parent traversal is unsupported and that users should use the absolute `/...` config group form instead.

## Notes

- This is a breaking API change for users relying on accidental/semi-working `../` traversal behavior.
- This should be handled in Hydra 1.4 before the behavior is treated as supported by a public release.
- `group: subgroup/option` is not the canonical style, but it should continue to work in general; the unsupported part is parent traversal in the config group path.

thatbagu:
Could I suggest to add link to searchpath_plugin in the error message if relative path escapes root configuration dir?

omry:
Searchpath plugin is actually the oldest way to add things to the searchpath. it's appropriate for libraries but not so much for users.
If anything, the [searchpath page](https://hydra.cc/docs/advanced/search_path/) is more apporpriate.

But really, this is more about the [defaults list](https://hydra.cc/docs/advanced/defaults_list/).
The relevant line there is "CONFIG_GROUP : A path to a set of configs. The path is relative to the containing config. It can be made absolute by prefixing it with a /. The path separator is / regardless of the operating system.".

thatbagu:
It seems it has been so long I didn't use hydra conf actively, thank you for clarification, adjusted my error messages in the PR :)

omry:
Implemented by #3246 and present on current `main`. I re-ran the focused validation with `nox -s test_core-3.14 -- tests/defaults_list/test_defaults_list.py -k parent_traversal`: all 13 selected parent-traversal tests passed. Closing this as completed.

Issue #2878, opened 2024-03-28
[Bug] Config hierarchy contains empty strings when using relative path in defaults 


Jahysama:
# 🐛 Bug
## Description
When complex config is created using relative paths in defaults list, it results in config hierarchy containing empty strings.

## Checklist
- [x] I checked on the latest version of Hydra
- [x] I created a minimal repro (See [this](https://stackoverflow.com/help/minimal-reproducible-example) for tips).

## To reproduce
Config structure:
`.
├── dir1
│   └── main.yaml
├── dir2
│   └── child.yaml
└── test.py
```
main.yaml:
```yaml
defaults:
 - _self_
 - ../dir2: child.yaml
```
child.yaml:
```yaml
param: value
```
test.py:
```python
from hydra import compose, initialize
from omegaconf import OmegaConf


config_path = "dir1"
initialize(config_path, version_base="1.3.2")
cfg = compose(config_name="main.yaml")
print(OmegaConf.to_yaml(cfg))
```

Running `test.py` results in the following config/output:
```
? ''
: ? ''
  : ? ''
    : dir2:
        param: value
```

## Expected Behavior
```
dir2:
  param: value
```

## System information
- **Hydra Version** : 1.3.2  
- **Python version** :  3.12.2
- **Virtual environment type and version** :  venv 
- **Operating system** :  Void Linux x86_64 

## Additional context
I need such a weird way of composing my config, because other folder is a git submodule and there is nothing i can do about it. Maybe there another way to solve this issue.

mk2001233:
+1

jesusdfc:
+1

zhaonat:
+1

zuble:
+1

timkpaine:
This is fixed in [lerna](https://github.com/nbprint/lerna?tab=readme-ov-file#relative-path-in-defaults-fix-2878)

omry:
Thanks for the report.

This is invalid because parent traversal in Defaults List config groups is not supported, especially when it is used to reach outside the configured config search path.

The empty-string hierarchy produced here is bad behavior, but the fix is not to make this form work. The fix is to reject this unsupported form explicitly with a clear error. That is tracked in #3202.

Please use a supported config source/search-path setup instead, for example by adding the other directory to Hydra’s config search path and selecting a config group from that search path, or by using an absolute config group path where applicable.

Closing this as unsupported/superseded by #3202.

PR #3246, opened 2026-06-28
Disallow parent traversal in Defaults List config group paths

thatbagu:
## Motivation

It has been 2 years since original issue https://github.com/facebookresearch/hydra/issues/2878 related to this PR have been opened by me, a lot has changed in my life since then, and now it is time to put this issue and following issues https://github.com/facebookresearch/hydra/issues/3202 to the conclusion.

Yes

## Test Plan

Tests have been added to cover and reproduce these new error messages appearing.

### DIFF
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -2,7 +2,6 @@
 
 import copy
 import os
-import posixpath
 import warnings
 from dataclasses import dataclass, field
 from textwrap import dedent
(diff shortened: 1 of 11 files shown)


hydra/_internal/defaults_list.py
thatbagu, line 436: I haven't put the check directly here, because I thought doing it downstream would be better for future readability, since all other checks would be near as well

omry:
Thanks for working on this.

I want this to reject `..` in config group paths regardless of where the defaults entry comes from: config files, nested defaults, and command-line defaults-list appends.

One reason I originally decided not to support relative parent traversal here is command-line override clarity. Defaults List config group paths have a stable identity: they are relative to the containing config, or absolute when prefixed with `/`. If parent traversal is allowed, it becomes harder to reason about what the canonical override key should be, especially from the command line. Seeing the full absolute group path in the Defaults List makes the behavior much clearer.

So the intended rule is simpler:

- `group/subgroup: option` remains supported.
- `/group/subgroup: option` remains supported.
- Any `..` path segment in a config group path should fail with a clear error.
- This should apply no matter whether the entry came from a config file or from a CLI append such as `+group=option`.

I think the current PR catches config-file entries but misses external appends, because `overrides.append_group_defaults` are added after the new validation point. Could you move/add the validation so appended defaults are covered too, and add a focused test for a CLI append containing `..`?

thatbagu:
Thank you for the explanation. I agree that stable, predictable group paths make override semantics clearer. I've simplified the validation to a single rule (reject any .. segment regardless of context) and added a test covering the CLI append path.

omry:
Thanks, this is moving in the right direction.

I think there is still one important gap: the validation currently checks only the config group path (`d.group`), but `..` can also appear on the selected option side.

For example, these are still accepted by the parser and flow into Defaults List processing:

```bash
+missing_group=../file1
+missing_group=[file1,../file1]
group1=../file1
```

The same gap exists in config files:

```yaml
defaults:
  - group1: ../file1
```

and:

```yaml
defaults:
  - group1:
      - file1
      - ../file1
```

The intended rule is that `..` is not allowed as a path segment anywhere in Defaults List config group selections, regardless of whether it appears in the group path or in the selected option path, and regardless of whether it came from a config file or a command-line override.

Suggested additional tests:

- CLI append with value traversal: `+missing_group=../file1`
- CLI append with list value traversal: `+missing_group=[file1,../file1]`
- CLI override of an existing defaults entry: `group1=../file1`
- config-file default: `group1: ../file1`
- config-file options list containing `../file1`

The distinction matters for the CLI tests:

- `+missing_group=...` should use a group that is not already in the defaults list, so it exercises the external append path.
- `group1=...` should use a config that already contains `group1` in the defaults list, so it exercises the override path instead of failing earlier with “No match in the defaults list.”

omry:
Thanks for working on this and for addressing the earlier feedback. I have taken over the PR, completed the remaining validation and test coverage, and rebased it onto current main. Your original commits and authorship are preserved. Thanks again for getting this started.

PR #3280, opened 2026-07-16
Fix broken Binder link in Jupyter notebooks doc

### LINKED ISSUE #2149: [Bug] Documentation Binder link is broken.
The link to the binder notebook seems to be broken at https://hydra.cc/docs/next/advanced/jupyter_notebooks/.

PS: It would be nice to have a way to submit these kind of documentation issues or feedback directly from the hydra.cc website e.g. with a widget "Did that page help you? y/n".

pctablet505:
The Binder badge link on the Jupyter notebooks doc page points to a broken URL. The `filepath` query parameter has `%2` where it should have `%2F` (the URL-encoded `/`), so it resolves to a nonexistent path instead of `examples/jupyter_notebooks`.

This fixes the encoding so the link points to the correct directory.

Fixes #2149

### DIFF
--- a/news/2149.docs
+++ b/news/2149.docs
@@ -1,0 +1,1 @@
+Fix broken Binder link in the Jupyter notebooks documentation page.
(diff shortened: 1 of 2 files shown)


omry:
Thanks!
Please accept the CLA.

If you are looking for additional bugs to fix, take your pick from the 1.4 milestone.

PR #3316, opened 2026-07-25
Preserve dataclass and attrs call-site arguments

### LINKED ISSUE #2364: [Feature Request] Preserve dataclasses as instantiate-arguments

omry:
Pass already-constructed dataclass and attrs instances to instantiate targets unchanged instead of implicitly converting them to Structured Configs.

Keep native and explicit OmegaConf containers on the configuration path. Add regression coverage for identity passthrough, special constructors, attrs, and explicit structured overrides.

Document the Hydra 1.4 breaking change and migration, and add attrs to the test environments.

Closes #2364

### DIFF
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -386,7 +386,9 @@ def _prepare_input_container(
 def _prepare_input_value(
     value: Any,
 ) -> Any:
-    if isinstance(value, (dict, list)) or type(value) is tuple:
+    if not is_structured_config(value) and (
+        isinstance(value, (dict, list)) or type(value) is tuple
+    ):
         return _prepare_input_container(value)
     return value
 
@@ -491,8 +493,8 @@ def instantiate(
     :param kwargs: Optional named parameters to override
                    parameters in the config object. Parameters not present
                    in the config objects are being passed as is to the target.
-                   IMPORTANT: dataclasses instances in kwargs are interpreted as config
-                              and cannot be used as passthrough
+                   Dataclass and attrs instances are passed through without
+                   conversion or recursive instantiation.
     :return: if _target_ is a class name: the instantiated object
              if _target_ is a callable: the return value of the call
     """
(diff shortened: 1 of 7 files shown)


omry:
@codex review

PR #3317, opened 2026-07-25
Honor custom exception hooks

### LINKED ISSUE #1431: [Bug] PrettyErrors doesn't work with Hydra
# 🐛 Bug
## Description
[PrettyErrors](https://github.com/onelivesleft/PrettyErrors) is a package by @onelivesleft, which makes stack trace more readable in terminal. Unfortunetely it seems it simply doesn't do anything when Hydra is used.

Is there any way to make Hydra and PrettyErrors compatible?

## Checklist

## To reproduce
Isntall PrettyErrors with:
`python -m pip install pretty_errors`
`python -m pretty_errors`

Run any file with Hydra:
```python
import hydra

@hydra.main(config_path="", config_name="config.yaml")
def main(config):
    raise Exception("Some exception")
    
if __name__ == "__main__":
    main()
```
(if we just comment out the hydra.main decorator then stack trace is printed correctly)

## Additional context
Checked both on ubuntu bash and windows powershell.

omry:
Delegate sanitized task exceptions to custom sys.excepthook handlers while preserving Hydra's default renderer when no custom hook is installed.

Rebuild sanitized tracebacks as real traceback objects so third-party handlers can consume them, and fall back to default rendering when a custom handler fails. Add regression coverage for Hydra and OmegaConf frame stripping and handler failure.

Fixes #1431

### DIFF
--- a/hydra/_internal/utils.py
+++ b/hydra/_internal/utils.py
@@ -6,9 +6,8 @@ import os
 import sys
 import traceback
 import warnings
-from dataclasses import dataclass
 from os.path import dirname, join, normpath, realpath
-from types import FrameType, TracebackType
+from types import TracebackType
 from typing import Any, List, Optional, Sequence, Tuple
 
 from omegaconf.errors import OmegaConfBaseException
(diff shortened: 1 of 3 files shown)


omry:
@codex?

PR #3319, opened 2026-07-25
Fail composition when a dangling override is rescued by an append

### LINKED ISSUE #3318: Dangling Defaults List override can override a command-line append
## Summary

A Defaults List override with no earlier Group Default to override is invalid by itself, but a matching command-line append can make the composition succeed. Hydra then discards the value requested by the append and selects the dangling override instead.

## Minimal reproduction

```yaml
# config.yaml
defaults:
  - override db: sqlite
```

Without a command-line append, this correctly fails because there is no `db` Group Default to override.

However:

```bash
python app.py +db=mysql
```

currently succeeds and selects `db: sqlite`.

## Expected behavior

The dangling `override db: sqlite` should fail before Hydra processes command-line overrides because it does not target an earlier `db` Group Default. Therefore, `+db=mysql` should never have an opportunity to make the invalid Defaults List appear valid.

## Current behavior

Hydra appends the external `db: mysql` Group Default, registers `override db: sqlite`, and applies that override to the appended entry. The result is `db: sqlite`.

This behavior is present in Hydra 1.3.0, Hydra 1.3.2, and current `main`.

## Relationship to other work

This was discovered while reviewing [PR #3315](https://github.com/facebookresearch/hydra/pull/3315), which fixes the separate override-precedence bug reported in [issue #3229](https://github.com/facebookresearch/hydra/issues/3229).

[Issue #3229](https://github.com/facebookresearch/hydra/issues/3229) concerns which of multiple valid Defaults List overrides wins. This issue concerns validation of an override with no target and its interaction with a command-line append. It should be fixed separately from [PR #3315](https://github.com/facebookresearch/hydra/pull/3315).

TheSaiEaranti:
Fixes #3318

Rebased onto `main` now that #3315 has landed — single commit.

## Problem

Per #3318: a Defaults List override with no earlier Group Default correctly fails on its own, but a matching command line append made the composition succeed and silently discarded the appended value:

```yaml
# config.yaml
defaults:
  - override db: sqlite
```

```
$ python app.py +db=mysql      # composed db=sqlite on main and on #3315
```

## Fix

#3315 preserved this behavior deliberately (registering the list's own override when an externally appended group default of the same key resolves), on the principle that a precedence fix shouldn't change unrelated semantics. With #3318 ruling the behavior itself invalid, that registration path is removed. The appended entry now composes without consulting the dangling override, the override is reported unused, and composition fails exactly as it does without the append:

```
In 'config': Could not override 'db'.
Did you mean to override db?
```

- The failure is identical in cause to the no-append case (`ensure_overrides_used`), so `+db=mysql` never rescues the invalid Defaults List.
- Command line **overrides** of appended groups (`+db=mysql db=postgres`) are registered before traversal and still apply — covered by the existing suites.
- Both precedence behaviors from #3315 (the #3229 ordering and the review's two-defaults-list ordering) are unaffected; their regression tests still pass.

Per review, config-sourced dangling overrides are now reported directly instead of through the command-line-oriented candidate message:

```
In '<config>': Invalid Defaults List override '<key>: <value>'.
No earlier Group Default for '<full key>' exists to override.
```

The diagnostic is byte-identical with and without a command-line append and never suggests candidates. Command-line overrides keep their existing messages. Validity is evaluated against the effective depth-first Defaults List after command-line selections apply — an earlier selection that introduces the target makes a later config override valid, while later appends never do — covered by the five-case `dangling_nested_override` matrix from review.

## Tests

- `group_override_only` + `+group1=file1` → `ConfigCompositionException`. This test **fails on #3315** (where it composed), pinning the new semantics.
- `group_override_only` with no append → same failure as before the change, guarding against the fix drifting the base case.

`pytest tests/defaults_list tests/test_compose.py` → 423 passed. Full suite excluding `test_completion` → 2056 passed.

### DIFF
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -574,16 +574,6 @@ def _create_defaults_tree_impl(
 
     _update_overrides(defaults_list, overrides, parent, interpolated_subtree)
 
-    # An externally appended group default is the one case where a group default
-    # may legally follow an override of the same group in a single defaults list.
-    # The reverse traversal below visits the appended entry before that override,
-    # so map each key to the list's last override for it, allowing the appended
-    # entry to register its override before resolving.
-    list_overrides: Dict[str, GroupDefault] = {}
-    for d in defaults_list:
-        if isinstance(d, GroupDefault) and d.is_override():
-            list_overrides[d.get_override_key()] = d
-
     def add_child(
         child_list: List[Union[InputDefault, DefaultsTreeNode]],
         new_root_: DefaultsTreeNode,
@@ -612,11 +602,6 @@ def _create_defaults_tree_impl(
                 continue
 
             d.update_parent(parent.get_group_path(), parent.get_final_package())
-
-            if isinstance(d, GroupDefault) and d.is_external_append():
-                pending = list_overrides.get(d.get_override_key())
-                if pending is not None:
-                    overrides.add_override(parent.get_config_path(), pending)
 
             if overrides.is_overridden(d):
                 assert isinstance(d, GroupDefault)
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -146,6 +146,20 @@ class Overrides:
     def ensure_overrides_used(self) -> None:
         for key, meta in self.override_metadata.items():
             if not meta.used:
+                if not meta.external_override:
+                    # A Defaults List override must target a Group Default that
+                    # precedes it in the effective depth-first Defaults List.
+                    # Later entries, such as command line appends, are not
+                    # eligible targets and are not suggested as candidates.
+                    value = self.override_choices[key]
+                    msg = (
+                        f"Invalid Defaults List override '{meta.relative_key}: {value}'."
+                        f"\nNo earlier Group Default for '{key}' exists to override."
+                    )
+                    if meta.containing_config_path is not None:
+                        msg = f"In '{meta.containing_config_path}': {msg}"
+                    raise ConfigCompositionException(msg)
+
                 group = key.split("@")[0]
                 choices = (
                     self.known_choices_per_group[group]
--- a/news/3318.bugfix
+++ b/news/3318.bugfix
@@ -1,0 +1,1 @@
+A Defaults List override that does not target an earlier Group Default now fails to compose even when a command line append introduces a matching group, instead of silently discarding the appended value.
--- a/tests/defaults_list/data/dangling_nested_override.yaml
+++ b/tests/defaults_list/data/dangling_nested_override.yaml
@@ -1,0 +1,3 @@
+defaults:
+  - group1: file1
+  - override group1/group2: file2
--- a/tests/defaults_list/test_defaults_tree.py
+++ b/tests/defaults_list/test_defaults_tree.py
@@ -215,11 +215,15 @@ def test_simple_defaults_tree_cases(
         param(
             "group_override_only",
             ["+group1=file1"],
-            DefaultsTreeNode(
-                node=ConfigDefault(path="group_override_only"),
-                children=[GroupDefault(group="group1", value="file2")],
+            raises(
+                ConfigCompositionException,
+                match=re.escape(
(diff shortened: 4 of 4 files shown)


omry (changes requested):
Thanks for calling out the error message in the description. I agree that it should be improved in this PR.

The behavior is now correct, but with `+db=mysql` the current diagnostic is confusing:

```text
Could not override 'db'.
Did you mean to override db?
```

The actual problem is the dangling `override db: sqlite` in `config`, not the command-line append. Please report config-sourced dangling overrides directly, for example:

```text
In 'config': Invalid Defaults List override 'db: sqlite'.
No earlier Group Default for 'db' exists to override.
```

The diagnostic should be identical with and without `+db=mysql`. Please add regression coverage for both cases. Candidate suggestions for command-line overrides can retain their existing behavior.

omry:
One clarification to my review: a config-sourced override cannot always fail immediately when first encountered. A command-line selection can change an earlier subtree and thereby introduce the Group Default that makes the later override valid. Please add regression coverage for this full matrix:

1. An earlier command-line selection introduces `db`; composition succeeds and the later config `override db: sqlite` applies.
2. The selected earlier subtree does not introduce `db`; composition fails with the dangling config-override error.
3. `db=mysql` by itself does not recover the dangling override, because it can only change an existing Group Default.
4. `+db=mysql` does not recover it, because the append comes after the config override. This should report the dangling config override without suggesting `db` as a candidate.
5. When an earlier selection introduces `db`, `db=mysql` succeeds and retains command-line precedence over the config's `override db: sqlite`.

The validity check therefore needs to use the effective depth-first Defaults List while distinguishing eligible earlier defaults from later command-line appends.

TheSaiEaranti:
Both points addressed in 3ca3035.

**Diagnostic.** Config-sourced unused overrides now report directly, in your suggested form:

```
In 'dangling_nested_override': Invalid Defaults List override 'group1/group2: file2'.
No earlier Group Default for 'group1/group2' exists to override.
```

The message is byte-identical with and without a command-line append — both regression tests assert the same string — and never lists candidates, so the appended group is no longer suggested. Command-line overrides keep their existing "Could not override / Did you mean" messages and the `+key=value` append hint. A side effect worth noting: the nested pre-existing cases (`group1/override_invalid*`, `experiment/error_override_without_*`) previously produced the self-contradictory "Could not override 'x'. Did you mean to override x?" — they now get the direct form as well, and their assertions are updated.

**The matrix.** All five cases are covered by the `dangling_nested_override` params (fixture: `[group1: file1, override group1/group2: file2]`, where `group1=group_item1` selects a subtree that introduces `group2`):

1. `group1=group_item1` — the selection introduces `group1/group2`; composition succeeds and the config override applies (`group2: file2`).
2. no overrides — no earlier Group Default; fails with the dangling-override error.
3. `group1/group2=file3` — fails; the command-line override cannot introduce the group (existing CLI message retained).
4. `+group1/group2=file3` — fails with the dangling-override error, no candidate suggestion.
5. `group1=group_item1 group1/group2=file3` — composition succeeds with `group2: file3`; command line takes precedence over the config's override.

On your clarification: the check indeed cannot run at first encounter, and the implementation doesn't — validity falls out of the traversal itself. An override is marked used when it applies to a Group Default during composition of the effective tree (after command-line selections resolve), and the reverse traversal visits appends before any config override registers, so appends can never mark one used. `ensure_overrides_used` then only has to classify the leftovers, which is where the new message lives. No separate validity pass was needed.

`tests/defaults_list` + `tests/test_compose.py`: 428 passed. Full suite excluding `test_completion`: 2066 passed.

omry (approved):
All good. Thanks for your help with this and for addressing the review feedback, Sai.

If you would like to chat about [Hydra](https://github.com/facebookresearch/hydra), [OmegaConf](https://github.com/omry/omegaconf), [Arbiter](https://github.com/omry/arbiter), [OmegaFlow](https://github.com/omry/omegaflow), or [Reploy](https://github.com/omry/reploy), you are welcome to join the [Hydra chat](https://hydra-framework.zulipchat.com/) and ping me there.

PR #3330, opened 2026-07-31
Fix instantiate to set allow_objects flag when config is already an OmegaConf node

### LINKED ISSUE #1865: Instantiation fails when _target_ arg interpolates to another node with _target_ key and object arg
The following example demonstrates a bug with the `hydra.utils.instantiate` function:
```yaml
# conf.yaml
vec:
  _target_: __main__.ClassB
  arg_B: ${as_myobject:abc}
pipe:
  _target_: __main__.ClassA
  arg_A: ${vec}
```
```python
# my_app.py
import hydra
import omegaconf
from pytest import raises

class MyObject:
    def __init__(self, arg):
        self.arg = arg

omegaconf.OmegaConf.register_new_resolver("as_myobject", lambda arg: MyObject(arg))

class ClassA:
    def __init__(self, arg_A):
        self.arg_A = arg_A

class ClassB:
    def __init__(self, arg_B: MyObject):
        self.arg_B = arg_B

@hydra.main(config_path=".", config_name="conf")
def main(cfg):
    instantiate(cfg)
    instantiate(cfg.vec)
    with raises(omegaconf.errors.UnsupportedValueType):
        instantiate(cfg.pipe)

if __name__ == "__main__":
    main()
```

Running the above minimal example at the command line (with `python my_app.py`):
  - `instantiate(cfg)` works as expected
  - `instantiate(cfg.vec)` works as expected
  - `instantiate(cfg.pipe)` raises an `UnsupportedValueType` exception. This is unexpected.

This issue is based on a comment [here](https://github.com/omry/omegaconf/issues/392#issuecomment-948956968) and is also related to [this](stackoverflow.com/questions/69651138) stackoverflow question. As the OP correctly pointed out, this is related to use of the `"allow_objects"` flag by the `instantiate` function.

Edit: possibly also related to #1758?

anishesg:
## Motivation

When calling `instantiate()` with a sub-node of an existing OmegaConf config (e.g., `instantiate(cfg.pipe)`), instantiation fails if the config contains interpolations to nodes with `_target_` keys that use custom resolvers returning objects. This occurs because the `allow_objects` flag was only set when converting plain dicts/structured configs to OmegaConf, but not when the config was already an OmegaConf node.

## Contributing Guidelines acknowledgment

Yes

## Design alignment

This fix addresses issue #1865, which was confirmed by maintainers and is tracked for Hydra 1.4.

## Test Plan

Added regression test `test_issue_1865_nested_instantiation_with_resolver_objects` that reproduces the exact failure scenario: instantiating a config node where an argument interpolates to another node with `_target_` and a custom resolver returning objects. The test verifies that:

1. Instantiating the full config works
2. Instantiating the interpolated node directly works  
3. Instantiating the parent node that references the interpolated node works (previously failed)

The fix ensures the `allow_objects` flag is set on the root of the config tree when an already-constructed OmegaConf node is passed to `instantiate()`, enabling custom resolvers to return object instances during interpolation resolution.

## Related Issues and PRs

Fixes #1865

### DIFF
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -517,6 +517,14 @@ def instantiate(
         or type(config) is tuple
     ):
         config = OmegaConf.structured(config, flags={"allow_objects": True})
+    elif OmegaConf.is_config(config):
+        # If config is already an OmegaConf node, ensure allow_objects is set
+        # on the root to support custom resolvers that return objects
+        root = config
+        while root._get_parent() is not None:
+            root = root._get_parent()
+        if not root._metadata.flags.get("allow_objects", False):
+            root._metadata.flags["allow_objects"] = True
 
     if OmegaConf.is_dict(config):
         return instantiate_node(
--- a/news/1865.bugfix
+++ b/news/1865.bugfix
@@ -1,0 +1,1 @@
+Fix instantiate to set allow_objects flag on config root when input is already an OmegaConf node
--- a/test_issue_1865.py
+++ b/test_issue_1865.py
@@ -1,0 +1,51 @@
+import omegaconf
+from hydra.utils import instantiate
+from omegaconf import OmegaConf
+
+class MyObject:
+    def __init__(self, arg):
+        self.arg = arg
+
+omegaconf.OmegaConf.register_new_resolver("as_myobject", lambda arg: MyObject(arg))
+
+class ClassA:
+    def __init__(self, arg_A):
+        self.arg_A = arg_A
+
+class ClassB:
+    def __init__(self, arg_B: MyObject):
+        self.arg_B = arg_B
+
+# Create config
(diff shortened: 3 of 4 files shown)


omry:
Please wait with this. there has been a major rewrite of the instantiate logic and allow_object is likely no longer needed.
do you have a concrete problem you are solving or are you just following the issue?

omry:
I verified this against current `main` after the `instantiate` rewrite, and the original #1865 reproduction now passes. I closed #1865 as fixed.

Please revert the implementation changes and reduce this PR to a simple, focused regression test covering the exact bug reported in #1865. The test should exercise the formerly failing `instantiate(cfg.pipe)` path. Please also remove the standalone reproduction and news fragment; no news fragment is needed for a test-only change.

anishesg:
closing this since #1865 is already resolved elsewhere.

PR #3350, opened 2026-08-04
Normalize slash-containing group default options in Defaults List (Issue #3222)

### LINKED ISSUE #3222: Normalize slash-containing group default options
Hydra currently accepts Defaults List entries where the selected option contains `/`, for example:

```yaml
defaults:
  - foo: bar/baz
```

This form is not documented. The documented/canonical representation is:

```yaml
defaults:
  - foo/bar: baz
```

These two forms are closely related conceptually, but today they do not behave equivalently. Treating `bar/baz` as an option path under group `foo` creates surprising relative-defaults behavior in nested configs. #3060 is an example: the accepted shorthand leads users toward nested defaults such as `nested/second`, which then compose into repeated path components like `group2/nested/nested/third`.

This shorthand is already accepted by Hydra and is used in practice, so the Hydra 1.4 direction is to normalize it rather than reject it.

Direction for Hydra 1.4:

- Normalize `foo: bar/baz` to `foo/bar: baz` early during Defaults List creation.
- Ensure nested defaults inside the selected config then behave as if the canonical form had been written originally.
- Add regression tests for the canonical form and the normalized shorthand.
- Document the supported shorthand and its equivalence to the canonical representation.
- Add a dedicated Hydra 1.3-to-1.4 migration page under `website/docs/upgrades/` for this breaking behavior change.
- Where possible, detect configs that appear to rely on the old behavior and raise a clear error that points to that migration page.

This is behavior-changing and should be resolved deliberately for Hydra 1.4.

nithin42:
Fixes #3222

### Summary
Hydra accepted Defaults List entries where the selected option contains a slash, like `foo: bar/baz`. However, this shorthand was undocumented and behaved differently from the canonical form `foo/bar: baz` during nested relative-default composition. Relative defaults nested inside `foo/bar/baz.yaml` resolved relative to the parent group `foo` instead of `foo/bar`, causing missing config paths or incorrect relative nesting.

### Root cause
In `_create_defaults_list()` of `hydra/_internal/config_repository.py`, when parsing a DictConfig default item like `foo: bar/baz`, the group name was kept as `foo` and the option value as `bar/baz`. Consequently, the parent group path of that default element in the defaults tree was set to `foo`, causing any relative child defaults to inherit `foo` as their base directory instead of `foo/bar`.

### Fix
1. **Early Normalization**: We normalize `foo: bar/baz` to group `foo/bar` and option value `baz` early in `_create_defaults_list()`. Normalization excludes strings that are `???`, contain `${`, or contain `..` as a path segment (substrings like `sub..group` are allowed and will still be normalized).
2. **Deferred Interpolation Normalization**: If a Defaults List item uses dynamic interpolation (e.g. `foo: ${prefix}/item`), we normalize it immediately once the interpolation resolves to a slash-containing value during the resolution phase.
3. **Metadata Tracking**: We save `original_group` and `normalized_shorthand` on `GroupDefault`.
4. **Migration Check**: We implement a check in `config_not_found_error()` using a `_get_old_path()` helper. If a config fails to load and any ancestor was normalized, we check if the file exists at the old location. If it does, we raise a `ConfigCompositionException` pointing to the upgrade guide.
5. **Docs & Tests**: 
   - Created a dedicated upgrade guide under `website/docs/upgrades/1.3_to_1.4/slash_in_default.md`.
   - Documented the shorthand behavior in `website/docs/advanced/defaults_list.md` and added it to the `sidebars.js` Upgrade Guide sidebar.
   - Added comprehensive integration and unit tests in `tests/defaults_list/test_slash_normalization.py`.


omry:
Hi @nithin42.
Thanks for your help.
As you have noticed, I have been busy preparing for the 1.4 release. there are many existing issues with that milestone.

In general, before sending a PR you need to have a discussion about the issue and get a buy in from a maintainer (that's me) before attempting a fix.

There are two reasons:
I may disagree with your direction, in which case:
1. you wasted your time creating the PR.
2. you wasted mine reviewing it.

Its better to get the agreement beforehand, or to pick one of the many issues already triaged for 1.4.

I will take me some time to get this this one and form an opinion.

Thanks.

omry:
To clarify: it's best to either create a new issue or to get to a consensus over an existing issue before sending a PR.

nithin42:
Understood completely, thank you for the clarification! I will make sure to start a discussion and get consensus on the issue page first before opening any future Pull Requests.

omry:
Correction to my earlier comments: I missed the reference to #3222 at the end of the PR title. This PR does address an existing issue that was explicitly triaged for Hydra 1.4, so my concern that the work lacked prior direction was misplaced.

Please add `Fixes #3222` prominently to the PR body. That makes the relationship visible during review and ensures the issue closes automatically when the PR is merged.

The overall direction is accepted. I will review the specific implementation, tests, and migration handling separately.
There is no need to get a buy-in for issues I already triaged and assigned the 1.3 milestone (like the one you picked).

nithin42:
Thank you for the update! I have updated the PR description to include `Fixes #3222` at the very top as requested. 

I also synced the branch with `upstream/main` to resolve the merge conflict in the documentation files. 

Please let me know if you need any changes during your review of the code and tests. Thanks again!

omry (changes requested):
Thanks for working on this. The overall direction matches https://github.com/facebookresearch/hydra/issues/3222, but the implementation needs several changes before it is ready:

1. The migration detector crashes during normal Hydra composition. With `prepend_hydra=True`, `_get_old_path()` reaches `VirtualRoot` and calls `get_group_path()`, which raises `NotImplementedError`. The current test hides this by using `prepend_hydra=False`. Please test this through the normal composition path. The legacy-path calculation also needs to handle the virtual root and restore tree state safely if it temporarily mutates nodes.

2. The normalization condition is too broad and incomplete:

   - Names containing `..` as a substring are valid; only an exact `..` path segment is parent traversal. For example, `foo: sub..group/item` should still be normalized.
   - Deferred Defaults List interpolations must also be normalized after resolution. If `foo: ${prefix}/item` resolves to `foo: sub/item`, it should then be normalized internally to `foo/sub: item`. The current implementation resolves the value but does not perform this second normalization step.

3. The tests do not yet demonstrate the behavior requested by the issue. Please add canonical and shorthand cases, compose both end-to-end, and verify equivalent Defaults Trees and output—including a selected config with nested relative defaults. The tests should use the normal Hydra composition path and cover the package and override-key consequences, legal `..` substrings, and deferred interpolation.

4. The migration guide needs to explain the full breaking surface. Normalizing `foo: bar/baz` to `foo/bar: baz` keeps the config path but changes the default package from `foo` to `foo.bar` and the override key from `foo` to `foo/bar`. This can move composed output and require command-line overrides to change.

5. Please document the supported shorthand and its equivalence to the canonical representation in the regular Defaults List documentation, as required by the issue, and add the migration page to the 1.3-to-1.4 sidebar.

6. The changed Python files currently fail Ruff and formatting checks.

Finally, please rebase against the current `main` branch, resolve the conflict, and add:

`Fixes https://github.com/facebookresearch/hydra/issues/3222`

to the PR body.

Once these items are addressed, I will review the updated implementation.

nithin42:
Thanks for the detailed review @omry! I have addressed all 6 points and pushed the updates: fixed the virtual root crash, refined segment checks (allowing `sub..group`), normalized deferred interpolations, added integration tests using standard composition, updated the docs/sidebar, and formatted all files.

Please let me know if you need anything else!

omry:
Thanks, the substantive issues are now addressed. Three items remain:

1. `pyrefly` reports two `InputDefault.group` errors in `defaults_list.py`. Please narrow those values to `GroupDefault` before accessing `.group`.
2. `ruff format --check` still wants to reformat `defaults_list.py`; please run the repository formatter.
3. Please update the PR description: it still says normalization excludes any string containing `..`, and it does not mention that deferred Defaults List interpolations are normalized after resolution.

The focused tests and the full Defaults List test suite pass.

nithin42:
Thanks @omry! I have fixed all three remaining items (Pyre type errors, Ruff formatting, and the PR description updates). Let me know if everything looks good!

omry:
Thanks. I approved the workflow runs so CI can evaluate the current update. The substantive behavior changes from the previous review are now addressed.

Three items remain:

1. `_has_normalized_ancestor()` still returns `Tuple[bool, Optional[InputDefault]]`, but its caller accesses `norm_node.group`. Please return `Optional[GroupDefault]` directly and remove the redundant boolean.
2. The documentation heading says the shorthand is “New in 1.4.” Hydra already accepted this syntax; Hydra 1.4 only changes its normalization semantics. Please adjust the wording accordingly.
3. Please rebase onto the current `main`. The branch is four commits behind and overlaps the recently updated `breaking_changes.md`.

To run the same core lint checks locally, from the repository root use:

```bash
nox -s lint-core
```

Please make sure that passes before pushing the revision. Once the updates are pushed and CI is clean, I will do the final review.

nithin42:
Thanks @omry! I have rebased the branch onto main, updated the documentation headers, and added the type assertions to solve the Pyre/pyrefly warnings. Let me know if everything looks good now!

omry:
Thank you for taking this on and for working through the review feedback. The updated implementation, tests, and migration documentation look good. Merging now.

Issue #3353, opened 2026-08-06
`instantiate()` call-site overrides must be visible to dependent interpolations


omry:
## Problem

Hydra 1.4 changed `hydra.utils.instantiate()` to keep call-site arguments separate from the input configuration. As a result, configured interpolations resolve against the original values instead of the effective values passed at the call site.

```python
from hydra.utils import UNSAFE_ALLOW_ALL_TARGETS, instantiate

cfg = {
    "_target_": "builtins.dict",
    "b": 200,
    "c": "${b}",
}

result = instantiate(
    cfg,
    b=99,
    _target_whitelist_=UNSAFE_ALLOW_ALL_TARGETS,
)
```

Hydra 1.3 produced: 

```python
{"b": 99, "c": 99}
```

Current Hydra 1.4 development builds produce:

```python
{"b": 99, "c": 200}
```

This regression was introduced by https://github.com/facebookresearch/hydra/pull/3302. Existing interpolation tests were changed from expecting `c=99` to expecting `c=200`, and the new result was documented as a compatibility change. The separate runtime overlay is useful internally, but it must still behave as an effective configuration for interpolation resolution.

## Expected behavior

Call-site values must participate in the effective configuration being instantiated. Any configured interpolation that survives the override must resolve against those effective values. The input configuration itself must remain unchanged.

For the example above, `c` must resolve to `99`.

The fix should retain the useful properties of the Hydra 1.4 instantiate rewrite:

- Do not mutate the input configuration.
- Do not eagerly resolve the full configuration tree.
- Do not process a configured value that has been completely replaced by a call-site value.
- Preserve lazy interpolation and parent context where applicable.

Coverage should include normal and partial instantiation, sibling and nested interpolation dependencies, and verification that the source configuration remains unchanged.

## Scope

This is independent of https://github.com/facebookresearch/hydra/issues/2350 and https://github.com/facebookresearch/hydra/pull/3352, which decide whether a dictionary-valued call-site argument replaces or patches the configured dictionary.

If an entire dictionary is replaced, interpolations inside the discarded dictionary no longer exist. However, any surviving interpolation elsewhere that references the replaced dictionary must resolve against the replacement value.

This regression exists only on the Hydra 1.4 development line and must be fixed before the stable 1.4 release.

PR #3310, opened 2026-07-24
Add non-target recursive instantiation coverage

omry:
## Summary

- Add an explicit regression test for `_recursive_: false` on a non-target config node.
- Add the corresponding feature news fragment.

## Context

The behavior requested by #2577 is already present on `main` after #3302 rewrote instantiation. This PR adds direct coverage for the motivating target → non-target → target shape and records the feature for the 1.4 release.

## Test plan

- `.venv/bin/pytest tests/instantiate/test_instantiate.py -k non_target_node_obeys_recursive_false`
- `PATH=/home/omry/dev/hydra/.venv/bin:$PATH .venv/bin/pytest`
- `.venv/bin/nox -s lint`

PR #3355, opened 2026-08-06
Treat a call-site ??? as a literal value

### LINKED ISSUE #3354: `instantiate()` must treat call-site `???` as a literal value
## Problem

Hydra 1.4 treats arguments passed directly to `hydra.utils.instantiate()` as ordinary Python values rather than configuration fragments. A call-site string such as `"${value}"` is therefore passed literally instead of being interpreted as an OmegaConf interpolation.

The same rule must apply to `"???"`, but partial instantiation still special-cases that value as an OmegaConf missing marker and silently omits the argument:

```python

cfg = {"_target_": "builtins.dict"}

assert instantiate(cfg, value="???") == {"value": "???"}

factory = instantiate(cfg, value="???", _partial_=True)
assert factory() == {"value": "???"}  # currently produces {}
```

This happens because `_is_missing_parameter()` explicitly interprets a call-site override equal to `"???"` as missing. That retains behavior from when call-site arguments were merged into an OmegaConf configuration, but contradicts the Hydra 1.4 runtime-argument model.

## Expected behavior

Values supplied directly at the call site are regular Python values:

- `instantiate(cfg, value="???")` passes the literal string `"???"`.
- `instantiate(cfg, value="???", _partial_=True)` captures and later passes the literal string `"???"`.
- Hydra does not resolve, merge, or otherwise interpret the call-site value as a configuration fragment.

The behavior of `???` originating in the configuration does not change:

```python
cfg = {
    "_target_": "builtins.dict",
    "value": "???",
}

instantiate(cfg)  # raises MissingMandatoryValue
```

Configured missing values may still be left unbound during partial instantiation according to the existing partial behavior.

## Scope

Remove the call-site override special case from missing-parameter detection and add coverage for both normal and partial instantiation. Preserve the existing behavior for missing values originating in Hydra/OmegaConf configuration.

This issue applies to the Hydra 1.4 development line and should be resolved before the stable 1.4 release.

adarshsm:
## Motivation

Fixes #3354.

Call-site arguments are runtime values in Hydra 1.4, but partial instantiation still read an
argument equal to `"???"` as an OmegaConf missing marker and silently omitted it:

```python
cfg = {"_target_": "builtins.dict"}

instantiate(cfg, value="???")                    # {'value': '???'} — already correct
instantiate(cfg, value="???", _partial_=True)()  # {} — should be {'value': '???'}
```

Confirmed on `main` at `6978bcf593`.

`_is_missing_parameter()` returned `True` for a call-site override equal to `"???"`, which
retained behavior from when call-site arguments were merged into an OmegaConf
configuration. Its only call site is the `is_partial` branch in `instantiate_node`, which
is why normal instantiation was already correct and only partial instantiation dropped the
argument. A call-site argument is never missing, so that branch now returns `False`.

`???` originating in the configuration is unchanged: it still raises
`MissingMandatoryValue` on normal instantiation, and still stays unbound during partial
instantiation. `test_instantiate_with_missing` and `test_partial_with_missing` both pass
unmodified.

### One existing expectation changed

`test_class_instantiate[class+override+partial1+missing]` passed `{"a": "???", "_partial_":
True}` and expected `partial(AClass, b=20, c=30)`, with `a` dropped. That parametrization
is the behavior this issue asks to remove, so it now expects `partial(AClass, a="???",
b=20, c=30)`.

### On the news fragment

I did not add one, since the regression exists only on the unreleased 1.4 development line
and nothing changes for anyone on a shipped release — the 1.4 call-site model is already
described by `3216.api_change`. This follows #3302, #3310, #3316 and #3325, which made
comparable development-line instantiate corrections without a fragment. Happy to add one if
you would rather have it.

No documentation change either: nothing under `website/docs/` described the previous
call-site `???` behavior.

Yes.

Yes — #3354 specifies both the expected behavior and the scope ("Remove the call-site
override special case from missing-parameter detection and add coverage for both normal and
partial instantiation"). This PR implements exactly that and nothing more.

## Test Plan

No special setup beyond a normal core checkout.

Two new tests in `tests/instantiate/test_instantiate.py`, placed next to
`test_partial_with_missing`, their configured-`???` counterpart:

- `test_partial_callsite_missing_string_is_literal` — the regression. Confirmed failing
  before the change with `assert {} == {'value': '???'}`.
- `test_callsite_missing_string_is_literal` — the normal-instantiation half, which was
  already correct; included per the coverage scope in the issue.

```
pytest tests/instantiate/                  # 483 passed (481 before + 2 new)
pytest --ignore=tests/test_completion.py   # 2162 passed, 0 failed
```

Both assertions in the issue body were also run directly and pass.

`tests/test_completion.py` has 130 failures in my environment (macOS zsh/bash shell
integration). They are unrelated: the list of failing test ids is identical to the one
pristine `main` produces here.

Lint clean: `ruff format --check .`, `ruff check .`, `yamllint --strict .`,
`bandit -ll -r .`. `pyrefly check --config pyproject.toml` reports one error in
`build_helpers/build_helpers.py`, which is untouched here and reports identically on
pristine `main`.

## Related Issues and PRs

- Fixes #3354
- Sibling regression #3353 is not addressed here.
- Independent of #3352, per the scope note in #3353. Both touch
  `hydra/_internal/instantiate/_instantiate2.py` but different functions
  (`_is_missing_parameter` here, `_instantiate_effective_value` there), so they should not
  conflict meaningfully in either merge order.

### DIFF
--- a/hydra/_internal/instantiate/_instantiate2.py
+++ b/hydra/_internal/instantiate/_instantiate2.py
@@ -618,8 +618,10 @@ def _get_effective_control(
 def _is_missing_parameter(
     node: Any, overrides: Optional[ConfigOverlay], key: str
 ) -> bool:
+    # A call-site argument is a runtime value, so "???" is a plain string
+    # rather than a missing marker.
     if overrides is not None and key in overrides:
-        return isinstance(overrides[key], str) and overrides[key] == "???"
+        return False
     return OmegaConf.is_missing(node, key)
 
 
--- a/tests/instantiate/test_instantiate.py
+++ b/tests/instantiate/test_instantiate.py
@@ -205,10 +205,11 @@ def register_test_resolver(name: str, value: Any) -> Any:
             partial(AClass, a=10, b=20, c=30),
             id="class+override+partial1",
         ),
+        # "???" passed at the call-site is a literal string, not a missing value
         param(
             {"_target_": "tests.instantiate.AClass", "b": 20, "c": 30},
             {"a": "???", "_partial_": True},
-            partial(AClass, b=20, c=30),
+            partial(AClass, a="???", b=20, c=30),
             id="class+override+partial1+missing",
         ),
         param(
@@ -465,6 +466,33 @@ def test_partial_with_missing(instantiate_func: Any) -> Any:
     obj = partial_obj(a=10)
     assert partial_equal(obj, AClass(a=10, b=20, c=30))
     assert str(config) == original_config_str
+
+
+def test_callsite_missing_string_is_literal(instantiate_func: Any) -> Any:
+    """
+    A call-site argument is a runtime value, so "???" is passed to the target
+    as a string instead of being read as an OmegaConf missing marker.
+    """
+    result = instantiate_func(
+        {"_target_": "tests.instantiate.ArgsClass"},
+        value="???",
+    )
+
+    assert result.kwargs == {"value": "???"}
+
+
+def test_partial_callsite_missing_string_is_literal(instantiate_func: Any) -> Any:
+    """
+    In 3354, partial instantiation read a call-site "???" as a missing value
+    and silently omitted the argument.
+    """
+    factory = instantiate_func(
+        {"_target_": "tests.instantiate.ArgsClass"},
+        value="???",
+        _partial_=True,
+    )
+
+    assert factory().kwargs == {"value": "???"}
 
 
 def test_instantiate_with_missing(instantiate_func: Any) -> Any:


adarshsm:
Moved to draft while @omry lands the Hydra 1.4 instantiate work (context in #3352). Not requesting review — happy to close this or rebase it once that work is done, whichever he prefers.

omry:
Thanks for implementing the behavior requested by https://github.com/facebookresearch/hydra/issues/3354.

The design changed during the broader `instantiate()` review: https://github.com/facebookresearch/hydra/pull/3359 now rejects `???` and interpolation syntax in plain Python call-site overrides instead of treating them as literal runtime values. That change has landed, so this PR is no longer applicable. Closing as superseded.

PR #3363, opened 2026-08-08
Avoid passing default object_store_memory when connecting to existing Ray cluster (Issue #2893)

### LINKED ISSUE #2893: [Bug] When connecting to an existing cluster, object_store_memory must not be provided
# 🐛 Bug
## Description
With a recent version of ray, launching jobs with `hydra-ray-launcher` on an existing ray cluster rise an error:

`ValueError: When connecting to an existing cluster, object_store_memory must not be provided.`

Commenting out the following line fix the problem:

https://github.com/facebookresearch/hydra/blob/2e682d84e789d82dd11ab1f329f2dd1966fa6b54/plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core.py#L36

## System information
- **Hydra Version** :  1.3.2
- **hydra-ray-launcher** :  1.2.1
- **Python version** : 3.10.14
- **Virtual environment type and version** : Micromamba 1.5.8
- **Operating system** : Ubuntu 22.04
- **Ray[default]** :  2.11.0

nithin42:
Fixes #2893

### Summary
The Hydra Ray launcher was injecting `object_store_memory=78643200` into every `ray.init()` call as an old CI workaround. Ray rejects resource configuration settings when connecting to an existing cluster, so this PR removes the automatic production default entirely. Users can still configure `hydra.launcher.ray.init.object_store_memory` explicitly when needed.

### Changes
1. **Core fix**: Removed automatic `object_store_memory` injection from `plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core.py`.
2. **Focused coverage**: Added a unit test covering local and existing-cluster addresses, both with and without an explicitly configured `object_store_memory`, while verifying that the launch path reaches `start_ray()`.
3. **Integration tests**: Configured the historical memory value explicitly where the test environment requires it.
4. **Cleanup**: Removed unnecessary Windows Ray mocks.
5. **News fragment**: Added `plugins/hydra_ray_launcher/news/2893.bugfix`.

### DIFF
--- a/plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core.py
+++ b/plugins/hydra_ray_launcher/hydra_plugins/hydra_ray_launcher/_core.py
@@ -32,12 +32,10 @@ def launch(
         f"sweep output dir: {sweep_dir}"
     )
 
-    # Avoid allocating too little memory in CI https://github.com/ray-project/ray/issues/11966#issuecomment-1318100747
     ray_init = cast(
         DictConfig,
         OmegaConf.create(OmegaConf.to_container(launcher.ray_cfg.init, resolve=True)),
     )
-    ray_init.setdefault("object_store_memory", 78643200)
     start_ray(ray_init)
 
     runs = []
--- a/plugins/hydra_ray_launcher/news/2893.bugfix
+++ b/plugins/hydra_ray_launcher/news/2893.bugfix
@@ -1,0 +1,1 @@
+Avoid default object_store_memory when connecting to an existing Ray cluster
--- a/plugins/hydra_ray_launcher/tests/test_ray_launcher.py
+++ b/plugins/hydra_ray_launcher/tests/test_ray_launcher.py
@@ -32,7 +32,10 @@ def test_discovery() -> None:
 
 
 @mark.skipif(sys.platform.startswith("win"), reason=win_msg)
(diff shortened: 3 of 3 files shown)


omry (changes requested):
Thanks for working on this. The issue is valid: it affects the supported local `ray` launcher, not the best-effort `ray_aws` launcher.

Looking at the root cause, the current condition is not quite right. `object_store_memory=78643200` was added as an old CI workaround and should not be injected into production calls at all. Ray may connect to an existing cluster with `address=None`, while `address="local"` starts a new cluster, so checking whether the address is `None` does not reliably distinguish the two cases.

Please remove the automatic `object_store_memory` default entirely. If Hydra's tests still require that value, configure it explicitly in the tests or CI. Focused unit coverage is sufficient here: verify that existing-cluster addresses do not receive an injected value, an explicitly configured `object_store_memory` remains untouched, and the ordinary local-launch path still reaches `start_ray()`.

Also remove the Windows Ray mocks. The Ray launcher is excluded from Windows plugin testing based on its package classifiers, and its integration tests already skip Windows. This should also eliminate the CodeQL unused-import finding.

omry (approved):
Thank you for addressing the review feedback and adding focused coverage for the local and existing-cluster paths. The final change cleanly removes the old production default while preserving explicit configuration. Looks good!

PR #3365, opened 2026-08-09
Reject backslashes in Defaults List paths

### LINKED ISSUE #3362: Reject backslashes in Defaults List paths
## Problem

Hydra uses `/` as the config-group separator in Defaults List paths, but backslashes are not explicitly rejected.

This creates platform-dependent behavior. `FileConfigSource` passes paths through operating-system path handling, so Windows can interpret `\` as a filesystem separator while POSIX and `ConfigStore` treat it as a literal character. A path such as `..\secret` can also bypass validation that checks parent-traversal segments by splitting only on `/`.

## Expected behavior

Reject `\` in Defaults List config group and option paths, including values produced by interpolation, with a clear error directing users to use `/`.

Do not normalize `\` into `/`; `/` should remain Hydra's only supported config-group separator.

Add platform-independent validation tests covering ordinary backslash separators and `..\` traversal.

Related to https://github.com/facebookresearch/hydra/issues/3202 and discovered while reviewing https://github.com/facebookresearch/hydra/pull/3350.

nileshpatil6:
Closes #3362

Backslashes in Defaults List paths were never rejected, so behavior depended on the platform: `FileConfigSource` hands paths to OS path handling, where Windows treats `\` as a separator while POSIX and `ConfigStore` treat it as a literal character. A path like `..\secret` also slipped past the parent traversal check, which splits on `/` only, so the `..` never showed up as its own segment.

This rejects `\` anywhere in a Defaults List config, config group, or config option path with a `ConfigCompositionException` pointing users at `/`. No normalization, per the issue: `/` stays the only supported separator.

Implementation: the check lives in `_check_parent_traversal`, which I renamed to `_validate_paths` since it now validates two things. That function is already called in the three right places, including on resolved deferred interpolations, so interpolated values are covered without new plumbing. The backslash check runs before the traversal check so `..\secret` reports the actual problem.

Tests cover config paths, group paths, single options, options lists, an interpolated value with a literal backslash, and the three command line forms (override, append, append list), which I verified all flow into the same validation. Also verified the pre-change behavior on this branch's parent: the first three error configs composed without any error at the defaults-list level, which is exactly the silent pass-through the issue describes.

Ran `tests/defaults_list` (136 passed), `tests/test_config_loader.py` + `tests/test_compose.py` (223 passed), ruff check/format, yamllint on the new data files, and pyrefly on `defaults_list.py`, all clean. News fragment and the defaults list doc's separator note included.

---

Update: @adarshsm independently built the same fix in #3366 and closed it in favor of this PR. I folded in the three gaps they flagged (4ccac8f): a `..\secret` case for bare config paths, which exercises the `ConfigDefault` branch; a full-message test asserting the `In <config>:` location prefix; and a bullet on the 1.3 to 1.4 breaking changes page next to the parent traversal one. Their PR also documents why a resolution-produced backslash test is not constructible: Defaults List interpolation only substitutes known group choices, which are themselves validated first. The check stays wired at the deferred-resolution call site as defense in depth.

### DIFF
--- a/hydra/_internal/defaults_list.py
+++ b/hydra/_internal/defaults_list.py
@@ -411,7 +411,7 @@ def _resolve_deferred_interpolations(
                 # Another deferred subtree may provide the missing choice.
                 continue
 
-            _check_parent_traversal(candidate, tree.node)
+            _validate_paths(candidate, tree.node)
             candidate.update_parent(
                 tree.node.get_group_path(),
                 tree.node.get_final_package(),
@@ -452,7 +452,7 @@ def _resolve_deferred_interpolations(
     fail_on_unresolved(root)
 
 
-def _check_parent_traversal(default: InputDefault, parent: InputDefault) -> None:
+def _validate_paths(default: InputDefault, parent: InputDefault) -> None:
     if isinstance(default, ConfigDefault):
         assert default.path is not None
         paths = [("config", default.path)]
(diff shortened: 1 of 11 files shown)


omry (approved):
Thank you for the careful work here. The centralized validation, focused platform-independent coverage, documentation, and news fragment make the behavior and migration clear. I also appreciate you incorporating the useful edge cases and documentation feedback from #3366. This looks good to merge!

PR #3381, opened 2026-08-12
Show project homepages in the Hydra Landscape

### LINKED ISSUE #3375: Show project homepages in the Hydra Landscape
The Hydra Landscape currently links each project only to its repository (or relevant repository subdirectory).

Capture a project homepage when one is available, in addition to the repository URL, and show both links in the Landscape UI.

- Keep the repository link for every entry.
- Add an optional homepage URL to the generated Landscape data.
- Populate it from project metadata when available.
- Show separate **Homepage** and **Repository** links when both exist.
- Continue showing only the repository link when no homepage is available.
- Update validation and tests for the optional field.

adarshsm:
Closes #3375.

Landscape entries linked only to a repository. A listing may now also carry an optional `homepage`, shown beside the repository link; entries without one render exactly as before.

## What changed

| Stage | Change |
| --- | --- |
| `refresh_landscape_metadata.py` | Requests `homepageUrl` and records it verbatim apart from surrounding whitespace, so an unset homepage (GitHub returns `null` **or** `""`) becomes `None`. |
| `build_landscape_review.py` | Reports the homepage in the review queue — the point where a maintainer decides whether to promote it. |
| `build_public_landscape.py` | Accepts optional `listing.homepage`, requires HTTPS, rejects one that merely repeats `listing.url`, and omits the key entirely when absent. |
| `validate-landscape.mjs` | Tolerates the optional field and enforces the same two rules on generated data. |
| `LandscapeDirectory` | Renders a **Homepage** link next to the repository link when one exists. |
| `data/decisions.json` | Records the homepage GitHub reports for 61 of the 104 included projects. |

The repository link is untouched and still present on every card.

## Verification

- `PYTHONPATH=tools/landscape pytest tools/landscape` → **95 passed**. Reverting the three tool changes makes **10 of the new tests fail**, so they pin the new behaviour rather than passing vacuously.
- `npm run validate:landscape` passes. I also confirmed the new validator rules actually bite by mutating the generated file: a `http://` homepage, a whitespace-only homepage, and a homepage equal to `url` each fail with the intended message; an unknown field is still rejected.
- `npm run build` succeeds and the built `docs/landscape/index.html` contains exactly **61** `Homepage` links, matching the 61 populated entries.
- Re-running the generator reproduces `landscape.json` byte-for-byte, and the `decisions.json` diff contains **only** added `"homepage"` lines.

## Two curation calls that are yours, not mine

**1. Ten of the 61 homepages are publications rather than project sites.** These are what the projects themselves set on GitHub, so I recorded them faithfully, but a reader clicking "Homepage" may not expect a paper:

`HazyResearch/hyena-dna`, `PRIME-RL/TTRL`, `SWivid/F5-TTS`, `SysCV/sam-pt`, `aiming-lab/SkillRL`, `pdebench/PDEBench` (arXiv) · `eloialonso/iris` (OpenReview) · `microsoft/mattergen` (Nature) · `mir-group/nequip` (RSC) · `langfengQ/verl-agent` (HF papers). `DBD-research-group/BirdSet` points at a HF dataset.

Dropping them is a one-line filter in the ledger if you prefer. Note `mir-group/nequip`'s URL returns 403 to automated probes — publisher bot protection, the page is real.

**2. Three repositories declare an HTTP-only homepage and are therefore not included:** `dssg/aequitas` and `mees/calvin` have no working HTTPS at all, so they are omitted rather than silently upgraded. `hydra-ecosystem/hydra` declares `http://hydra.cc/`, which does serve HTTPS, so it is recorded as `https://hydra.cc/`.

## Smaller decisions worth a glance

- **`schemaVersion` stays at 2.** The field is purely additive and the only consumer ships in this repo. Happy to bump it if you would rather every shape change be visible.
- **Homepages are not required to be unique**, unlike `name` and `url`. Two listings from one monorepo can legitimately share a homepage, so I deliberately left them out of the uniqueness sets.
- **The card keeps the repository slug as its repository link** and adds a labelled `Homepage` beside it, which is the smallest change satisfying every bullet in the issue. If you would rather see two explicitly labelled `Homepage` / `Repository` links instead of the slug, that is a small edit to `LandscapeCard`.


omry:
closing as stale.

PR #3399, opened 2026-08-21
Fix nested Compose API initialization

### LINKED ISSUE #3364: Make Compose API initialization contexts nestable and self-contained
## Problem

The Compose API initializers serve two roles:

```python
initialize_config_dir(...)        # persistent global initialization
```

and:

```python
with initialize_config_dir(...):  # scoped initialization
    cfg = compose(...)
```

This applies equally to:

- `initialize`
- `initialize_config_module`
- `initialize_config_dir`

All three initialize Hydra in `__init__`, save the previous `GlobalHydra`, use a no-op `__enter__`, and restore the saved state in `__exit__`.

Because `GlobalHydra.initialize()` rejects an already initialized singleton, the scoped form fails before entering its context:

```python
with initialize_config_dir(config_dir=first):
    with initialize_config_dir(config_dir=second):
        ...
```

The error instructs callers to clear `GlobalHydra` themselves. Doing so leaks Hydra's global-state implementation into application code and can permanently destroy state owned by an enclosing Hydra application.

A current real-world example is [rLLM PR #805](https://github.com/rllm-org/rllm/pull/805), which clears `GlobalHydra` before and after scoped composition.

## Expected behavior

Scoped Compose API initialization should:

1. Work when Hydra is already initialized.
2. Temporarily install its own Hydra state.
3. Restore the exact previous state on normal exit and exceptions.
4. Support repeated and nested composition without callers importing or clearing `GlobalHydra`.

```python
with initialize_config_dir(config_dir=first):
    first_cfg = compose(config_name="config")

    with initialize_config_dir(config_dir=second):
        second_cfg = compose(config_name="config")

    # The first context is active again.
    first_cfg_again = compose(config_name="config")
```

## Design constraint

The current dual function-call/context-manager semantics are the root problem: initialization happens before Python reveals whether the returned object will be used as a context manager.

The fix may require separating persistent global initialization from scoped initialization. Do not preserve the current API shape by requiring callers to manipulate `GlobalHydra`.

## Acceptance criteria

- All three scoped initializer variants support an already initialized Hydra.
- Nested contexts restore the enclosing context.
- State is restored when initialization or composition raises.
- Sequential scoped calls require no explicit cleanup.
- Persistent/global initialization remains available through an explicit supported path.
- Existing function-style behavior is either preserved or given a documented migration path.
- Tests cover uninitialized, initialized, nested, sequential, and exceptional cases.
- Compose API documentation is updated.
- Add a user-facing news fragment and migration guidance if the API changes.

tandede:
## Motivation

Compose API initializers currently serve both persistent method calls and context-manager calls. Because initialization happens in the constructor, the existing context form cannot distinguish a nested scope before `__enter__` and therefore rejects an already initialized `GlobalHydra`.

This change adds explicit `.scoped()` entry points to all three Compose API initializers. A scoped initializer temporarily installs its own Hydra state and restores the exact previous singleton on normal exit, composition errors, and initialization errors. Existing method-style calls keep their original behavior: the first call initializes Hydra persistently, while repeated calls raise `ValueError` without replacing the existing instance or deep-copying initialized state.

The documentation describes the explicit scoped form, and the news fragment records the behavior change.

### Have you read the Contributing Guidelines on pull requests?

Yes.

Issue #3364, opened by the maintainer, defines the expected behavior, design constraint, and acceptance criteria for separating persistent and scoped initialization.

## Test Plan

- `python -m pytest -q tests/test_compose.py` (113 passed)
- `python -m pytest -q tests --ignore=tests/test_completion.py` (2243 passed)
- `PATH="$PWD/.venv/bin:$PATH" python -m pytest -q tests/test_completion.py` (651 passed, 73 skipped, 17 xfailed)
- `pre-commit run --files hydra/initialize.py tests/test_compose.py website/docs/advanced/compose_api.md news/3364.bugfix`
- `ruff check hydra tests` and `ruff format --check hydra tests`
- `pyrefly check hydra/initialize.py tests/test_compose.py`
- `towncrier build --draft`

Regression coverage includes all three scoped variants over existing state, nested and sequential contexts, initialization and composition failures, unchanged method-style re-initialization errors, and restoration after method-style construction failures.

## Related Issues and PRs

Closes #3364.

### DIFF
--- a/hydra/initialize.py
+++ b/hydra/initialize.py
@@ -16,15 +16,29 @@ from hydra.errors import HydraException
 
 
 def get_gh_backup() -> Any:
-    if GlobalHydra in Singleton._instances:
-        return copy.deepcopy(Singleton._instances[GlobalHydra])
-    return None
+    gh = Singleton._instances.get(GlobalHydra)
+    if gh is None:
+        return None
+    assert isinstance(gh, GlobalHydra)
+    if not gh.is_initialized():
+        return copy.deepcopy(gh)
+    return gh
 
 
 def restore_gh_from_backup(_gh_backup: Any) -> Any:
     Singleton._instances.pop(GlobalHydra, None)
     if _gh_backup is not None:
         Singleton._instances[GlobalHydra] = _gh_backup
+
+
+def _initialize_hydra(create_hydra: Callable[[], Any]) -> Any:
+    gh_backup = get_gh_backup()
+    try:
+        create_hydra()
+    except BaseException:
+        restore_gh_from_backup(gh_backup)
+        raise
+    return gh_backup
 
 
 class _ScopedInitializer:
@@ -79,12 +93,13 @@ class initialize:
                 calling_file=calling_file, calling_module=calling_module
             )
 
-        self._gh_backup = get_gh_backup()
-        Hydra.create_main_hydra_file_or_module(
-            calling_file=calling_file,
-            calling_module=calling_module,
-            config_path=config_path,
-            job_name=job_name,
+        self._gh_backup = _initialize_hydra(
+            lambda: Hydra.create_main_hydra_file_or_module(
+                calling_file=calling_file,
+                calling_module=calling_module,
+                config_path=config_path,
+                job_name=job_name,
+            )
         )
 
     @classmethod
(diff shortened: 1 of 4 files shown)


omry:
Hi, thanks for working on it.
After taking a closer look, I decided to do this one myself.
It's a sensitive area of global mutable state that is consumed by various downstream consumers (plugins, user code etc).

a few issues I identified:

1. I do not want to change the signature of the initialization methods. Changing such a core Hydra API should be done very carefully and not in a commit by a new contributor. #3364 asked for the existing form to nest, and that still raises here, so `scoped()` ends up as a second way to do the same thing.

2. there are additional singletons that are leaking. `Hydra.__init__` calls `JobRuntime().set("name", task_name)` and nothing restores it, so an inner scope leaves its job name on the outer app. Only `GlobalHydra` and `VersionBase` are saved, 2 of the 7.

a proper fix would likely involve some design changes for how Hydra singletons are implemented.

Feel free to pick up other issues, happy to have you.

PR #3432, opened 2026-09-01
feat: introduce NSGA-III based sampler config

himkt:
## Motivation

Provide the interface for Optuna NSGA-III GA sampler on hydra plugin.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/hydra-ecosystem/hydra/blob/main/CONTRIBUTING.md)?

Yes

### Link

🔗. https://github.com/hydra-ecosystem/hydra/issues/3426

## Test Plan

Running NSGA-III based optimization on multi-objective test.

### DIFF
--- a/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
+++ b/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
@@ -107,7 +107,7 @@ class NSGAIIISamplerConfig(SamplerConfig):
     https://optuna.readthedocs.io/en/v4.9.0/reference/samplers/generated/optuna.samplers.NSGAIIISampler.html
     """
 
-    _target_: str = "optuna.samplers.NSGAIIISampler"
+    _target_: str = "hydra_plugins.hydra_optuna_sweeper._impl.create_nsgaiii_sampler"
     seed: Optional[int] = None
 
     population_size: int = 50
--- a/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
+++ b/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
@@ -102,6 +102,25 @@ class NSGAIISamplerConfig(SamplerConfig):
 
 
 @dataclass
+class NSGAIIISamplerConfig(SamplerConfig):
+    """
+    https://optuna.readthedocs.io/en/v4.9.0/reference/samplers/generated/optuna.samplers.NSGAIIISampler.html
+    """
+
+    _target_: str = "optuna.samplers.NSGAIIISampler"
+    seed: Optional[int] = None
+
+    population_size: int = 50
+    mutation_prob: Optional[float] = None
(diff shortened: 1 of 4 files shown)


plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/config.py
```diff
@@ -101,6 +101,25 @@ class NSGAIISamplerConfig(SamplerConfig):
...
+    _target_: str = "optuna.samplers.NSGAIIISampler"
+    seed: Optional[int] = None
+
+    population_size: int = 50
+    mutation_prob: Optional[float] = None
+    crossover: Optional[Any] = None
+    crossover_prob: float = 0.9
+    swapping_prob: float = 0.5
+    constraints_func: Optional[Any] = None
+    reference_points: Optional[List[List[float]]] = None
```
himkt, line 128: `reference_points` is `np.ndarray` on the Optuna but it works with list. In my understanding, it is not straightforward to take non-standard Python objects so I tried `List[List[float]]` here. There is room for a discussion.

https://github.com/hydra-ecosystem/hydra/issues/3426#issuecomment-5489558498

I also filed the issue on Optuna https://github.com/optuna/optuna/issues/6836.

omry (changes requested):
Thanks. I looked more closely at Optuna's contract and implementation.

Optuna documents `reference_points` as a two-dimensional `numpy.ndarray`. A
Python list happens to work today because the implementation stores the value
unchanged and later uses NumPy operations that accept array-like inputs, but
that behavior is not part of the documented API. This is also what
[optuna/optuna#6836](https://github.com/optuna/optuna/issues/6836) proposes to
formalize.

Since Optuna already has NumPy as a required dependency, I think the Hydra
plugin should expose `reference_points` as a YAML-friendly nested list and
convert it to an array before calling Optuna. We can intercept construction by
changing this sampler config's `_target_` to a small factory. For example:

```python
def create_nsgaiii_sampler(
    reference_points: Optional[List[List[float]]] = None,
    **kwargs: Any,
) -> optuna.samplers.NSGAIIISampler:
    if reference_points is not None:
        reference_points = np.asarray(reference_points, dtype=float)
    return optuna.samplers.NSGAIIISampler(
        reference_points=reference_points,
        **kwargs,
    )
```

Then configure:

```python
_target_: str = (
    "hydra_plugins.hydra_optuna_sweeper._impl.create_nsgaiii_sampler"
)
```

Hydra's normal recursive `instantiate()` call will invoke the factory, which
converts the OmegaConf list before constructing Optuna's sampler.

Please also add a regression confirming that nested-list configuration is
converted to an array during instantiation.

plugins/hydra_optuna_sweeper/tests/test_optuna_sweeper_plugin.py
```diff
@@ -29,10 +31,12 @@
 
 from hydra_plugins.hydra_optuna_sweeper import _impl
 from hydra_plugins.hydra_optuna_sweeper.config import (
+    MOTPESamplerConfig,
```
himkt:
```suggestion
```

PR #3333, opened 2026-07-31
fix(optuna-sweeper): pass values=None when marking trial as FAIL, add NaN check

AbdullahRasheed45:
## Problem

Closes #2237

When a Hydra run returns NaN objective values (e.g. a diverged model), the Optuna sweeper crashes with a double exception:

1. `study.tell(state=COMPLETE, values=[nan])` raises `ValueError: Trial N failed, because the objective function returned nan.`
2. The `except Exception` handler sets `state = TrialState.FAIL` but still passes the original `values` to a second `study.tell` call, which raises another `ValueError: Values cannot be specified when state is TrialState.PRUNED or TrialState.FAIL.`

This second exception is **unhandled** and crashes the entire sweep.

## Fix

Two-part change in `plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py`:

**1. Proactive NaN check** — after computing `values`, raise a descriptive `ValueError` if any value is NaN, so the failure is captured cleanly by the existing `except` block with a readable warning message.

**2. Fix the except block** — change `study.tell(trial=trial, state=state, values=values)` to `study.tell(trial=trial, state=state, values=None)`. Optuna's API requires `values=None` when state is `FAIL` or `PRUNED`; passing the NaN values caused the secondary crash.

## Behaviour after fix

A trial that returns NaN objective values is recorded as `FAIL` in the Optuna study (consistent with a diverged run), a warning is logged, and the sweep continues to the next trial instead of crashing.

PR #3451, opened 2026-09-16
Fix Optuna sweeper crash on NaN or invalid trial results

### LINKED ISSUE #2237: [Bug] Optuna sweeper raises uncaught exception when run ends with nan value

saivedant169:
## Contributor License Agreement

Hydra requires contributors to accept its
[Individual Contributor License Agreement](https://hydra.cc/legal/individual-cla/).
CLA Assistant will check the pull request and provide a signing link if needed.
Obtain any approval required by your employer or other obligations before
signing.

## Motivation

When Optuna rejects a trial result, the sweeper retries `study.tell()` with `state=FAIL` but still passes the rejected values. Optuna doesn't accept values for failed trials, so the retry raises and the whole sweep stops. This passes `values=None` on that retry, which is the fix agreed in #2237 and in the review on #3333.

The same retry path is hit by a multi-objective result with the wrong number of values, so the test covers that case too.

One thing I left as is: with the default `max_failure_rate=0.0`, the rate check logs an error for these trials but the sweep keeps going, because `ret.return_value` only re-raises for jobs that failed and these jobs returned normally. That matches "subsequent trials can continue" from the #3333 review, so I didn't change it here.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/hydra-ecosystem/hydra/blob/main/CONTRIBUTING.md)?

Yes

#2237

## Test Plan

- Added `test_rejected_result_marks_trial_failed` for a NaN objective and for the wrong number of values with two objectives. Each checks that the first trial is recorded as `FAIL` and the next two still run and complete. Both cases fail without the fix.
- `pytest` in `plugins/hydra_optuna_sweeper`: 59 passed.
- `ruff check`, `ruff format --check`, `yamllint --strict` and `pyrefly check` on the plugin are clean.

I ran these against hydra-core 1.4.0.dev10, since building core from main needs Java for the ANTLR parser.

## Related Issues and PRs

Fixes #2237. Picks up #3333.


omry (approved):
Thanks for the fix. Recording rejected results as failed trials while later trials continue matches #2237. The NaN and invalid multi-objective regression cases cover the retry, and I clarified the news fragment to describe individual failed trials. Approved.

Issue #3444, opened 2026-09-10
Include ANTLR license with the build JAR in Hydra 1.4 sdists


omry:
## Summary

Hydra 1.4's source distribution includes `build_helpers/bin/antlr-4.11.1-complete.jar`, but does not include the repository's existing `ATTRIBUTION/LICENSE-antlr4` file.

Hydra 1.4 does not carry its own copy of the ANTLR Python runtime: it depends on OmegaConf 2.4 and the generated parser imports `omegaconf.vendor.antlr4`. Runtime attribution therefore belongs to the OmegaConf distribution. This issue is specifically about the ANTLR generator JAR redistributed in Hydra's sdist.

## Proposed fix

- Include `ATTRIBUTION/LICENSE-antlr4` in the sdist alongside the JAR.
- Declare it as a distribution license file using the packaging configuration supported by the project.
- Add an artifact-level check that verifies:
  - an sdist containing the ANTLR JAR also contains the ANTLR license;
  - the wheel does not accidentally contain the build-only JAR.

Alternatively, if building from an sdist no longer needs the JAR, exclude the JAR from the sdist.

PR #3460, opened 2026-09-22
Check the ANTLR license travels with the JAR in built artifacts

adarshsm:
Refs #3444 — the remaining item from that issue.

## What was already done

`89e45f3fb3` ("Include ANTLR license in Hydra source distributions") landed the
first two bullets on `main`: `MANIFEST.in` includes `ATTRIBUTION/LICENSE-antlr4`,
and `setup.py` declares `license_files=["LICENSE", "ATTRIBUTION/LICENSE-antlr4"]`.

The third bullet is still open:

> Add an artifact-level check that verifies:
>   - an sdist containing the ANTLR JAR also contains the ANTLR license;
>   - the wheel does not accidentally contain the build-only JAR.

## This PR

`check_bundled_jar_licenses()` in `tools/release/release.py`, called from
`check_build_artifacts()`, so it runs in `publish.yml`'s `build-artifacts` job
ahead of the `pypi-publish` approval gate:

* an sdist (`*.tar.gz`) that contains any `.jar` must also contain
  `ATTRIBUTION/LICENSE-antlr4` — otherwise `ValueError`;
* a wheel that contains any `.jar` is rejected outright, since the ANTLR JAR is
  a build-time tool;
* artifacts with no JAR (every plugin sdist, and the hydra-core wheel) pass
  unchanged.

This asserts today's correct state rather than changing it: `main` already ships
the license in the sdist, and `packages=find_namespace_packages(include=["hydra",
"hydra.*"])` keeps `build_helpers/` out of the wheel. The point is that a later
packaging change cannot silently undo either.

## Verification

Five new tests in `tools/release/test_release.py` build synthetic archives and
assert both directions (sdist with JAR + license, sdist with JAR only, sdist
with no JAR, wheel with JAR, wheel without). All five fail on `main` and pass
here; `pytest tools/release/test_release.py` is 45 passed.

One existing-test fix came with it: `test_check_build_artifacts_upgrades_smoke_
environment_pip` created its wheel with `Path.touch()`, which is not a readable
zip, so the new check could not open it. It now writes a real one-entry zip.

`ruff format --check` and `ruff check` clean on both files. No news fragment —
per AGENTS.md this is release tooling and does not change the shipped product.

Happy to fold this into the packaging commit instead if you'd rather keep #3444
as one change.


omry:
Thanks for adding this release artifact check. The change looks good to me. Before I can accept the PR, I need the CLA signed and the `license/cla` check passing.

